Issue #84, opened 2015-01-13
Clang compiler crashes on LinkedList range-based for


wivlaro:
I reported an upstream bug to Apple. Not sure if/when it will get fixed (or how to make it public). My reflog would have me believe it was introduced between @8611f9f9027c0b8fd69a1bf60ff53c2e4af6a5db and @f30a4677eb441419db7440b1fa2d15418f1ae9aa because it was building fine before then.

```
Scanning dependencies of target MagnumMeshTools
[ 53%] Building CXX object src/Magnum/MeshTools/CMakeFiles/MagnumMeshTools.dir/CombineIndexedArrays.cpp.o
Stack dump:
[...]

PR #85, opened 2015-01-14
Fixes required to fix compiler crashes in issue #84

wivlaro:
For issue #84


mosra:
As said in #84, fixed the original issue in mosra/corrade@18ee9912274babd31b8dbc3b24f1c0f9d6b3fe8f, so this is no longer needed.

PR #87, opened 2015-01-24
Fix for shader warning strings with null characters in them

wivlaro:
This was corrupting my console output with the null 0 characters.


mosra:
Huh. Looking at that, the problems is somewhere else. I'll fix the original issue rather than the consequences.

mosra:
Should be fixed (and prevented from appearing again) in 2ed432aeb19f311d6b639a7ff4a25f66dfd2da07 and 0fcf763900765bd86aa860cfef5c817677492e73.

PR #88, opened 2015-02-01
Added Matrix4::lookAt

wivlaro:


mosra:
Thinking about the two `normalized()` calls, am I assuming right that if `up` was expected to be already normalized, doing just
~~~cpp
Vector3<T> xaxis = Vector3<T>::cross(upNormalized,zaxis); // The "right" vector.
~~~
would be enough?

Not sure about the desired use cases, but if the user could ensure that `up` is normalized (which I assume in most cases is), the function would be one `sqrt` operation cheaper.

wivlaro:
Sure. I guess we can just assert that up is normalized.

mosra:
Merged in 80e01ae18d7df57ca0db5e1c54e90816a1168773, thanks.

Sorry about the normalization, *brain fart* on my side (result of cross product is not always unit vector).

PR #94, opened 2015-04-09
Minor documentation fixes

Squareys:
Hello mosra, 

here is the promised pull request for the doc fixes. I didn't find many more, so here are four in total.
Before merging, you may just let this PR hang around until I have some more, if you want to.

I hope commit style is alright so far :)

Greetings,
Squareys


mosra:
Hey, thanks! Regarding the `public` modifier, the thing is that it's indeed possible to inherit the drawable privately and it would still work, but... explaining this in the documentation would do more harm than good and this way it's much simpler.

PR #107, opened 2015-08-27
Audio Improvements

Squareys:
Hi @mosra !

As usual, here is a work-in-progress pullrequest.

This pullrequest consist of:
 - Getters for `Source` and `Listener`
 - Tests of getters and setters of `Source` and `Listener`
 - Debug operator for `Source::DistanceModel`

**Moved to other branches:**
- [x] Add support for OpenAL extensions
   - [ ] Finish doc
   - [ ] Ensure Format Type parameter is actually a useful config setting.
   - [ ] Tests
- [x] Implement `AL_EXT_float32` and `AL_EXT_double`
- [x] Add support for the OpenAL soft hrtfs
- [ ] Implement OggAudioImporter (separate PR).

Greetings, Squareys

### DIFF
--- a/src/Magnum/Audio/CMakeLists.txt
+++ b/src/Magnum/Audio/CMakeLists.txt
@@ -54,7 +54,7 @@ if(BUILD_STATIC_PIC)
     set_target_properties(MagnumAudio PROPERTIES POSITION_INDEPENDENT_CODE ON)
 endif()
 
-target_link_libraries(MagnumAudio ${CORRADE_PLUGINMANAGER_LIBRARIES} ${OPENAL_LIBRARY})
+target_link_libraries(MagnumAudio Magnum ${CORRADE_PLUGINMANAGER_LIBRARIES} ${OPENAL_LIBRARY})
 
 install(TARGETS MagnumAudio
     RUNTIME DESTINATION ${MAGNUM_BINARY_INSTALL_DIR}
--- a/src/Magnum/Audio/Renderer.h
+++ b/src/Magnum/Audio/Renderer.h
@@ -29,6 +29,8 @@
  * @brief Class @ref Magnum::Audio::Renderer
  */
 
+#include <array>
+
 #include <al.h>
 
 #include "Magnum/Magnum.h"
@@ -90,6 +92,8 @@ class Renderer {
 
         /**
          * @brief Listener orientation
+         *
+         * The returned array consists of forward (index 0) and up (index 1) vectors.
          * @see @ref setListenerOrientation(), @fn_al{GetListenerfv} with @def_al{ORIENTATION}
          */
         static std::array<Vector3, 2> listenerOrientation() {
--- a/src/Magnum/Audio/Renderer.h
+++ b/src/Magnum/Audio/Renderer.h
@@ -62,6 +62,16 @@ class Renderer {
         /** @{ @name Listener positioning */
 
         /**
+         * @brief Listener position
+         * @see @ref setListenerPosition(), @fn_al{GetListenerfv} with @def_al{POSITION}
+         */
+        static Vector3 listenerPosition() {
+            Vector3 position;
+            alGetListenerfv(AL_POSITION, position.data());
+            return position;
+        }
+
+        /**
          * @brief Set listener position
          *
          * Default is `{0.0f, 0.0f, 0.0f}`.
@@ -79,18 +89,44 @@ class Renderer {
         }
 
         /**
+         * @brief Listener orientation
+         * @see @ref setListenerOrientation(), @fn_al{GetListenerfv} with @def_al{ORIENTATION}
+         */
+        static std::array<Vector3, 2> listenerOrientation() {
+            std::array<Vector3, 2> data;
+            alGetListenerfv(AL_ORIENTATION, data[0].data());
+            return data;
+        }
+
+        /**
          * @brief Set listener orientation
          *
          * The values must be linearly independent and don't need to be
          * normalized. Default is -Z and +Y.
          * @see @fn_al{Listenerfv} with @def_al{ORIENTATION}
          */
-        static void setListenerOrientation(const Vector3& forward, const Vector3& up);
+        static void setListenerOrientation(const Vector3& forward, const Vector3& up) {
+            const Vector3 data[] = {forward, up};
+            alListenerfv(AL_ORIENTATION, data[0].data());
+        }
 
         /** @overload
          * @see @fn_al{Listeneriv} with @def_al{ORIENTATION}
          */
-        static void setListenerOrientation(const Vector3i& forward, const Vector3i& up);
+        static void setListenerOrientation(const Vector3i& forward, const Vector3i& up) {
+            const Vector3i data[] = {forward, up};
+            alListeneriv(AL_ORIENTATION, data[0].data());
+        }
+
+        /**
+         * @brief Listener position
+         * @see @ref setListenerVelocity(), @fn_al{GetListenerfv} with @def_al{VELOCITY}
+         */
+        static Vector3 listenerVelocity() {
+            Vector3 velocity;
+            alGetListenerfv(AL_VELOCITY, velocity.data());
+            return velocity;
+        }
 
         /**
          * @brief Set listener velocity
@@ -142,6 +178,16 @@ class Renderer {
         };
 
         /**
+         * @brief Listener gain
+         * @see @ref setListenerGain(), @fn_al{GetListenerf} with @def_al{GAIN}
+         */
+        static Float listenerGain() {
+            Float gain;
+            alGetListenerf(AL_GAIN, &gain);
+            return gain;
+        }
+
+        /**
          * @brief Set listener gain
          *
          * Default is `1.0f`, which means that the sound is unattenuated.
@@ -150,6 +196,14 @@ class Renderer {
          */
         static void setListenerGain(Float gain) {
             alListenerf(AL_GAIN, gain);
+        }
+
+        /**
+         * @brief Doppler factor
+         * @see @ref setDopplerFactor(), @fn_al{GetFloat} with @def_al{DOPPLER_FACTOR}
+         */
+        static Float dopplerFactor() {
+            return alGetFloat(AL_DOPPLER_FACTOR);
         }
 
         /**
@@ -163,6 +217,14 @@ class Renderer {
         }
 
         /**
+         * @brief Speed of sound
+         * @see @ref setSpeedOfSound(), @fn_al{GetFloat} with @def_al{SPEED_OF_SOUND}
+         */
+        static Float speedOfSound() {
+            return alGetFloat(AL_SPEED_OF_SOUND);
+        }
+
+        /**
          * @brief Set speed of sound
          *
          * Default is `343.3f` (meters per second).
@@ -170,6 +232,14 @@ class Renderer {
          */
         static void setSpeedOfSound(Float speed) {
             alSpeedOfSound(speed);
+        }
+
+        /**
+         * @brief Distance model
+         * @see @ref setDistanceModel(), @fn_al{GetInteger} with @def_al{DISTANCE_MODEL}
+         */
+        static DistanceModel distanceModel() {
+            return DistanceModel(alGetInteger(AL_DISTANCE_MODEL));
         }
 
         /**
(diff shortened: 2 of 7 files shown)


src/Magnum/Audio/CMakeLists.txt
```diff
@@ -45,6 +45,11 @@ set(MagnumAudio_HEADERS
 
     visibility.h)
 
+if(WITH_SCENEGRAPH)
```
mosra: This is great. Exactly what I wanted to note but then I saw the conditional compilation is already done :)

mosra:
That's amazing amount of work you are doing here.

Re the naming: 

 - the Drawable feature was named after "what it is able to do" (drawable) or "what does it do" (draw), but the Audible is "what does it feel like", so it's a little inconsistent (with this approach the Drawable would have to be named Seeable). My first idea was Playable, more consistent with the rest, although it might be confused with Animable (but that should be okay, as it is in Audio namespace anyway)
 - the Listening feature - just Listener (similarly to how Camera is named)

OggAudioImporter - will you base it on stb_vorbis or the "official" libvorbis? I think it will be better to have it in magnum-plugins repo (keeping only the most essential plugins in the core repo).

src/Magnum/Audio/CMakeLists.txt
Squareys: Yeah, I figured you would obviously want to have Audio without SceneGraph sometimes.

mosra: Yup, I wanted to fix that for DebugTools, if you enable them, they bring everything else and a kitchen sink now :)

Squareys:
> will you base it on stb_vorbis or the "official" libvorbis?

I would have used stb_vorbis as you suggested

mosra:
Sure! :)

Just a quick note about the `stb_*` libraries -- the original author is developing them as quick-and-dirty solutions to get up and running in demo and prototype code, they are small, easy to use, without dependencies, but they have some limitations and *might* not be as fast as the "official" decoders. That's no problem here, though -- for initial support this is good enough, very easy to implement, and if anyone would need the official "libvorbis" plugin, we can add it later.

The naming -- to be consistent with the rest, let's name the plugin `StbVorbisAudioImporter`, the class will be then `Audio::StbVorbisImporter` (similarly to the distinction in `WavAudioImporter`).

Squareys:
@mosra regarding the velocity in `Playable`. This would require access to a `Timeline` or something, correct? I could have flags for `position`, `velocity` and `orientation`, so that the `veloctiy` flag, for example, is only activated when a timeline has been set.

Squareys:
I added some commits for initial OpenAL extensions support. Still very sketchy and experimental and untested. But hrtfs work. 

I also added a Configuration struct for the Audio Context, that is very sketchy aswell. I might move Buffer::Format out of Buffer in case Configuration will *actually* need it. In that case I will take the necessary deprecation measures.

Squareys:
Okay, I rebased to have only the audio-improvement commits in this PR. Tests run and all code unrelated to improvements of existing code should have been moved away.

As soon as this is merged, I will properly rebase the other branches `audio-features` and `audio-extension-support` in my fork and file the work-in-progress pullrequests.

Good Night! :)

src/Magnum/Audio/Test/RendererTest.cpp
```diff
@@ -34,10 +36,28 @@ struct RendererTest: TestSuite::Tester {
...
+    void speedOfSound();
+    void dopplerFactor();
+    void distanceModel();
+
+    Context _context;
 };
 
-RendererTest::RendererTest() {
-    addTests({&RendererTest::debugError});
+RendererTest::RendererTest(): _context{} {
```
mosra: No need to call `_context{}` explicitly (it will be done implicitly by the compiler).

src/Magnum/Audio/Test/SourceTest.cpp
```diff
@@ -46,6 +71,102 @@ void SourceTest::debugState() {
     CORRADE_COMPARE(out.str(), "Audio::Source::State::Playing\n");
 }
 
+void SourceTest::position() {
+    Source source{};
```
mosra: The {} at the end is superfluous, similarly in all other cases below.

src/Magnum/Audio/Test/RendererTest.cpp
```diff
@@ -46,6 +66,63 @@ void RendererTest::debugError() {
...
+
+void RendererTest::listenerVelocity() {
+    constexpr Vector3 vel{1, 3, 2};
+    Renderer::setListenerVelocity(vel);
+
+    CORRADE_COMPARE(Renderer::listenerVelocity(), vel);
+}
+
+void RendererTest::listenerGain() {
+    constexpr Float gain = 0.512;
```
mosra: Please use `f` suffix on floats (MSVC otherwise complains *a lot* and sometimes even refuses to compile that code).

src/Magnum/Audio/Test/RendererTest.cpp
```diff
@@ -46,6 +66,63 @@ void RendererTest::debugError() {
...
 }
 
+void RendererTest::debugDistanceModel() {
+    std::ostringstream out;
+    Debug(&out) << Renderer::DistanceModel::Inverse;
+    CORRADE_COMPARE(out.str(), "Audio::Renderer::DistanceModel::Inverse\n");
+}
+
+void RendererTest::listenerOrientation() {
+    constexpr Vector3 up{1, 2, 3}, fwd{3, 2, 1};
```
mosra: Please use `{1.0f, 2.0f, 3.0f}` instead of integers. In combination with the {}-style initialization it causes implicit conversion warnings/errors on some compilers.

src/Magnum/Audio/Test/SourceTest.cpp
```diff
@@ -46,6 +71,102 @@ void SourceTest::debugState() {
...
+    source.setLooping(true);
+    CORRADE_COMPARE(true, source.isLooping());
+    source.setLooping(false);
+    CORRADE_COMPARE(false, source.isLooping());
+}
+
+void SourceTest::relative() {
+    Source source{};
+    source.setRelative(true);
+    CORRADE_COMPARE(true, source.isRelative());
```
mosra:
```cpp
source.setRelative(true);
CORRADE_VERIFY(source.isRelative());
source.setRelative(false);
CORRADE_VERIFY(!source.isRelative());
```

In case of boolean values it's shorter to write :)

src/Magnum/Audio/Test/SourceTest.cpp
```diff
@@ -46,6 +71,102 @@ void SourceTest::debugState() {
...
+    Source source{};
+    constexpr Float gain = 0.4144;
+    source.setMinGain(gain);
+
+    CORRADE_COMPARE(source.minGain(), gain);
+}
+
+void SourceTest::coneAnglesAndGain() {
+    Source source{};
+    constexpr Deg outerAngle = Deg(12);
```
mosra:
```cpp
constexpr auto outerAngle = 12.0_degf;
```

Less typing :)

src/Magnum/Audio/Renderer.h
```diff
@@ -188,6 +240,21 @@ class Renderer {
 /** @debugoperatorclassenum{Magnum::Audio::Renderer,Magnum::Audio::Renderer::Error} */
 Debug MAGNUM_AUDIO_EXPORT operator<<(Debug debug, Renderer::Error value);
 
+/** @debugoperatorclassenum{Magnum::Audio::Renderer,Magnum::Audio::Renderer::DistanceModel} */
+Debug MAGNUM_AUDIO_EXPORT operator<<(Debug debug, Renderer::DistanceModel value);
+
+inline Vector3 Renderer::listenerPosition() {
```
mosra: I think you can put these directly into the class definition, they aren't too long to be hidden below.

src/Magnum/Audio/Renderer.h
```diff
@@ -188,6 +240,21 @@ class Renderer {
...
+    Vector3 position;
+    alGetListenerfv(AL_POSITION, position.data());
+    return position;
+}
+
+inline std::array<Vector3, 2> Renderer::listenerOrientation() {
+    std::array<Vector3, 2> data = {Vector3{}, Vector3{}};
```
mosra: Initializing the array is not needed, it's done implicitly and will be overwritten by the getter below anyway.

mosra:
Commented on a few minor things inline, other than that I have no objections, great stuff :)

Please rebase on the latest master, the AppVeyor CI now builds also the Audio library and thus will be able to test this PR on MSVC 2015 right away.

Squareys:
Thanks! I will hopefully get around to fixing those by tomorrow.

src/Magnum/Audio/Test/RendererTest.cpp
Squareys: - [x] *for keeping track*

src/Magnum/Audio/Test/SourceTest.cpp
Squareys: - [x] *for keeping track*

src/Magnum/Audio/Test/RendererTest.cpp
Squareys: - [x] *for keeping track*

src/Magnum/Audio/Test/RendererTest.cpp
Squareys: - [x] *for keeping track*

src/Magnum/Audio/Test/SourceTest.cpp
Squareys: - [x] *for keeping track*

src/Magnum/Audio/Test/SourceTest.cpp
Squareys: - [x] *for keeping track*

src/Magnum/Audio/Renderer.h
Squareys: - [x] *for keeping track*

src/Magnum/Audio/Renderer.h
Squareys: - [x] *for keeping track*

Squareys:
Alright, the tests still fail on AppVeyor. I believe this must be a problem with the audio context (and device) not beeing created:

> Audio::Context: cannot open sound device OpenAL Soft
AL lib: (EE) ALCmmdevPlayback_open: Device init failed: 0x80070490

The things you pointed out have been fixed :)

Squareys:
I think `null` device could be the solution: https://github.com/kcat/openal-soft/blob/master/Alc/ALc.c#L104
That should be passed as a context configuration parameter, though. Which means that the `audio-extension` pullrequest needs to be finished first.

mosra:
Flawless! :)

The AppVeyor config is in [package/ci/appveyor.yml](https://github.com/mosra/magnum/blob/master/package/ci/appveyor.yml), commiting any change to it will immediately affect the AppVeyor build. The file has to be in `%AppData%`, right? So I think adding something like the following after [the block where OpenAL is extracted](https://github.com/mosra/magnum/blob/03c8272aea9ff5d30131786d906427bd81c00f9b/package/ci/appveyor.yml#L19) should do the trick:

```cmd
echo "[General]^r^ndrivers=null" > %AppData%/alsoft.ini
```

Guessing from the top of my head, might need to be slightly different, can't test right now.

**EDIT:** ugh, it seems like echoing a newline needs quite a lot of weird trickery: http://stackoverflow.com/questions/132799/how-can-you-echo-a-newline-in-batch-files Why, Windows, why? :D

Squareys:
Alright, awesome, will do so later today!

Squareys:
> (echo [General] && echo drivers=null) > %AppData%/alsoft.ini

Works for me locally, but AppVeyor does not like it. Maybe insufficient permissions?

LB--:
Shouldn't that be `(echo [General] && echo.drivers=null)`? The period immediately after `echo` adds a newline.

Squareys:
Well... it works on my system, output is (and should be):
```
[General] 
drivers=null
```

Squareys:
Alright, works now. This is ready for merge then, once again.

mosra:
Merged, thank you a lot! :)

PR #110, opened 2015-09-09
[WIP] Audio features

Squareys:
Hi @mosra, Hi everybody!

As discussed, here is a work-in-progress pullrequest for everything related to Audio and SceneGraph.

TODOs:
 - [x] PlayableGroup
 - [x] Playable
 - [x] Listener
 - Adapt audio-example (later)

Greetings, Squareys

### DIFF
--- a/src/Magnum/Audio/CMakeLists.txt
+++ b/src/Magnum/Audio/CMakeLists.txt
@@ -45,6 +45,12 @@ set(MagnumAudio_HEADERS
 
     visibility.h)
 
+if(WITH_SCENEGRAPH)
+    list(APPEND MagnumAudio_HEADERS
+        Playable.h
+        PlayableGroup.h)
+endif()
+
 # Audio library
 add_library(MagnumAudio ${SHARED_OR_STATIC}
     ${MagnumAudio_SRCS}
@@ -55,6 +61,10 @@ if(BUILD_STATIC_PIC)
 endif()
 
 target_link_libraries(MagnumAudio Magnum ${CORRADE_PLUGINMANAGER_LIBRARIES} ${OPENAL_LIBRARY})
+
+if(WITH_SCENEGRAPH)
+    target_link_libraries(MagnumAudio MagnumSceneGraph)
+endif()
 
 install(TARGETS MagnumAudio
     RUNTIME DESTINATION ${MAGNUM_BINARY_INSTALL_DIR}
--- a/src/Magnum/Audio/CMakeLists.txt
+++ b/src/Magnum/Audio/CMakeLists.txt
@@ -47,8 +47,12 @@ set(MagnumAudio_HEADERS
 
 if(WITH_SCENEGRAPH)
     list(APPEND MagnumAudio_HEADERS
+        Listener.h
         Playable.h
         PlayableGroup.h)
+
+    list(APPEND MagnumAudio_SRCS
+        Listener.cpp)
 endif()
 
 # Audio library
--- a/src/Magnum/Audio/Listener.cpp
+++ b/src/Magnum/Audio/Listener.cpp
@@ -1,0 +1,113 @@
+/*
+    This file is part of Magnum.
+
+    Copyright © 2010, 2011, 2012, 2013, 2014, 2015
+              Vladimír Vondruš <mosra@centrum.cz>
+    Copyright © 2015 Jonathan Hale <squareys@googlemail.com>
+
+    Permission is hereby granted, free of charge, to any person obtaining a
+    copy of this software and associated documentation files (the "Software"),
+    to deal in the Software without restriction, including without limitation
+    the rights to use, copy, modify, merge, publish, distribute, sublicense,
+    and/or sell copies of the Software, and to permit persons to whom the
+    Software is furnished to do so, subject to the following conditions:
+
+    The above copyright notice and this permission notice shall be included
+    in all copies or substantial portions of the Software.
+
+    THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+    IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+    FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+    THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+    LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+    FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+    DEALINGS IN THE SOFTWARE.
+*/
+
+#include "Magnum/Audio/Listener.h"
+#include "Magnum/Audio/Playable.h"
+#include "Magnum/Audio/PlayableGroup.h"
+
+
+namespace Magnum { namespace Audio {
+
+using namespace SceneGraph;
+
+namespace Implementation {
+    void* activeListener = nullptr;
+}
+
+namespace {
+
+/* Create a Matrix4 from a Matrix3/Matrix4 */
+inline Matrix4 padMatrix4(const Matrix3& m) {
+    return Matrix4{Vector4::pad(m[0]), Vector4::pad(m[1]), Vector4::pad(m[2]), Vector4{}};
+}
+
+inline Matrix4 padMatrix4(const Matrix4& m) {
+    return m;
+}
+
+}
+
+template<UnsignedInt dimensions> void Listener<dimensions>::clean(const MatrixTypeFor<dimensions, Float>& absoluteTransformationMatrix) {
+    if(!isActive()) {
+        /* Only clean if this Listener is active */
+        return;
+    }
+
+    Renderer::setListenerPosition(soundTransformation().transformVector(Vector3::pad(absoluteTransformationMatrix.translation())));
+
+    const Vector3 fwd = soundTransformation().transformVector(-padMatrix4(absoluteTransformationMatrix).backward());
+    const Vector3 up = soundTransformation().transformVector(padMatrix4(absoluteTransformationMatrix).up());
+
+    Renderer::setListenerOrientation(fwd, up);
+
+    Renderer::setListenerGain(_gain);
+
+    // TODO: velocity
+}
+
+template<UnsignedInt dimensions> void Listener<dimensions>::update(std::initializer_list<std::reference_wrapper<PlayableGroup<dimensions>>> groups) {
+
+    /* Check if active listener just changed to this */
+    if(this != Implementation::activeListener) {
+        /* Ensure that clean() is called also when switching between (clean)
+           listeners */
+        Implementation::activeListener = this;
+        this->object().setDirty();
+
+        /* Listener gain needs to be updated only when active listener changed
+           and in setGain() */
+        Renderer::setListenerGain(_gain);
+    }
+
+    /* Add all objects of the Playables in the PlayableGroups to a vector to
+       later setClean() */
+    std::vector<std::reference_wrapper<AbstractObject<dimensions, Float>>> objects;
+
+    objects.push_back(this->object());
+    for(PlayableGroup<dimensions>& group : groups) {
+        for(UnsignedInt i = 0; i < groups.size(); ++i) {
+            objects.push_back(group[i].object());
+        }
+    }
+
+    /* Use the more performant way to set multiple objects clean */
+    AbstractObject<dimensions, Float>::setClean(objects);
+}
+
+/* On non-MinGW Windows the instantiations are already marked with extern
+   template */
+#if !defined(CORRADE_TARGET_WINDOWS) || defined(__MINGW32__)
+#define MAGNUM_AUDIO_EXPORT_HPP MAGNUM_AUDIO_EXPORT
+#else
+#define MAGNUM_AUDIO_EXPORT_HPP
+#endif
+
+#ifndef DOXYGEN_GENERATING_OUTPUT
+template class MAGNUM_AUDIO_EXPORT_HPP Listener<2>;
+template class MAGNUM_AUDIO_EXPORT_HPP Listener<3>;
+#endif
+
+}}
--- a/src/Magnum/Audio/Listener.h
+++ b/src/Magnum/Audio/Listener.h
@@ -1,0 +1,224 @@
+#ifndef Magnum_Audio_Listener_h
+#define Magnum_Audio_Listener_h
+/*
+    This file is part of Magnum.
+
+    Copyright © 2010, 2011, 2012, 2013, 2014, 2015
+              Vladimír Vondruš <mosra@centrum.cz>
+    Copyright © 2015 Jonathan Hale <squareys@googlemail.com>
+
+    Permission is hereby granted, free of charge, to any person obtaining a
+    copy of this software and associated documentation files (the "Software"),
+    to deal in the Software without restriction, including without limitation
+    the rights to use, copy, modify, merge, publish, distribute, sublicense,
+    and/or sell copies of the Software, and to permit persons to whom the
+    Software is furnished to do so, subject to the following conditions:
+
+    The above copyright notice and this permission notice shall be included
+    in all copies or substantial portions of the Software.
+
+    THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+    IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+    FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+    THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+    LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+    FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+    DEALINGS IN THE SOFTWARE.
+*/
+
+/** @file
+ * @brief Class @ref Magnum::Audio::Listener, @ref Magnum::Audio::Listener2D, @ref Magnum::Audio::Listener3D
+ */
+
+#include <string>
+#include <al.h>
+
+#include <Magnum/Math/Matrix3.h>
+#include <Magnum/Math/Matrix4.h>
+#include <Magnum/Math/Quaternion.h>
+#include <Magnum/SceneGraph/AbstractFeature.h>
+#include <Magnum/SceneGraph/AbstractObject.h>
+
+#include "Magnum/Audio/Renderer.h"
+#include "Magnum/Audio/PlayableGroup.h"
+
+namespace Magnum { namespace Audio {
+
+namespace Implementation {
+    /* Pointer to the currently active listener. Should never be dereferenced,
+       since may be null at any given time. */
+    extern MAGNUM_AUDIO_EXPORT void* activeListener;
+}
+
+/**
+@brief Listener
+
+Feature which manages the position, orientation and gain of the OpenAL listener
+for an @ref SceneGraph::Object.
+
+## Usage
+
+Minimal scene setup for a @ref Listener3D will require the following code:
+
+@code
+Scene3D scene;
+Object3D object{&scene};
+Listener3D listener{object};
+
+// ... every frame, update the listener to changes in object transformation:
+listener.update({});
+
+@endcode
+
+For two dimensional scenes simply replace all `3D` with `2D`.
+
+### Using Listener with PlayableGroup
+
+When using @ref PlayableGroup, you can update the listener and groups as
+follows:
+
+@code
+// ...
+Listener3D listener{object};
+PlayableGroup3D group1, group2;
+
+// ... and every frame:
+listener.update({group1, group2});
+@endcode
+
+## Active Listener
+
+There can only be at the most *one* active listener at a given time, i.e. the
+one on which @ref Listener::update() was called last. This is because OpenAL
+only supports notion of one listener. Having multiple @ref Listener2D or
+@ref Listener3D instances can still be useful for conveniently switching
+between them for cinematics for example.
+
+@ref AbstractObject::setClean() will not affect inactive listeners.
+
+## Sound Transformation
+
+@ref Listener::setSoundTransformation() enables you to set a transformation
+matrix which is applied to the listeners orientation and position before passed
+onto OpenAL. This can be used for rotating two dimensional scenes as planes
+into the three dimensional audio space or even scaling the audio scene to match
+a certain world scale. In the later case you might want to instead consider
+@ref Renderer::setSpeedOfSound().
+
+-   @ref Listener2D
+-   @ref Listener3D
+
+@see @ref Audio::Renderer, @ref Playable, @ref PlayableGroup
+*/
+template <UnsignedInt dimensions> class Listener: public SceneGraph::AbstractFeature<dimensions, Float> {
+        friend class ActiveListenerHolder;
+
+    public:
+
+        /**
+         * @brief Constructor
+         * @param object    Object this listener belongs to
+         *
+         * Creates a listener with a forward vector of `{0.0f, 0.0f, -1.0f}` and
+         * up vector of `{0.0f, 1.0f, 0.0f}`. These vectors cannot be changed,
+         * the listeners orientation and translation can be instead affected by
+         * `object` or via @ref Listener::setSoundTransformation().
+         * @see @ref setGain()
+         */
+        explicit Listener(SceneGraph::AbstractObject<dimensions, Float>& object):
+            SceneGraph::AbstractFeature<dimensions, Float>(object),
+            _gain{1.0f}
+        {
+            SceneGraph::AbstractFeature<dimensions, Float>::setCachedTransformations(SceneGraph::CachedTransformation::Absolute);
+        }
+
+        /**
+         * @brief Sound transformation
+         * @see @ref Listener::setSoundTransformation()
+         */
+        const Matrix4& soundTransformation() const {
+            return _soundTransformation;
+        }
+
+        /**
+         * @brief Set sound transformation
+         * @param soundTransformation Transformation for transforming from
+         *        world to listener space
+         * @return Reference to self (for method chaining)
+         * @see @ref Listener::soundTransformation()
+         */
+        Listener<dimensions>& setSoundTransformation(const Matrix4& soundTransformation) {
+            _soundTransformation = soundTransformation;
+            this->object().setDirty();
+            return *this;
+        }
+
+        /**
+         * @brief Update the listener
+         * @param groups Groups to update
+         *
+         * Makes this Listener the active listener and calls
+         * @ref SceneGraph::AbstractObject::setClean() on its parent object and
+         * all objects of the @ref Playable s in the group. Updates listene
+         * related configuration for @ref Renderer (position, orientation, gain).
+         */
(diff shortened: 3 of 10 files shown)


src/Magnum/Audio/Test/ListenerTest.cpp
```diff
@@ -0,0 +1,70 @@
...
+void ListenerTest::testFeature() {
+    Scene3D scene;
+    Object3D object{&scene};
+    Listener3D listener{object};
+
+    constexpr Vector3 offset{1.0f, 2.0f, 3.0f};
+    object.translate(offset);
+    object.setClean();
+
+    CORRADE_COMPARE(Renderer::listenerPosition(), Vector3(1.0f, 2.0f, 3.0f));
```
Squareys: I should probably compare to `offset` here.

mosra:
As per our discussion, I would suggest having `Listener` with `use(PlayableGroup&)` function (just like there is `Camera::draw(DrawableGroup&)`, which would "make the listener active for given frame", i.e. calling `use()` sets all `Renderer` properties instead of doing that in the `clean()` function.

The listener (both 2D and 3D one) *might* contain general 3x3 rotation/scaling matrix (for proper scaling in 2D/3D and rotation in 2D) and  instead of just rotation axis in 2D (just like the Camera has projection matrix) that would be applied to all playables and also the listener on `use()`, so the listener implementation can then be more simple and the same for both 2D and 3D.

There are some performance considerations and corner cases that I need to think more about, though..

Squareys:
@mosra A few things I would like to discuss:

1. Can we name `Listener.use(PlayableGroup&)` `Listener.update(PlayableGroup&)`? I believe it fits better for the following reasons:

**use**: Is not clear about whether we are using the PlayableGroup or the Listener and does not signal that this method should be called every frame.
**update**: Obviously needs to be called every frame and updates the sources transformation.

(I also considered `.play()` to be consistent with the naming in `Camera`, but that isn't what that method is actually doing and therefore would just confuse people.)

2. A common usecase for PlayableGroups could also be grouping sounds by type (Music, SFX, Speech for example) to set their gain according to some setting. This, of course, would require "using" or "updating" mutliple PlayableGroups. Therefore setting the listeners properties would be done multiple times.

I would suggest a `Listener.makeActive()` method or using `clean()` together with some "is the active listener" flag.

3. > The listener (both 2D and 3D one) *might* contain general 3x3 rotation/scaling matrix

So, you mean, it allways contains it, but may not be used? Or what does the "might" refer to?

mosra:
Okay, I think I have it. Scrap all my previous thoughts, let me explain again from scratch:

What do we need (or, rather, what I think the user would need):

-   Organize sounds into different groups, being able to play/pause given group, make given group louder/quieter etc.
-   Having more than one listener (useful for cinematics), of course only one listener active at a time
-   Being able to transform the "hearable" part in addition to the "visible" transformation given by the scenegraph (mainly to scale the scene so one unit is really one meter for speed of sound etc. to work properly)

What do we want to avoid:

-   Doing too much stuff in one frame (i.e. bulk update of audio source positions)

My idea:

-   `PlayableGroup` -- contains 4x4 matrix that defines additional "audio" transformation for all playables in that group
-   `Playable` -- manages source position, part of some `PlayableGroup`. Passes source position and direction (combination of scenegraph and "audio" transformation) to OpenAL only on `clean()` or when its group changes the transformation matrix.
-   `Listener` -- also contains 4x4 matrix that defines its additional "audio" transformation (in addition to its transformation given by scenegraph).

Different playable groups and listeners obviously can have different transformations (mainly to achieve certain effects in 2D), but the common usage would be having one common transformation for all of them (i.e. scale everything by a factor of 100).

Now for the multiple listeners -- I could not think of any better way to switch between multiple listeners than needing to call some activating/updating function on one of them per frame. That function will call `Audio::Renderer::setOrientation()` and `Audio::Renderer::setPosition()` combined from listener's scenegraph transformation and additional "audio" transformation. Doing that in `clean()` would mean that there needs to be some sort of singleton that ensures only one of these listeners is calling functions on `Audio::Renderer` at a time. Ugly.

The listener would be, contrary to what I said earlier, completely independent of any `PlayableGroup`. As you pointed out, it would only complicate things even further.

For the naming -- let's call it `Listener::update()` as you said (and no parameters). `makeActive()` sounds to me like it doesn't need to be called every frame.

I'll comment on implementation details inline tomorrow, I need to get some rest now :)

Squareys:
Wow, first of all: thank you for taking so much time to answer!

I'd like to suggest one last thing:

> Doing that in clean() would mean that there needs to be some sort of singleton that ensures only one of these listeners is calling functions on Audio::Renderer at a time. Ugly.

I see what you mean, but I think there is a way: `Listener` could have one static pointer to a `Listener`, with the sole purpose to track the *active Listener*. It is not singleton, since it does not ensure there is only one instance. On a `setActive()` call, this pointer would be set to the Listener the method is called on and the listener is set to dirty to be updated on the next frame. This would further push

> What do we want to avoid:
- Doing too much stuff in one frame (i.e. bulk update of audio source positions)

Would you still consider that ugly? If yes, I will implement as suggested. Since I believe that cinematics is a less common use case than just single-listener for gameplay, I'm kinda trying to avoid the (minor?) performance sacrifice for it (especially for static scenes) ;)
[This function](https://github.com/kcat/openal-soft/blob/d240077bc59ea457e207418d718ebdc503a0156b/Alc/ALu.c#L276) is called after listener transformation change.

> makeActive() sounds to me like it doesn't need to be called every frame.

Which would be the case for the above suggestion, but for the original idea I agree, of course. I think there is no need to comment on further implementation details, I have a pretty good idea of what this should look like now.

Squareys:
@mosra I'm currently implementing this and noticed that `setClean()` is currently not called on PlayableGroups currently. This should be done every frame. How about an `PlayableGroup::update[All]()` method? Or should the user call `setClean()` on every Playable himself?

Squareys:
I see now what you meant with the singleton thing. I somehow didn't realize that static members in templates obviously wont work. So, updating every frame is a cleaner solution.

The "updating PlayableGroups issue" still holds, though.

Squareys:
Alright, this is *very close to done*! @mosra: for a last cleanup, it would awesome, if you could share an oppinion on the following:

- Are you okay with my current solution for the active listener?
- Naming: `PlayableGroup::update()` or `PlayableGroup::setClean()`.
- Sound transformation makes `Listener2D::setNormalizedRotationAxis()` superfluous, remove it?

I know you currently have close to no time, I will therefore work on #111 in the meanwhile.
Thank you in advance!
Greetings, 
Squareys

mosra:
Sorry again about the silence, another crazy week.

Addressing the questions one by one:

*   Calling `setClean()` on all `Playable`s every frame -- yeah, sorry, I completely forgot that it has to be done to have the positions properly updated and thus my design idea is seriously flawed.

    The camera doesn't call `setClean()` on any of the objects because the `Drawable` instances, unlike `Playable`s, are not caching anything -- the transformation matrix has to be uploaded onto the GPU each frame for each object anyway and the computation is just simple matrix multiplication.

    Two solutions that come to my mind, none of which is ideal:

    -   ~~Update the OpenAL source properties not in `clean()` but in `markDirty()`~~ um, no, that is very costly, because then you would calculate absolute transformation from scratch for every object.
    -   Make the `Listener::clean()` function take a list (or initializer-list) of `PlayableGroup`s. The `clean()` function would then call `setClean()` on every group, which in turn calls `setClean()` on every object. This solution is cleaner, but requires the user to know about each and every playable group that is active in given frame (but at least it's not needed to call `setClean()` on every group/playable manually, which is error-prone even more and very unintuitive).

*   The "active listener" singleton -- currently you have to call both `setActive()` and `update()` if you want to switch to another listener, right? I had an idea about making the "active listener" just an cache of which listener was active the last time. There would be just the `update()` function (okay, the name might not make sense anymore). Instead of calling `setClean()` unconditionally, it would call it only if current "active listener" is different than `this` or if the object is dirty (and then updates the "active listener" pointer to `this`). The `clean()` implementation, if ever called, wouldn't do any additional checking. Instead of base class, the "active listener" could be just an `extern` variable in `Audio::Implementation`, because the only operation done on that is "compare and swap".

    I think it would avoid some unintuitive cases (e.g. `Listener::update()` strangely not doing anything because some code earlier called `setActive()` on some other listener).

    I had this idea a few days ago but did not have time to write it down, so I'm sorry if it would cause a lot of rewriting. As always, I'm open to discussion.

*   > Sound transformation makes `Listener2D::setNormalizedRotationAxis()` superfluous

    Yup, remove it and get rid of that ugly `atan2` :) If I'm seeing correctly, it would also make it possible to merge `Listener2D` and `Listener3D` into a single template?
    
*   > Naming: `PlayableGroup::update()` or `PlayableGroup::setClean()`

    `Shapes::ShapeGroup` has `setClean()` too.
    
Will comment on some more things inline.

src/Magnum/Audio/Listener.h
```diff
@@ -0,0 +1,247 @@
...
+    protected:
+        /** @brief Constructor */
+        explicit AbstractListenerBase() = default;
+
+    private:
+        /* Pointer to the currently active listener. Should never be dereferenced, since may be null at any given time. */
+        static AbstractListenerBase* _activeListener;
+};
+
+AbstractListenerBase* AbstractListenerBase::_activeListener = nullptr;
```
mosra: This should be in some `.cpp` file so it really gets defined only once.

src/Magnum/Audio/Listener.h
```diff
@@ -0,0 +1,247 @@
...
+cinematics for example.
+
+## Sound Transformation
+
+@ref AbstractListener::setSoundTransformation() enables you to set transformation which is applied to the listeners orientation and position
+before passed onto OpenAL. This can be used for rotating two dimensional scenes as planes into three dimensional space or even scaling the audio
+scene to match a certain world scale. In the later case you might want to instead consider @ref Renderer::setSpeedOfSound().
+
+*/
+template <UnsignedInt dimensions, class Derived> class AbstractListener: public SceneGraph::AbstractFeature<dimensions, Float>, public AbstractListenerBase {
```
mosra: I might be blind, but why the `Derived` template parameter is needed?

src/Magnum/Audio/Listener.h
```diff
@@ -0,0 +1,247 @@
...
+        /**
+         * @brief Update the listener
+         *
+         * Calls @ref SceneGraph::AbstractObject::setClean() on its parent object.
+         */
+        void update() { SceneGraph::AbstractFeature<dimensions, Float>::object().setClean(); } // TODO: Only if active?
+
+    private:
+        Matrix3x3 _soundTransformation;
+        Float _gain;
```
mosra: Maybe another dead code, where is `_gain` used?

src/Magnum/Audio/Playable.h
```diff
@@ -0,0 +1,178 @@
...
+           Called in Playable::setGain() and PlayableGroup::setGain() */
+        void cleanGain() {
+            if(_playableGroup) {
+                _source.setGain(_gain*_playableGroup->gain());
+            } else {
+                _source.setGain(_gain);
+            }
+        }
+
+        PlayableGroup<dimensions>* _playableGroup;
```
mosra: Better solution might be to create `playables()` function that just `static_cast`s value returned from `AbstractGroupedFeature::group()` -- i.e., no need to store another variable. See `SceneGraph::Animable::animables()` for an implementation. 

This would break when the user would set the group by calling `AbstractGroupedFeature::setGroup()`.

src/Magnum/Audio/Playable.h
```diff
@@ -0,0 +1,178 @@
...
+        VectorTypeFor<dimensions, Float> _fwd;
+        Float _gain;
+        Source& _source;
```
mosra: Why is the source not part of the feature (i.e., stored by value) and the user has to manage it somewhere separately?

src/Magnum/Audio/PlayableGroup.h
```diff
@@ -0,0 +1,194 @@
...
+         */
+        explicit PlayableGroup(): SceneGraph::FeatureGroup<dimensions, Playable<dimensions>, Float>{} {}
+
+        /**
+         * @brief Play all sound sources in this group
+         * @return Reference to self (for method chaining)
+         */
+        PlayableGroup<dimensions>& playAll() {
+            for(UnsignedInt i = 0; i < this->size(); ++i)
+                (*this)[i].source().play();
```
mosra: Using [Source::play(...)](http://mosra.cz/blog/magnum-doc/classMagnum_1_1Audio_1_1Source.html#ace53ea6c28c35326c3265e029d906d90) will ensure that all sources start playing at the same time. I agree that current API of that function is not ideal, will fix it later.

Similarly for `pause()`/`stop()`/`rewind()` below.

Maybe naming these just `play()` (without the `All` suffix) could be enough?

src/Magnum/Audio/PlayableGroup.h
```diff
@@ -0,0 +1,194 @@
...
+
+            return *this;
+        }
+
+        /**
+         * @brief Clean all the contained Playables.
+         *
+         * @see @ref AbstractObject::setClean()
+         */
+        void setAllClean() {
```
mosra: Also -- just `setClean()`, I think.

src/Magnum/Audio/PlayableGroup.h
```diff
@@ -0,0 +1,194 @@
...
+            for(UnsignedInt i = 0; i < this->size(); ++i)
+                (*this)[i].object().setClean();
+        }
+
+    private:
+        Matrix3x3 _soundTransform;
```
mosra: Why 3x3 and not `Matrix4` along with translation? (someone, especially in 2D scenes, might want some sounds appear "higher"/"lower").

src/Magnum/Audio/PlayableGroup.h
```diff
@@ -0,0 +1,194 @@
...
+        Float _gain;
```
mosra: I think it may be wise to initialize this to some value (`1.0f`?) (and then mentioning that in the `setGain()` docs). Otherwise there would be some random value which can kill your eardrum ;)

src/Magnum/Audio/Listener.h
```diff
@@ -0,0 +1,247 @@
...
+
+There can only be at the most *one* active listener at a given time. This is because OpenAL only supports notion of one listener.
+Having multiple @ref Listener2D or @ref Listener3D instances can still be useful for conveniently switching between them for
+cinematics for example.
+
+## Sound Transformation
+
+@ref AbstractListener::setSoundTransformation() enables you to set transformation which is applied to the listeners orientation and position
+before passed onto OpenAL. This can be used for rotating two dimensional scenes as planes into three dimensional space or even scaling the audio
+scene to match a certain world scale. In the later case you might want to instead consider @ref Renderer::setSpeedOfSound().
```
mosra: Please wrap the docs at column 80. It allows having two editors opened side-by-side (or editor and Git console) and also makes the reading easier.

src/Magnum/Audio/Listener.h
Squareys: Uhm, it's in the return value of `setSoundTransformation()`. I guess that is not necessary?

src/Magnum/Audio/Listener.h
Squareys: I took the getter and setters out for some reason. I guess I must have forgotten it. Not sure why I took it out in the first place, though, to be honest.

src/Magnum/Audio/Playable.h
Squareys: Ah, I see, I was already wondering about that, thanks!

src/Magnum/Audio/Playable.h
Squareys: Well, I never looked at it that way I guess, makes sense, though :)

src/Magnum/Audio/PlayableGroup.h
Squareys: Alright, will do. I missed `Source::play()` for some reason.

src/Magnum/Audio/PlayableGroup.h
Squareys: Actually, you suggested that in an IRC chat ;) I have to admin, I didn't find a use case for translation, though, so I didn't question it.

src/Magnum/Audio/Listener.h
mosra: Oh, right, so it would be for `Audio::Renderer::setListenerGain()`? Useful thing to have :)

src/Magnum/Audio/Listener.h
mosra: Right. Then I think it would make sense to combine all classes (`AbstractListener`, `Listener2D` and `Listener3D`) into one template (along with removal of `setRotationAxis()`). The amount of generated code would be the same as in current case.

mosra:
By the way, the AppVeyor build is fixed on current master and added another one, so if you rebase, you get testing on three different platforms for free :)

I think MSVC 2015 errored out on some uses of constructor calls via `{}`, the solution is to use the "old-style" `()` in affected cases.

Squareys:
Build success, yay! :D

Squareys:
@mosra Regarding the new Listener.hpp file: SceneGraph has a "instantiation.cpp". Should I add one of those for Audio, too?
It sounds a bit like overkill for just Listener. What would be the alternative? Make it a cpp?

src/Magnum/Audio/CMakeLists.txt
```diff
@@ -56,6 +66,10 @@ endif()
 
 target_link_libraries(MagnumAudio Magnum ${CORRADE_PLUGINMANAGER_LIBRARIES} ${OPENAL_LIBRARY})
 
+if(WITH_SCENEGRAPH)
+    target_link_libraries(MagnumAudio ${MAGNUM_SCENEGRAPH_LIBRARIES})
```
mosra: Just a wild guess -- the `${MAGNUM_SCENEGRAPH_LIBRARIES}` variable is probably not defined at all here and thus the AppVeyor build fails -- try `MagnumSceneGraph` instead

Squareys: What a precise wild guess, indeed :D Worked, of course.

Squareys: Thanks! Would have taken me a while ;)

Squareys:
> There would be just the update() function

"just" meaning no `clean()`?

> call it only if current "active listener" is different than this or if the object is dirty

So `this != _activeListener || this->object()->isDirty()`? `setClean()` checks for dirty, though, right? I am a little confused by that condition.

> Make the Listener::clean() function take a list (or initializer-list) of PlayableGroups

How about an `add(/remove)PlayableGroup()` or `(un)registerPlayableGroup()` or whatever? That would remove the requirement of knowing every playable group.

mosra:
> "just" meaning no clean()?

To clarify: the Listener would have public `update(...)` function(taking list of playable groups) and private `clean()` implementation. (I confused the two in some places, sorry about that, hopefully fixed now.)

> setClean() checks for dirty, though, right? I am a little confused by that condition.

If the code would be only `if(this != _activeListener) object().setClean();`, then `setClean()` will be called only in case the active listener changes. If the active listener stays the same and the object just becomes dirty, then the condition evaluates to false and `setClean()` won't be called, which is wrong.

> How about an add(/remove)PlayableGroup() or (un)registerPlayableGroup() or whatever? That would remove the requirement of knowing every playable group.

I assume you mean having these on the listener. That would bring unnecessary complications in my opinion. I imagine the `listener.update({playables, playables2, ...})` (sorry, accidentally named it `clean()` above)  bit will be called at one and only one place in the code (the actual render/update loop) and so it doesn't matter too much. With your proposed approach each Listener has to store all groups it affects, which means additional private member, additional allocations and additional thing that can be forgotten by the user. Even worse, having multiple listeners would mean registering a group to multiple instances (which defeats whole purpose in my opinion).

Hopefully I didn't add more confusion to it :)

Squareys:
@mosra  I am *very* sorry that I still don't fully understand your suggestion. In case we completely talked past each other, some pseudo code for how one would use the Listener could help.

I have the following issue with that condition thing:

**short version:**
The if statement in `if(isDirty() || (this != _activeListener)) { setClean(); }` is superfluous.
Skip the following, if the short version already convinced you.

========
**long version:** *(with hopefully mathematically correct proof)*
We know that if `isDirty()` is `true` and we call `setClean()`, the `clean()` method is called ([see here](https://github.com/mosra/magnum/blob/master/src/Magnum/SceneGraph/Object.hpp#L130)). We can write that as `isDirty && setClean() => clean()`.

Now, you are suggesting `(this != _activeListener) || isDirty => setClean()`, right? Replacing the `setClean()` in the above formula we get:
`isDirty && ((this != _activeListener) || isDirty) => clean()` which with distributivity law can be transformed into `(isDirty && (this != _activeListener)) || (isDirty && isDirty) => clean()`.
Case `isDirty == false`: we get `(false && ?) || false => false` (last `false` because clean is not called). 
Case `isDirty == true`: we get `(true && ?) || true => true`.

Which means that `isDirty => clean()` must hold, `(this != _activeListener)` is therefore superfluous.

As a result the entire test would be superfluous since `setClean()` already tests for `isDirty`.

========
**Conclusion:**
With the `myListener.update({group1, group2, ...})` being called every frame anyway, I don't think there is any use left for `_activeListener`.

I thought about a static method `Listener::activeListener()` to be able to update the cached (lastly updated?) listener, but that could be 2D/3D, which is why we can only return `AbstractListenerBase*` which for the same reason cannot have a `virtual update(?)` method.

So... remove `_activeListener`?

mosra:
oh.

You are right. I am terribly sorry, the condition felt somewhat weird but I didn't give it much thought. Now I know why. I feel braindead :)

(Pseudo)code is sometimes better than words, so here's what I thought:

```cpp
// Listener.h

namespace Implementation {
    // Just a typeless variable is enough
    extern MAGNUM_AUDIO_EXPORT void* activeListener;
}

// No special base class
template<UnsignedInt dimensions> class Listener: public SceneGraph::AbstractFeature<dimensions, Float> {
    public:
        // ...

        // Only one update() overload, no setActive(), no clean()
        void update(std::initializer_list<std::reference_wrapper<PlayableGroup<dimensions>>> groups);

        // ...
};

// Listener.cpp

namespace Implementation {
    void* activeListener = nullptr;
}

template<UnsignedInt dimensions> void Listener<dimensions>::update(std::initializer_list<std::reference_wrapper<PlayableGroup<dimensions>>> groups) {
    // instead of setActive() the following:

    /* Ensure that clean() is called also when switching between (clean) listeners */
    if(this != Implementation::activeListener) {
        Implementation::activeListener = this;
        object().markDirty();
    }

    /* If object transformation or active listener changed, update renderer
       configuration. Not doing that in clean() implementation, because that
       one can be called unconditionally by any other code and thus messing up
       our active listener choice. */
    if(object().isDirty()) {
        // this is the code from clean()
        Renderer::setListenerPosition(soundTransformation().transformVector(Vector3::pad(absoluteTransformationMatrix.translation())));

        const Vector3 fwd = soundTransformation().transformVector(
            padMatrix4(absoluteTransformationMatrix).rotationScaling()*Vector3{0.0, 0.0, 1.0});
        const Vector3 up = soundTransformation().transformVector(
            padMatrix4(absoluteTransformationMatrix).rotationScaling()*Vector3{0.0, 1.0, 0.0});
        Renderer::setListenerOrientation(fwd, up);

        /* Do the rest so it's not dirty anymore */
        object().setClean();
    }

    // cleaning the playable groups like before
}

// Usage:

Listener3D listener{object};

// each frame:
listener.update({groups...});
// or
listener.update({});

// (and nothing else)
```

The active listener pointer is only handling the case where one would switch between (otherwise clean) listeners. We're doing just comparison and writes to it, so it can be just `void*` -- which makes everything simpler as we don't need to inherit from any common type. Originally I wanted to have the `Renderer` updates in the `clean()` implementation, but then I discovered that it might cause random updates of renderer configuration when someone else would attempt to clean the scenegraph.

src/Magnum/Audio/Playable.h
Squareys: Wow, I completely forgot about this. You mean the Playable should create its own source, correct?

mosra: You couldn't switch source with references anyway ;) I think having exactly one source per Playable is more common use case and that's what I would design for.

As you say, users can switch playables if they really need to.

Squareys:
After 20 days, ended up with merely two simple commits with 987 lines added in the end. Waiting for the CIs to finish, and then this is finally **ready to merge**!

Thank you *alot* for all your patience with this and the confusion that was going on. One could pack the discussion of this PR into a entire novell, but we got it figured out in the end, so it was worth it.

Was fun working with you, as always :)

src/Magnum/Audio/Listener.cpp
```diff
@@ -0,0 +1,117 @@
...
+using namespace SceneGraph;
+
+namespace Implementation {
+    void* activeListener = nullptr;
+}
+
+namespace {
+
+/* Create a Matrix4 from a Matrix3/Matrix4 */
+Matrix4 padMatrix4(const Matrix3&);
```
mosra: I think the forward declarations are not needed since this is in anonymous namespace.

src/Magnum/Audio/Listener.cpp
```diff
@@ -0,0 +1,117 @@
...
+Matrix4 padMatrix4(const Matrix4&);
+
+Matrix4 padMatrix4(const Matrix3& m) {
```
mosra: Make these `inline`?

src/Magnum/Audio/Listener.cpp
```diff
@@ -0,0 +1,117 @@
...
+        /* Only clean if this Listener is active */
+        return;
+    }
+
+    Renderer::setListenerPosition(soundTransformation().transformVector(Vector3::pad(absoluteTransformationMatrix.translation())));
+
+    const Vector3 fwd = soundTransformation().transformVector(
+                padMatrix4(absoluteTransformationMatrix).rotationScaling()*Vector3{0.0, 0.0, 1.0});
+    const Vector3 up = soundTransformation().transformVector(
+                padMatrix4(absoluteTransformationMatrix).rotationScaling()*Vector3{0.0, 1.0, 0.0});
```
mosra: Aren't these multiplications the same as the following?
```cpp
const Vector3 fwd = soundTransformation().transformVector(-padMatrix4(absoluteTransformationMatrix).backward())
const Vector3 up = soundTransformation().transformVector(padMatrix4(absoluteTransformationMatrix).up())
```
Not sure about the signs.

src/Magnum/Audio/Listener.h
```diff
@@ -0,0 +1,222 @@
...
+            this->object().setDirty();
+            return *this;
+        }
+
+        /**
+         * @brief Update the listener
+         * @param groups Groups to update
+         *
+         * Calls @ref SceneGraph::AbstractObject::setClean() on its parent
+         * object and all objects of the @ref Playable s in the group.
```
mosra: Maybe mention that this also sets the listener as active and updates renderer configuration.

src/Magnum/Audio/Listener.h
```diff
@@ -0,0 +1,222 @@
...
+        explicit Listener(SceneGraph::AbstractObject<dimensions, Float>& object):
+            SceneGraph::AbstractFeature<dimensions, Float>(object),
+            _gain{1.0f}
+        {
+            SceneGraph::AbstractFeature<dimensions, Float>::setCachedTransformations(SceneGraph::CachedTransformation::Absolute);
+        }
+
+        /**
+         * @brief Sound transformation
+         * @see @ref Listener::setSoundTransformation()
```
mosra: No need to cross-link getters and setters (should be obvious).

mosra:
Thanks for all the hard work (as always).

Commented on a few tiny things inline, other than that I'm happy to merge this.

src/Magnum/Audio/Listener.cpp
Squareys: I believe it should be `fwd = soundTransformation().transformVector(padMatrix4(absoluteTransformationMatrix).backward())`, probably because OpenALs default fwd is `{0.0f, 0.0f, -1.0f}`, and yes, that seems to be the same.

Squareys: So, having `{0.0f, 0.0f, 1.0f}` as fwd was actually a bug.

mosra:
Will merge it today evening. Sorry for the constant delays.

Squareys:
No problem! :)

PR #111, opened 2015-09-10
OpenAL extension support

Squareys:
Hi @mosra, Hi everybody!

here is a work-in-progress pullrequest for anything related to Audio and OpenAL extension support.

TODOs:
- [x] Add support for OpenAL extensions
   - [x] Finish doc
   - [x] Ensure `Format Type` parameter is a useful config setting. (format type and channels are for the loopback extension, will will not be supported for now.)
   - [x] Tests
- [x] Implement `AL_EXT_float32` and `AL_EXT_double`
   - Not really anything to test.
- [x] Add support for the OpenAL soft hrtfs
   - [x] Tests

Greetings, Squareys

### DIFF
--- a/Doxyfile
+++ b/Doxyfile
@@ -277,7 +277,9 @@ ALIASES                = \
     "es_extension2{3}=<a href=\"http://www.khronos.org/registry/gles/extensions/\1/\3.txt\"><tt>\1_\2</tt></a>" \
     "webgl_extension{2}=<a href=\"https://www.khronos.org/registry/webgl/extensions/\1_\2/\"><tt>\1_\2</tt></a>" \
     "fn_al{1}=`al\1()`" \
+    "fn_alc{1}=`alc\1()`" \
     "def_al{1}=`AL_\1`" \
+    "def_alc{1}=`ALC_\1`" \
     "al_extension{2}=<a href=\"http://icculus.org/alextreg/index.php?operation=op_showext&extname=\1_\2\"><tt>AL_\1_\2</tt></a>" \
     "alc_extension{2}=<a href=\"http://icculus.org/alextreg/index.php?operation=op_showext&extname=\1_\2\"><tt>ALC_\1_\2</tt></a>"
 
--- a/Doxyfile
+++ b/Doxyfile
@@ -277,7 +277,9 @@ ALIASES                = \
     "es_extension2{3}=<a href=\"http://www.khronos.org/registry/gles/extensions/\1/\3.txt\"><tt>\1_\2</tt></a>" \
     "webgl_extension{2}=<a href=\"https://www.khronos.org/registry/webgl/extensions/\1_\2/\"><tt>\1_\2</tt></a>" \
     "fn_al{1}=`al\1()`" \
-    "def_al{1}=`AL_\1`"
+    "def_al{1}=`AL_\1`" \
+    "al_extension{2}=<a href=\"http://icculus.org/alextreg/index.php?operation=op_showext&extname=\1_\2\"><tt>AL_\1_\2</tt></a>" \
+    "alc_extension{2}=<a href=\"http://icculus.org/alextreg/index.php?operation=op_showext&extname=\1_\2\"><tt>ALC_\1_\2</tt></a>"
 
 # This tag can be used to specify a number of word-keyword mappings (TCL only).
 # A mapping has the form "name=value". For example adding "class=itcl::class"
--- a/doc/openal-support.dox
+++ b/doc/openal-support.dox
@@ -47,6 +47,7 @@ Extension                                   | Status
 Extension                                   | Status
 ------------------------------------------- | ------
 @alc_extension{SOFTX,HRTF}                  | done
+@alc_extension{SOFT,HRTF}                   | done
 
 
 */
--- a/doc/openal-support.dox
+++ b/doc/openal-support.dox
@@ -1,0 +1,52 @@
+/*
+    This file is part of Magnum.
+
+    Copyright © 2010, 2011, 2012, 2013, 2014, 2015
+              Vladimír Vondruš <mosra@centrum.cz>
+    Copyright © 2015 Jonathan Hale <squareys@googlemail.com>
+
+    Permission is hereby granted, free of charge, to any person obtaining a
+    copy of this software and associated documentation files (the "Software"),
+    to deal in the Software without restriction, including without limitation
+    the rights to use, copy, modify, merge, publish, distribute, sublicense,
+    and/or sell copies of the Software, and to permit persons to whom the
+    Software is furnished to do so, subject to the following conditions:
+
+    The above copyright notice and this permission notice shall be included
+    in all copies or substantial portions of the Software.
+
+    THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+    IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+    FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+    THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+    LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+    FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+    DEALINGS IN THE SOFTWARE.
+*/
+
+/** @page openal-support OpenAL support state
+@brief List of (un)supported OpenAL features and extensions.
+
+@tableofcontents
+
+@section openal-support-state OpenAL implementation state
+
+The extension implementation is considered complete if all its defined types,
+functions and enum values are exposed through the API.
+
+
+@subsection openal-extension-support Extensions
+
+Extension                                   | Status
+------------------------------------------- | ------
+@al_extension{EXT,double}                   | done
+@al_extension{EXT,float32}                  | done
+
+@subsection openal-extension-support OpenAL Soft Extensions
+
+Extension                                   | Status
+------------------------------------------- | ------
+@alc_extension{SOFTX,HRTF}                  | done
+
+
+*/
--- a/src/Magnum/Audio/Buffer.h
+++ b/src/Magnum/Audio/Buffer.h
@@ -31,6 +31,9 @@
 
 #include <utility>
 #include <al.h>
+#include <alc.h>
+#include "MagnumExternal/OpenAL/extensions.h"
+
 #include <Corrade/Containers/ArrayView.h>
 
 #include "Magnum/Magnum.h"
@@ -52,7 +55,21 @@ class Buffer {
             Mono8 = AL_FORMAT_MONO8,        /**< 8-bit unsigned mono */
             Mono16 = AL_FORMAT_MONO16,      /**< 16-bit signed mono */
             Stereo8 = AL_FORMAT_STEREO8,    /**< 8-bit interleaved unsigned stereo */
-            Stereo16 = AL_FORMAT_STEREO16   /**< 16-bit interleaved signed stereo */
+            Stereo16 = AL_FORMAT_STEREO16,   /**< 16-bit interleaved signed stereo */
+
+            /**
+             * @brief 32-bit mono
+             *
+             * @requires_al_extension extension @al_extension{EXT,float32}
+             */
+            MonoFloat32 = AL_FORMAT_MONO_FLOAT32,
+
+            /**
+             * @brief 32-bit interleaved stereo
+             *
+             * @requires_al_extension extension @al_extension{EXT,float32}
+             */
+            StereoFloat32 = AL_FORMAT_STEREO_FLOAT32,
         };
 
         /**
--- a/src/Magnum/Audio/Buffer.h
+++ b/src/Magnum/Audio/Buffer.h
@@ -70,6 +70,20 @@ class Buffer {
              * @requires_al_extension extension @al_extension{EXT,float32}
              */
             StereoFloat32 = AL_FORMAT_STEREO_FLOAT32,
+
+            /**
+             * @brief 64-bit mono
+             *
+             * @requires_al_extension extension @al_extension{EXT,double}
+             */
+            MonoDouble = AL_FORMAT_MONO_DOUBLE_EXT,
+
+            /**
+             * @brief 64-bit interleaved stereo
+             *
+             * @requires_al_extension extension @al_extension{EXT,double}
+             */
+            StereoDouble = AL_FORMAT_STEREO_DOUBLE_EXT,
         };
 
         /**
--- a/src/Magnum/Audio/Context.h
+++ b/src/Magnum/Audio/Context.h
@@ -240,13 +240,20 @@ class MAGNUM_AUDIO_EXPORT Context::Configuration {
             return *this;
         }
 
-        /** @brief Whether to use hrtfs */
+        /**
+         * @brief Whether to use hrtfs
+         * @requires_alc_extension for HRTFs, extension @alc_extension{SOFTX,HRTF}
+         *      or @alc_extension{SOFT,HRTF}
+         */
         EnabledState isHrtfEnabled() const { return _enableHrtf; }
 
         /**
          * @brief Set whether to use hrtfs
          *
          * Defaults to local OpenAL configuration or false.
+         * @requires_alc_extension for HRTFs otherwise setting will be ignored,
+         *      extension @alc_extension{SOFTX,HRTF} or
+         *      @alc_extension{SOFT,HRTF}
          */
         Configuration& setHrtfEnabled(EnabledState hrtf) {
             _enableHrtf = hrtf;
--- a/src/Magnum/Audio/Context.h
+++ b/src/Magnum/Audio/Context.h
@@ -140,7 +140,8 @@ class MAGNUM_AUDIO_EXPORT Context {
          * @ref supportedExtensions(), @ref Extension::extensions() or
          * @ref isExtensionSupported() for alternatives.
          * @see @fn_al{Get} with @def_al{NUM_EXTENSIONS}, @fn_al{GetString}
-         *      with @def_al{EXTENSIONS}
+         *      with @def_al{EXTENSIONS}, @fn_alc{GetString} with
+         *      @def_alc{EXTENSIONS}
          */
         std::vector<std::string> extensionStrings() const;
 
--- a/src/Magnum/Audio/Context.h
+++ b/src/Magnum/Audio/Context.h
@@ -5,6 +5,7 @@
 
     Copyright © 2010, 2011, 2012, 2013, 2014, 2015
               Vladimír Vondruš <mosra@centrum.cz>
+    Copyright © 2015 Jonathan Hale <squareys@googlemail.com>
 
     Permission is hereby granted, free of charge, to any person obtaining a
     copy of this software and associated documentation files (the "Software"),
@@ -30,9 +31,18 @@
  */
 
 #include <string>
+#include <vector>
+#include <array>
+#include <bitset>
 #include <al.h>
 
+#include <Corrade/Containers/EnumSet.h>
+
+#include <Magnum/Magnum.h>
+
 #include "Magnum/Audio/visibility.h"
+#include "Magnum/Audio/Buffer.h"
+
 
 #ifndef DOXYGEN_GENERATING_OUTPUT
 typedef struct ALCdevice_struct ALCdevice;
@@ -40,6 +50,36 @@ typedef struct ALCcontext_struct ALCcontext;
 #endif
 
 namespace Magnum { namespace Audio {
+
+class Context;
+
+/**
+@brief Run-time information about OpenAL extension
+
+Encapsulates runtime information about OpenAL extension, such as name string,
+minimal required OpenAL version and version in which the extension was adopted
+to core.
+
+See also @ref Audio::Extensions namespace, which contain compile-time information
+about OpenAL extensions.
+*/
+class MAGNUM_AUDIO_EXPORT Extension {
+    friend Context;
+
+    public:
+        /** @brief All OpenAL extensions */
+        static const std::vector<Extension>& extensions();
+
+        /** @brief Extension string */
+        constexpr const char* string() const { return _string; }
+
+    private:
+        /* MSVC seems to have problems with const members */
+        std::size_t _index;
+        const char* _string;
+
+        constexpr Extension(std::size_t index, const char* string): _index(index), _string(string) {}
+};
 
 /**
 @brief OpenAL context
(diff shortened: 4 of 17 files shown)


Squareys:
The travis-ci build seems to be failing because of `libopenal-dev_1%3a1.13-4ubuntu3_amd64.deb` being outdated. Current version is 1.16, and it seems alot has happened since 1.13.

Is there anything one can do about that (except for compiling it oneself from scratch)?

LB--:
You can try and update it with the `apt` section:
http://docs.travis-ci.com/user/apt/

Squareys:
@LB-- Hi! `libopenal-dev` *was* in the apt section. But with that I only get version 1.13. Or did I misunderstand? I don't see a way of updating there...

Squareys:
I found a way (see latest commit), but that doesn't help. `libopenal-dev is already the newest version.`

Squareys:
I believe the reason is that it is outdated on the package repository. `wily` has a recent version though.

Squareys:
I decided to go for the "build from source" solution, since I discovered that those builds can be cached. What is left to do is probably tie the git repository to the latest release tag. Or is this not required?

mosra:
Hm... I fear that restricting ourselves to OpenAL Soft will harm portability a lot.

Originally (~2010-2011) I had similar problems with OpenGL -- each platform had its own uniquely outdated GL headers and uniquely buggy GLEW version. It was painful to port and painful to maintain and the only sane solution was to switch to my own copies of GL headers.

I think it might be worth doing similar thing with OpenAL. Would adding our own copy of `alext.h` to `MagnumExternal/OpenAL` help? Or do we need more than just a header?

**Edit**:

```cpp
/**
 * OpenAL cross platform audio library
 * Copyright (C) 2008 by authors.
 * This library is free software; you can redistribute it and/or
 *  modify it under the terms of the GNU Library General Public
 *  License as published by the Free Software Foundation; either
 *  version 2 of the License, or (at your option) any later version.
 *
 * This library is distributed in the hope that it will be useful,
 *  but WITHOUT ANY WARRANTY; without even the implied warranty of
 *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
 *  Library General Public License for more details.
 *
 * You should have received a copy of the GNU Library General Public
 *  License along with this library; if not, write to the
 *  Free Software Foundation, Inc., 59 Temple Place - Suite 330,
 *  Boston, MA  02111-1307, USA.
 * Or go to http://www.gnu.org/copyleft/lgpl.html
 */
```

Ugh. The one from OpenAL Soft is GPL-poisoned, I don't think it is good to have a copy of this in our MIT-licensed code. Anybody who could say whether this is okay or not? 

What about creating our own `MagnumExternal/OpenAL/extensions.h` header that would contain just the `#define`s we need? There is just a few of them. I think that's the safest way.

Squareys:
> What about creating our own MagnumExternal/OpenAL/extensions.h header that would contain just the #defines we need? There is just a few of them. I think that's the safest way.

Agreed, I will see what I can do.

Squareys:
Aw yeah, build success.

I want to note, that for those defines in the `MagnumExternal/OpenAL/extensions.h", I solely used the extension definitions ([AL_EXT_float32](http://icculus.org/alextreg/index.php?operation=op_showext&extname=AL_EXT_float32) and the currently indev [ALC_SOFT_HRTF](https://openal.org/pipermail/openal/2015-October/000436.html)).

Squareys:
Two things left: Tests for HRTF extensions and discussion on how to handle AL extensions versus ALC extensions:

ALC extensions could be queried before context creation, which would make `Context::isExtensionSupported()` viable for use in `Configuration::attributes()`. This would, of course, mean that we would need to loop through two extension string vectors (very similar code) and since unknown attributes passed to `alcCreateContext()` seem to be simply ignored, we probably don't need use `Context::isExtensionSupported()` in `Configuration::attributes()`, anyway. @mosra What is your opinion on this?

src/Magnum/Audio/Context.h
```diff
@@ -41,6 +52,37 @@ typedef struct ALCcontext_struct ALCcontext;
...
+    public:
+        /** @brief All OpenAL extensions */
+        static const std::vector<Extension>& extensions();
+
+        /** @brief Extension string */
+        constexpr const char* string() const { return _string; }
+
+    private:
+        /* GCC 4.6 doesn't like const members, as std::vector doesn't have
+           proper move semantic yet */
```
mosra: Huh. This is a leftover comment from a long ago and doesn't make any sense even in `Magnum::Extension` :)

src/Magnum/Audio/Context.h
```diff
@@ -84,13 +133,221 @@ class MAGNUM_AUDIO_EXPORT Context {
...
+
+By default, if assertion fails, an message is printed to error output and the
+application aborts. If `CORRADE_NO_ASSERT` is defined, this macro does nothing.
+Example usage:
+@code
+MAGNUM_AUDIO_ASSERT_EXTENSION_SUPPORTED(Extensions::ALC::SOFTX::HRTF);
+@endcode
+
+@see @ref Magnum::Audio::Context::isExtensionSupported() "Context::isExtensionSupported()",
+    @ref MAGNUM_ASSERT_VERSION_SUPPORTED(), @ref CORRADE_ASSERT(),
```
mosra: I would make the link title explicitly `"Audio::Context::isExtensionSupported()"` so it doesn't get confused with the same function on `Magnum::Context`. Also please remove the `MAGNUM_ASSERT_VERSION_SUPPORTED()`, it is not relevant to OpenAL.

src/Magnum/Audio/Context.h
```diff
@@ -84,13 +133,221 @@ class MAGNUM_AUDIO_EXPORT Context {
...
+    @ref CORRADE_INTERNAL_ASSERT()
+*/
+#ifdef CORRADE_NO_ASSERT
+#define MAGNUM_AUDIO_ASSERT_EXTENSION_SUPPORTED(extension) do {} while(0)
+#else
+#define MAGNUM_AUDIO_ASSERT_EXTENSION_SUPPORTED(extension)                        \
```
mosra: What about `MAGNUM_ASSERT_AUDIO_EXTENSION_SUPPORTED`? I think the words "flow better" in that case (and we don't need to preserve namespaces so strictly in macros).

src/Magnum/Audio/Context.h
```diff
@@ -84,13 +133,221 @@ class MAGNUM_AUDIO_EXPORT Context {
...
+    do {                                                                    \
+        if(!Magnum::Audio::Context::current()->isExtensionSupported<extension>()) { \
+            Corrade::Utility::Error() << "Magnum: required openal extension" << extension::string() << "is not supported"; \
```
mosra: `... OpenAL extension" << ... ` ;)

src/Magnum/Audio/Extensions.h
```diff
@@ -0,0 +1,94 @@
...
+(@ref Extension::requiredVersion() "requiredVersion()",
+@ref Extension::coreVersion() "coreVersion()" and @ref Extension::string() "string()"),
+but these structs are better suited for compile-time decisions rather than
+@ref Extension instances. See @ref Context::isExtensionSupported() for example
+usage.
+
+This namespace is built by default. To use it, you need to add `${MAGNUM_AUDIO_INCLUDE_DIRS}`
+to include path and link to `${MAGNUM_AUDIO_LIBRARIES}`. See @ref building and
+@ref cmake for more information.
+@see @ref MAGNUM_ASSERT_EXTENSION_SUPPORTED()
```
mosra: `MAGNUM_ASSERT_AUDIO_EXTENSION_SUPPORTED()`. Also, the namespace is not built by default, it should say something similar to [what's here](https://github.com/mosra/magnum/blob/3adc93888b06e3fda5a9f6eee902f87a94d3e604/doc/namespaces.dox#L108-L114).

src/Magnum/Audio/Extensions.h
```diff
@@ -0,0 +1,94 @@
...
+    }
+} namespace ALC {
+    namespace EXT {
+        _extension_rev(ALC,EXT,ENUMERATION) // #???
+    }
+    namespace SOFTX {
+        _extension(ALC,SOFTX,HRTF) // #???
+    }
+}
+#undef _extension
```
mosra: `#undef _extension_rev`

Doxyfile
```diff
@@ -277,7 +277,10 @@ ALIASES                = \
     "es_extension2{3}=<a href=\"http://www.khronos.org/registry/gles/extensions/\1/\3.txt\"><tt>\1_\2</tt></a>" \
     "webgl_extension{2}=<a href=\"https://www.khronos.org/registry/webgl/extensions/\1_\2/\"><tt>\1_\2</tt></a>" \
     "fn_al{1}=`al\1()`" \
-    "def_al{1}=`AL_\1`"
+    "fn_alc{1}=`alc\1()`" \
+    "def_al{1}=`AL_\1`" \
+    "def_alc{1}=`ALC_\1`" \
+    "al_extension{2}=<a href=\"http://icculus.org/alextreg/index.php?operation=op_showext&extname=\1_\2\"><tt>\1_\2</tt></a>"
```
mosra: What about the `\alc_extension` and `\al_eax_extension` Doxygen commands? Could the `\al_eax_extension` be written in terms of `\al_extension` or they are something very special?  If there aren't any publicly available files for these, make it just a inline code block (so we can add the links later if they ever appear on the web):
```
"alc_extension{2}=`ALC_\1`"
```

doc/openal-support.dox
```diff
@@ -0,0 +1,93 @@
...
+@alc_extension{SOFT,pause_device}           | |
+@al_extension{SOFT,block_alignment}         | |
+@al_extension{SOFT,buffer_samples}          | |
+@al_extension{SOFT,buffer_sub_data}         | |
+@al_extension{SOFT,deferred_updates}        | |
+@al_extension{SOFT,direct_channels}         | |
+@al_extension{SOFT,loop_points}             | |
+@al_extension{SOFT,MSADPCM}                 | |
+@al_extension{SOFT,source_latency}          | |
+@al_extension{SOFT,source_length}           | |
```
mosra: I would suggest listing only those that are supported now, otherwise this might look like we commited ourselves to support every one of them, which I think is not the case.

With OpenGL I'm listing all extensions that I'm willing to/able to implement -- i.e. all extensions that are part of any core version, all core ARB/KHR extensions and some (but not all) selected vendor extensions on top of these.

src/Magnum/Audio/Buffer.h
```diff
@@ -52,7 +55,13 @@ class Buffer {
...
             Mono16 = AL_FORMAT_MONO16,      /**< 16-bit signed mono */
             Stereo8 = AL_FORMAT_STEREO8,    /**< 8-bit interleaved unsigned stereo */
-            Stereo16 = AL_FORMAT_STEREO16   /**< 16-bit interleaved signed stereo */
+            Stereo16 = AL_FORMAT_STEREO16,   /**< 16-bit interleaved signed stereo */
+
+            MonoFloat32 = AL_FORMAT_MONO_FLOAT32, /**< @al_extension{AL_EXT,float32} */
+            StereoFloat32 = AL_FORMAT_STEREO_FLOAT32, /**< @al_extension{AL_EXT,float32}, 32-bit interleaved float */
+
+            MonoDouble = AL_FORMAT_MONO_DOUBLE_EXT, /**< @al_extension{AL_EXT,double}, 64-bit interleaved double */
+            StereoDouble = AL_FORMAT_STEREO_DOUBLE_EXT, /**< @al_extension{AL_EXT,double}, 64-bit interleaved double */
```
mosra: Similarly to OpenGL, there should be a new Doxygen page and command `@requires_al_extension` that functions as an index for all this non-core functionality. 

I can add that later, so you don't have to bother.

src/Magnum/Audio/Context.h
```diff
@@ -30,9 +31,19 @@
...
+#include <bitset>
 #include <al.h>
 
+#include <Corrade/Containers/EnumSet.h>
+
+#include <Magnum/Magnum.h>
+
 #include "Magnum/Audio/visibility.h"
+#include "Magnum/Audio/Buffer.h"
+#include "MagnumExternal/Optional/optional.hpp"
```
mosra: I guess we don't need the `optional` here?

mosra:
Are all parameters key-value pairs? If yes (and I guess so), then it easy for the driver to ignore the unknown ones (sees an unknown key, so it ignores also its corresponding value) and we don't need to bother about checking for extension support in context creation phase.

I'm not checking extension support in GL context creation either, but the `Renderer` class is providing getters to check whether given functionality is really present (e.g. debug output). Does OpenAL have similar query for HRTF presence? If yes, then I think it would be good to have such query in `Audio::Renderer`.

I must say that I don't really like the private `Configuration::attributes()` method, because it does an unnecessary allocation, needs to friend the `Context` class and I think this kind of "low-level" stuff should be done in Context itself using fixed-size `int[]` array with enough padding at the end (like for example [in here](https://github.com/mosra/magnum/blob/3adc93888b06e3fda5a9f6eee902f87a94d3e604/src/Magnum/Platform/Implementation/GlxContextHandler.cpp#L78-L106)).

src/Magnum/Audio/Context.h
Squareys: [x] Done

src/Magnum/Audio/Context.h
Squareys: Since it is the MagnumAudio library, I thought that might be a good idea. `MAGNUM_ASSERT_AUDIO_EXTENSION_SUPPORTED` works too, though.

- [x] Done

src/Magnum/Audio/Context.h
Squareys: - [x] Done

src/Magnum/Audio/Extensions.h
Squareys: - [x] Done

src/Magnum/Audio/Context.h
Squareys: I don't even know why I added it in the first place.
- [x] Done

src/Magnum/Audio/Extensions.h
Squareys: - [x] Done

Doxyfile
Squareys: Looks like I used my own command incorrectly :D
Here's what I suggest I do: 
 - add alc_extension
 - remove all the eax, since I will not implement them any time soon.

doc/openal-support.dox
Squareys: Alright, removed all but the three that are implemented.

Squareys:
> Are all parameters key-value pairs?

Yes, they are. I will not split the ALC/AL extension detection then.

> Does OpenAL have similar query for HRTF presence?

Yes, and providing a query in `Audio::Renderer` makes indeed a lot of sense.

> I must say that I don't really like the private Configuration::attributes() method, because it does an unnecessary allocation, needs to friend the Context class and I think this kind of "low-level" stuff should be done in Context itself [...]

Will do :)

> Other than that, thanks for another awesome contribution and sorry about my lack of time and picky comments :)

As always, the pickier the better. I learn a lot, and Magnum code stays clean.

src/Magnum/Audio/Context.h
```diff
@@ -41,6 +51,35 @@ typedef struct ALCcontext_struct ALCcontext;
...
+    public:
+        /** @brief All OpenAL extensions */
+        static const std::vector<Extension>& extensions();
+
+        /** @brief Extension string */
+        constexpr const char* string() const { return _string; }
+
+    private:
+        const std::size_t _index;
+        const char* _string;
```
mosra: Hmm. I think the `const`s are breaking the AppVeyor build. I guess MSVC requires the type to be move-/copy-assignable to be stored in `std::vector`.

Squareys: Alright, thanks, I was wondering, what MSVC wanted to tell me, I was kinda guessing at this, but I didn't have time to try yet. I'll probably have some tonight.

Squareys: Yeah, that was it. Would explicitly providing an implementation of the copy assignment operator and friends fix it aswell?

Squareys:
> I must say that I don't really like the private Configuration::attributes() method, because it does an unnecessary allocation, needs to friend the Context class and I think this kind of "low-level" stuff should be done in Context itself [...]

Looking at this code again, the reason why I had that `attributes` method in `Configuration` was because I have those `_changed` flags. This was to be able to possibly not provide values such as the hint for how many stereo/mono sources to support or whether to use hrtf, which would then fall back to defaults from the driver or in hrtf case fall back to the user configuration `alsoft.ini`.

Any opinion on how to do this in Context/without `friend`?

mosra:
Yeah, sorry, completely overlooked that. My idea:

Currently the `_changed` flags are just a one-way setting -- the user can "enable" them but there's no way to revert them to the original values. I know there's no such API yet (and it's probably not needed yet), but imagine if there was something similar to `Platform::*Application::tryCreateContext()` -- to these functions you can pass a configuration with e.g. 16x MSAA and if that fails, reset the setting back to default and try again.

So my idea is to have some "initial/default" values instead of the `_changed` flags:
- It could be `-1` for the `Int` values like frequency, sample count etc. I guess `-1` isn't valid value for any of these.
- And for the HRTF (and other boolean values) I propose having tri-state enum with `Default = 0`, `Enable` and `Disable`.

That way the user can always reset the setting to the initial value and you don't need to maintain any additional private state.

doc/openal-support.dox
Squareys: - [x] Done (quite a while ago)

Doxyfile
Squareys: - [x] Done (quite a while ago)

Squareys: - [x] Done (quite a while ago)

Squareys:
@mosra Have a look at https://github.com/Squareys/magnum/commit/43a64b3a04fc2dc363c21cbfeb5510e2f4923a1f, I think that should be alot closer to what you suggested now. Don't hesitate to point out further flaws ;)

src/Magnum/Audio/Buffer.h
Squareys: I'll change it to use the command, though. I will need that in a few other places.

src/Magnum/Audio/Context.h
```diff
@@ -84,13 +131,215 @@ class MAGNUM_AUDIO_EXPORT Context {
...
+         * Returns `-1`, if no hint was set.
+         */
+        Int stereoSourcesCount() const { return _stereoSources; }
+
+        /**
+         * @brief Set hint for how stereo sources to support
+         *
+         * If not set, no hint will given to OpenAL.
+         */
+        Configuration& setStereoSourcesCount(Int sources) {
```
Squareys: @mosra Add `const` (and in other Configuration method parameters)?

src/Magnum/Audio/Renderer.h
```diff
@@ -265,6 +333,9 @@ MAGNUM_AUDIO_EXPORT Debug& operator<<(Debug& debug, Renderer::Error value);
 /** @debugoperatorclassenum{Magnum::Audio::Renderer,Magnum::Audio::Renderer::DistanceModel} */
 MAGNUM_AUDIO_EXPORT Debug& operator<<(Debug& debug, Renderer::DistanceModel value);
 
+/** @debugoperatorclassenum{Magnum::Audio::Renderer,Magnum::Audio::Renderer::HrtfStatus} */
+MAGNUM_AUDIO_EXPORT Debug& operator<<(Debug& debug, Renderer::HrtfStatus value); // TODO and TESTS!
```
mosra: We have tests now, right? :)

Squareys: Oh, yeah, right :P

src/Magnum/Audio/Buffer.h
```diff
@@ -52,7 +55,35 @@ class Buffer {
...
+             * @requires_al_extension extension @al_extension{EXT,float32}
+             */
+            MonoFloat32 = AL_FORMAT_MONO_FLOAT32,
+
+            /**
+             * @brief 32-bit interleaved stereo
+             *
+             * @requires_al_extension extension @al_extension{EXT,float32}
+             */
+            StereoFloat32 = AL_FORMAT_STEREO_FLOAT32, /**< @requires_al_extension{EXT,float32}, 32-bit interleaved float */
```
mosra: Why this has two separate documentation comments?

Squareys: Well, I forgot to remove the old one.

src/Magnum/Audio/Context.h
```diff
@@ -49,12 +89,19 @@ class MAGNUM_AUDIO_EXPORT Context {
...
+        class Configuration;
+
         /**
          * @brief Constructor
          *
-         * Creates OpenAL context.
+         * Creates OpenAL context with given configuration.
          */
+        #ifdef DOXYGEN_GENERATING_OUTPUT
+        explicit Context(const Arguments& arguments, const Configuration& configuration = Configuration());
```
mosra: This should be just `explicit Context(const Configuration& configuration = Configuration());`, I think

Squareys: huh.... bad copy paste error

src/Magnum/Audio/Context.h
mosra: In `*.cpp` files yes, but not in the headers, it clutters the documentation.

src/Magnum/Audio/Context.h
```diff
@@ -84,13 +131,222 @@ class MAGNUM_AUDIO_EXPORT Context {
...
+        Int _monoSources;
+        Int _stereoSources;
+
+        Int _refreshRate;
 };
 
+
+/** @hideinitializer
+@brief Assert that given OpenAL extension is supported
+@param extension    Extension name (from @ref Magnum::Audio::Extensions "Extensions"
```
mosra: `... @ref Magnum::Audio::Extensions "Audio::Extensions"`

src/Magnum/Audio/Renderer.h
```diff
@@ -58,6 +61,37 @@ class Renderer {
...
+             * user configuration that forces HRTF to be used.
+             */
+            Required = ALC_HRTF_REQUIRED_SOFT,
+            /**
+             * @brief HRTF is enabled automatically because the device reported
+             * headphones.
+             */
+            Detected = ALC_HRTF_HEADPHONES_DETECTED_SOFT,
+            /**
+             * @brief The device does not support HRTF with the current format
```
mosra: Please don't use `@brief` in enum value docblocks, it does weird things to the generated documentation ... so just the sentence with a period at the end.

Sorry again for nitpicking :)

src/Magnum/Audio/Context.h
Squareys: Alright, I remembered that, but I wasn't sure whether this still holds when the implementation is in the header. Thank you for telling me again ;)

src/Magnum/Audio/Renderer.h
Squareys: No problem, that is totally reasonable.

mosra:
First-class work. Is it okay if I merge this not now, but tomorrow or the day after? With documentation upload and all that stuff :)

Squareys:
@mosra No problem! I will rebase this by tomorrow evening. Just a reminder: You volunteered to add the `@requires_al_extension` and `@requires_alc_extension` doxygen macros. If time is short on your hand, just tell me, and I will do that ;)

Also, as always, don't hesitate to point out further flaws in the code, if you come across any, even if you see them as nitpicky.

mosra:
Yes, I know, will add them after the merge.

Squareys:
Amazing, the first commit for this pull request was almost 2 months ago. I rebased and cleaned out a few things in the commit history, should be finally good to go! (Whenever you find the time :) )

Squareys:
@mosra I added a minor fix for documentation I just came across. I felt opening another pull request for that would be overkill.

mosra:
Merged, finally. Thanks a lot for the hard work and sorry for the delays.

Squareys:
> sorry for the delays.

No problem, there were delays on my side aswell ;) Thank you for merging!

PR #114, opened 2015-10-08
Dual Quternion Screw Linear Interpolation

Squareys:
Hi @mosra !

Late, but finally, a pullrequest for the ScLERP implementation I worked on the last few days. A few things that may be important:

- I based my work upon these papers:
   - [Dual Quternions for Rigid Transformation Blending](http://dcgi.felk.cvut.cz/home/zara/papers/TCD-CS-2006-46.pdf), Kavan et al., 2006
   - [Dual Quaternions](http://www.xbdev.net/misc_demos/demos/dual_quaternions_beyond/paper.pdf), Kenweight, 2012
- The test values were computed using [this javascript implementation](http://laht.info/WebGL/test.html), the code also served as reference.
- ~~The PR contains an "Optimize" commit, which makes the code a little harder to read. I could rework that, if that is too much of a tradeoff.~~ I added a "inline comments" commit, which helps connect portions of the formula with the code. Although reading TeX code in the comments may not help alot, I think it is enough to justify leaving the optimize commit as is.
- ~~The TeX code has only been partially tested. Also it's a mess in general and the resulting formulas could probably be improved in readability by defining vars for dual/real parts of those dual variables.~~ Now successfully compiled in an external tool and cleaned up the formulas not only for readability, but also found some missing/wrong `\hat`s.

TODOs:
 - [x] Remove the TODO comment which was not supposed to be in the commits ever.
 - [x] Remove the angle method which was added but never used

I'm really interested to read what you think of this!
Greetings, 
Squareys

### DIFF
--- a/src/Magnum/Math/DualQuaternion.h
+++ b/src/Magnum/Math/DualQuaternion.h
@@ -29,6 +29,8 @@
  * @brief Class @ref Magnum::Math::DualQuaternion
  */
 
+#include <cmath>
+
 #include "Magnum/Math/Dual.h"
 #include "Magnum/Math/Matrix4.h"
 #include "Magnum/Math/Quaternion.h"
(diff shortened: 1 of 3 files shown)


Squareys:
@mosra Alright, I revised everything related to the concerns I originally had. Are those comments containing TeX code okay?

src/Magnum/Math/DualQuaternion.h
```diff
@@ -39,6 +41,62 @@ namespace Implementation {
...
+    @ref lerp(const T&, const T&, U)
+ */
+template<class T> inline DualQuaternion<T> sclerp(const DualQuaternion<T>& from, const DualQuaternion<T>& to, const T t) {
+    CORRADE_ASSERT(from.real().isNormalized() && to.real().isNormalized(),
+        "Math::sclerp(): real parts of quaternions must be normalized", {});
+
+    const T dotResult = dot(from.real(), to.real());
+    /* Multiplying with -1.0f ensures shortest path when dot < 0. */
+    /* diff = \hat d = p^*q */
+    const DualQuaternion<T> diff = from.quaternionConjugated()*((dotResult < .0) ? to*-1.0f : to);
```
mosra: Please use `T(0)`, `T(-1)` etc. to avoid implicit float/double/... conversions.

src/Magnum/Math/DualQuaternion.h
```diff
@@ -39,6 +41,62 @@ namespace Implementation {
...
+
+    /* angle = t \hat a_D */
+    const T angle = acos(diff.real().scalar())*t;
+    /* precompute sin/cos for manifold use */
+    const T sinAngle = sin(angle);
+    const T cosAngle = cos(angle);
```
mosra: Please use `std::` prefix to avoid accidental use of the (`double` only) C functions. (`std::acos`, `std::sin`, `std::cos`).

src/Magnum/Math/DualQuaternion.h
```diff
@@ -39,6 +41,62 @@ namespace Implementation {
...
+
+    /* merely a shortcut */
+    const Vector3<T>& m = diff.real().vector();
+    /* invr = \frac 1 {|l_V|} */
+    const T invr = 1/m.length();
```
mosra: There's `m.lengthInverted()` too :)

Squareys: nice :D

src/Magnum/Math/DualQuaternion.h
```diff
@@ -39,6 +41,62 @@ namespace Implementation {
...
+
+    /* Vector of real part of q_{ScLERP}
+       = \hat {\boldsymbol n}_R \cdot sin \left( t \frac {\hat a_R} 2 \right)*/
+    const Vector3<T> v = direction*sinAngle;
+
+    /* pitch = \frac {\hat a_D} 2 = m_S \frac 1 {|l_V|} */
+    T pitch = -diff.dual().scalar()*invr;
+    /* moment = \hat {\boldsymbol n}_D */
+    const Vector3<T> moment = (diff.dual().vector() - (direction*(pitch*diff.real().scalar())))*invr;
+    pitch *= t;
```
mosra: Could this be e.g. `const T pitchT = pitch*t` (and the original `const T pitch`)? I don't like mutating variables in math code :)

Squareys: Makes total sense! I got confused by this while "optimizing".

mosra:
Great stuff, as always.

I'm not really sure about the inline comments because, for me, TeX is write-only :) but I don't know how to do it better. Maybe just use "classic" inline notation (e.g. `pitch = a_D/2 = m_S/length(l_V)`)?

It would be great if the multiplication/division had unit tests too.

Squareys:
Alright, will do :)

Squareys:
Looks like the multiplication was already implemented via the Dual class.

Squareys:
@mosra And done! This is once again **ready for merge**.

*Good call on the multiplication test, btw!* I managed to implement it as doing nothing, because I expected `.dual()` and `.real()` to return references. And that cascaded into another bug in `sclerp()`, which is fixed now, too.

Squareys:
@mosra: This PR is currently waiting for your implementation of Dual/Scalar multiplication in Dual, right? Or did I forget to do something?

mosra:
Yes. I'm currently on the way to Switzerland, so I haven't got any time to finish it. I'll do the implementation during the week after things settle down a bit :)

Squareys:
@mosra No problem, just wanted to know where we're at, in case I was supposed to be doing something here ;) Have a good move!

Squareys:
@mosra Alright! Thanks for the trick with `-to <=> T(-1.0)*to` :)

Just need to wait until the CI finished, then this can be merged.
Thanks again!

mosra:
Finally merged this. Rebased onto current master, hope that's okay. I modified the math formulas to be consistent with conventions in the rest of math code.

While trying to understand the code I managed to simplify it a bit (mainly using introduction of `Math::sincos(Dual)`), it's now basically the same as the math formula in docs.

PR #116, opened 2015-10-18
Fixes to recent PlayableGroup and Listener

Squareys:
Hi @mosra !

I just started using my recent contribution to MagnumAudio myself and immediately found two bugs (very bad omen): Listener would crash if passed a empty playable group on update and PlayableGroup.h produces some weird compile errors when Playable.h is not included aswell.

This PR provides fixes for both issues.

Greetings, Squareys


mosra:
Merged. Thanks!

PR #121, opened 2015-11-15
Audio: Hrtf Extension Support Fixes

Squareys:
Hi @mosra !

As promised on https://github.com/mosra/magnum/commit/fd624d605e417317976ce5d8aa44b1dca0013f58, here are the changed required for fixing the `hrtfStatus()` and `isHrtfEnabled()` methods, which have now moved to `Audio::Context`.

I also added a method to query the name of the hrtf in use.

Greetings,
Squareys

### DIFF
--- a/src/Magnum/Audio/Context.cpp
+++ b/src/Magnum/Audio/Context.cpp
@@ -53,6 +53,21 @@ const std::vector<Extension>& Extension::extensions() {
     return extensions;
 }
 
+Debug& operator<<(Debug& debug, const Context::HrtfStatus value) {
+    switch(value) {
+        #define _c(value) case Context::HrtfStatus::value: return debug << "Audio::Context::HrtfStatus::" #value;
+        _c(Disabled)
+        _c(Enabled)
+        _c(Denied)
+        _c(Required)
+        _c(Detected)
+        _c(UnsupportedFormat)
+        #undef _c
+    }
+
+    return debug << "Audio::Context::HrtfStatus::(invalid)";
+}
+
 Context* Context::_current = nullptr;
 
 Context::Context(): Context{Configuration{}} {}
--- a/src/Magnum/Audio/Context.h
+++ b/src/Magnum/Audio/Context.h
@@ -40,6 +40,7 @@
 
 #include "Magnum/Audio/Audio.h"
 #include "Magnum/Audio/Buffer.h"
+#include "Magnum/Audio/Extensions.h"
 #include "Magnum/Audio/visibility.h"
 
 #ifndef DOXYGEN_GENERATING_OUTPUT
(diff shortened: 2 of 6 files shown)


src/Magnum/Audio/Context.cpp
```diff
@@ -53,6 +53,21 @@ const std::vector<Extension>& Extension::extensions() {
...
+        _c(Disabled)
+        _c(Enabled)
+        _c(Denied)
+        _c(Required)
+        _c(Detected)
+        _c(UnsupportedFormat)
+        #undef _c
+    }
+
+    return debug << "Audio::Renderer::HrtfStatus::(invalid)";
```
Squareys: Woups.

src/Magnum/Audio/Test/ContextTest.cpp
```diff
@@ -56,6 +60,18 @@ void ContextTest::isExtensionEnabled() {
...
 
+void ContextTest::hrtfStatus() {
+    std::ostringstream out;
+    Debug(&out) << Context::HrtfStatus::Denied;
+    CORRADE_COMPARE(out.str(), "Audio::Context::HrtfStatus::Denied\n");
+}
+
+void ContextTest::hrtfs() {
+    /* At least test that the hrtf status is `Disabled` */
+    CORRADE_VERIFY(!_context.isHrtfEnabled());
```
Squareys: I am having second thoughs on this, since this test relies on user configuration. This will work for the CIs, but I have hrtfs enabled for example, which is why tests fail. I believe this test should be removed. @mosra Would you agree?

src/Magnum/Audio/Context.h
```diff
@@ -107,6 +152,42 @@ class MAGNUM_AUDIO_EXPORT Context {
...
+        /**
+         * @brief Hrtf specifier
+         *
+         * Name of the hrtf being used.
+         *
+         * @see @fn_al{GetString} with @def_alc{HRTF_SPECIFIER_SOFT}
+         * @requires_al_extension @alc_extension{SOFT,HRTF}
+         */
+        std::string hrtfSpecifier() const {
+            if(!Context::current()->isExtensionSupported<Extensions::ALC::SOFT::HRTF>())
```
mosra: Personally I would omit this check. It's user responsibility to call only functions that are supported by the driver (and that fact is noted by the `@requires_al_extension` line) and OpenGL APIs behave the same (crash/whatever if you call something that is not supported by the driver).

src/Magnum/Audio/Test/ContextTest.cpp
mosra: Yes, I think so too.

src/Magnum/Audio/Context.h
Squareys: Alright, makes sense.

Squareys:
@mosra Done. `:)`

mosra:
Merged. Thanks a lot!

Issue #117, opened 2015-11-03
Quaternion::slerp returns invalid Quaternion given two identity Quaternions


wivlaro:
Magnum::Quaternion::slerp(Magnum::Quaternion(),Magnum::Quaternion(),0.25f)

Gives me a quaternion of not-a-numbers. I think it should probably return a unit quaternion again.

Squareys:
That is because the angle between those quaternions is `0.0`, and in `Math::sclerp`, we divide by `sin(angle) = 0.0` which results in the `Quaternion({nan, nan, nan}, nan)`.

I guess one simply needs to check if the angle is `0.0` in which case one can simply return the first quaternion. Should be simple, I will create a PR for this.

Squareys:
Actually, since the quaternions are normalized anyway and (correct me, if I'm wrong) there is always only one specific unambiguous quaternion to describe a certain rotation, the only case where `angle == 0.0` is when the two quaternions is equal. That should be cheeper than calculating `angle = std::acos(dot(normalizedA, normalizedB))`, right, @mosra ?

mosra:
I *think* that one rotation can be described with two different quaternions, but I'm not sure. In that case a simple equality comparison would not be enough, but we can surely do a lot better than calling `std::acos()`. Doing something like `TypeTraits<T>::equals(dot(normalizedA, normalizedB), T(1))` might be equivalent, I think? I need to check the math or the properties of dot product.

Oh... and couldn't similar problem affect also the dual quaternion `sclerp()`?

I'm completely braindead now, will think about this more deeply tomorrow.

Squareys:
Alright, you are correct, indeed: [Wikipedia](https://en.wikipedia.org/wiki/Rotation_group_SO(3)#Quaternions_of_unit_norm) says, that exactly two quaternions `q` and `-q` map to the same rotation matrix.

Which means, we can also check for equality with `q` and `-q`, in case that is cheeper than what you suggested.

Squareys:
Oh, and yes, `sclerp()` is also affected by *division by zero*, but only for cases in which the difference of the interpolated dual quaternions is a dual quaternion with a vector `v` with `|v| = 0.0` in the real part. That is because we need to use `.lengthInverted()`. I'm not sure when that is the case though, or how that would be interpreted geometrically.

Squareys:
Okay, I found this: http://www.euclideanspace.com/maths/algebra/realNormedAlgebra/quaternions/slerp/ which suggests a good solution.

[...]

PR #122, opened 2015-11-18
Math: Fix sclerp of dual quaternions with equal rotation and add test.

Squareys:
Hi @mosra !

For previous discussion, see #117 :) This pull request should finally close that issue.

Greetings, 
Squareys


mosra:
Awesome, thanks for being so thorough. I wanted to avoid having math headers dependent on `Functions.h`, so I'll need to invent some other clean way to do this, then I'll merge it.

Squareys:
@mosra A clean way would be to allow this kind of quaternion tranlation lerp (or whatever you want to call it) as a function in Quaternion.h :) If you do that on master, I can adapt this branch for you.

mosra:
Did a different (uglier?) alternative. Thanks a lot for the fix!

PR #126, opened 2015-12-10
On OSX, add target library path to RPATH

staticfloat:
I had problems on OSX when installing to a non-standard `CMAKE_INSTALL_PREFIX` (Things like `corrade-rc` would crash because it couldn't load .`libCorradeUtility.0.dylib`)  This fixes the problems on my computer, but I'm sure it requires testing on other platforms.  To trigger the problem on your computer, try installing `corrade` and then `magnum` using `-DCMAKE_INSTALL_PREFIX=$(echo ~)/local`.

It's possible that this needs to be done to more projects than just magnum and corrade, but I have not tested.


mosra:
Hello, thanks for the pull request!

This issue has been discussed several times, so let me elaborate. The problem can be divided into two separate things:
- For MacOS X one needs to explicitly enable `CMAKE_MACOSX_RPATH` to make proper RPATH handling work. The property was added in 2.8.12 (but default to `OFF`) and since 3.0 it defaults to `ON`. As I'm depending on 2.8.9 rather than 3.0, I additionally need to switch some policies to make it so. That is fixed in afd4ca73e706ef75c35a7023e3f9f81b1f45e519 in Magnum and similar commit is also in all other repositories, so you don't need to worry about `CMAKE_MACOSX_RPATH` anymore.
- The only other thing you need to set is the `CMAKE_INSTALL_RPATH` variable. I'm intentionally *not* doing that by default, because it may not be the desired behavior -- see #55 for in-depth discussion about this. [It's also mentioned in the documentation](https://github.com/mosra/magnum/blob/fb8755b0c4f886e836f90a5935f968a64b806f16/doc/building.dox#L86-L88). 

I'm not sure what the `CMAKE_INSTALL_RPATH_USE_LINK_PATH` variable does, it did not help with anything for me.

To conclude, if you install to non-standard prefix, now you *just* need to set proper `CMAKE_INSTALL_RPATH`. I also verified this in recent Travis builds that it really works [without doing anything else](https://github.com/mosra/magnum/blob/fb8755b0c4f886e836f90a5935f968a64b806f16/package/ci/travis.yml#L47).

staticfloat:
Excellent, thank you for the quick and thorough response, @mosra.  Now that the policy is set by default, I will just pass `-D CMAKE_INSTALL_RPATH=...` for the time being.

I still think that setting the RPATH for `corrade` and `magnum` executables by default is the correct behavior.  I don't think it is good that a broken installation is possible just by setting `-DCMAKE_INSTALL_PREFIX=...`.  Because OSX primarily uses `.app` bundles, RPATH-based linking is the standard method of discovering libraries shipped with your application, however in the `.app` case, the RPATH is often set relative to the loading executable.  This doesn't quite apply here because the binaries I'm concerned with (such as `corrade-rc`) aren't really meant to be redistributed, I don't think, so relative vs. fully-specified RPATHS don't matter unless you are planning to move the corrade/magnum installation around on your local machine (not really a use case I want to bother myself with).

The actual application loading magnum libraries or corrade libraries will, of course, have full control over its own RPATH and linking abilities, and so does not need to worry itself over the details of magnum/corrade's internal linking.  I, personally, will always use relative RPATHs to ensure that my apps are relocatable, but that's beside the point.

I can attest that any slowdown due to searching extra library paths is unmeasurable.  What other concerns about automatically setting `CMAKE_MACOSX_RPATH` to `MAGNUM_LIBRARY_INSTALL_DIR` do you have?  I am certain we can find a solution that fulfills all requirements you have, and makes this process automatic, as many other opensource projects that I have worked on (such as the [Julia Language](https://github.com/JuliaLang/julia), the [Homebrew packaging system](https://github.com/Homebrew/homebrew), etc...) all have some kind of automatic RPATH handling that "just works".

PR #133, opened 2016-02-05
Ability to create EGL contexts on headless machines

mosra:
According to NVidia blog post at https://devblogs.nvidia.com/parallelforall/egl-eye-opengl-visualization-without-x-server/, since NVidia 355 it's possible to create OpenGL and OpenGL ES contexts without having an X server running. That's awesome if you have a headless server somewhere and you want to use it for doing computations or just run GL tests on a build server.

The functionality is now exposed through `Platform::WindowlessEglApplication` and if you enable the `TARGET_HEADLESS` CMake option, all command-line utilities and also GL tests will use this one instead of platform-specific CGL/GLX/WGL APIs.

HOWEVER, I got it running only for OpenGL ES (which they say is supported *since forever*), trying to create desktop OpenGL context doesn't emit *any* error message and then fails on this:

> Context: unsupported OpenGL version (0, 0)

Which means that all the GL functions are somehow no-op. I'm clueless, from the blog post it looks like quite a few companies are already using this functionality so I must be doing something wrong. If someone has an idea how to fix it, that would be very appreciated.


mosra:
Appears to be working with NVidia 361.28, merged.

PR #134, opened 2016-02-07
Fix framebuffer binding on WebGL and GLES

smspillaz:
Framebuffer binding on WebGL was raising `GL_INVALID_ENUM` because `0` was being passed to `glBindFramebuffer`, `glCheckFramebufferStatus` and friends.

The problem mainly appears to be in `AbstractFramebuffer::FramebufferTarget`, as there was no support for the old `GL_FRAMEBUFFER` binding (or at least, an assumption that the enum value was always `1`).

WebGL builds also assumed that `GLES2` style dynamic framebuffer target picking were not supported and also defaulted to using the multi-binding code, which was unsupported.

These changes set the `Read` and `Draw` bindings in `FramebufferTarget` to `GL_FRAMEBUFFER` for compatibility with GLES and WebGL. They also make WebGL take the `bindImplementationSingle` path as that is the only sensible path for it to take.

### DIFF
--- a/src/Magnum/AbstractFramebuffer.h
+++ b/src/Magnum/AbstractFramebuffer.h
@@ -117,19 +117,19 @@ enum class FramebufferTarget: GLenum {
     /** Frambebuffer reading target */
     #ifndef MAGNUM_TARGET_GLES2
     Read = GL_READ_FRAMEBUFFER,
-    #elif !defined(MAGNUM_TARGET_WEBGL)
+    #elif defined(CORRADE_TARGET_APPLE)
     Read = GL_READ_FRAMEBUFFER_APPLE,
     #else
-    Read,
+    Read = GL_FRAMEBUFFER,
     #endif
 
     /** Framebuffer drawing target */
     #ifndef MAGNUM_TARGET_GLES2
     Draw = GL_DRAW_FRAMEBUFFER,
-    #elif !defined(MAGNUM_TARGET_WEBGL)
+    #elif defined(CORRADE_TARGET_APPLE)
     Draw = GL_DRAW_FRAMEBUFFER_APPLE,
     #else
-    Draw,
+    Draw = GL_FRAMEBUFFER,
     #endif
 
     #ifdef MAGNUM_BUILD_DEPRECATED
@@ -140,10 +140,10 @@ enum class FramebufferTarget: GLenum {
     ReadDraw CORRADE_DEPRECATED_ENUM("use FramebufferTarget::Draw instead") =
         #ifndef MAGNUM_TARGET_GLES2
         GL_DRAW_FRAMEBUFFER
-        #elif !defined(MAGNUM_TARGET_WEBGL)
+        #elif defined(MAGNUM_TARGET_APPLE)
         GL_DRAW_FRAMEBUFFER_APPLE
         #else
-        1
+        GL_FRAMEBUFFER
         #endif
     #endif
 };
--- a/src/Magnum/AbstractFramebuffer.cpp
+++ b/src/Magnum/AbstractFramebuffer.cpp
@@ -105,8 +105,12 @@ void AbstractFramebuffer::bind() {
 }
 
 void AbstractFramebuffer::bindInternal(FramebufferTarget target) {
-    #if defined(MAGNUM_TARGET_GLES2) && !defined(MAGNUM_TARGET_WEBGL)
+    #if defined(MAGNUM_TARGET_GLES2)
+    #if !defined(MAGNUM_TARGET_WEBGL)
     (this->*Context::current().state().framebuffer->bindImplementation)(target);
+    #else
+    bindImplementationSingle();
+    #endif
     #else
     bindImplementationDefault(target);
     #endif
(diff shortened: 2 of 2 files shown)


smspillaz:
Side note: Thanks so much for making Magnum! I wrote my toy music visualiser using it and its running on WebGL now.

(http://smspillaz.github.io/fmv)

src/Magnum/AbstractFramebuffer.h
```diff
@@ -117,19 +117,19 @@ enum class FramebufferTarget: GLenum {
     /** Frambebuffer reading target */
     #ifndef MAGNUM_TARGET_GLES2
     Read = GL_READ_FRAMEBUFFER,
-    #elif !defined(MAGNUM_TARGET_WEBGL)
+    #elif defined(CORRADE_TARGET_APPLE)
     Read = GL_READ_FRAMEBUFFER_APPLE,
```
smspillaz: Note: I'm not *entirely* certain if this is correct. Maybe the best solution here is to look at the supported extension and try and figure out the correct framebuffer binding from there. Only supporting multi-sampling on iOS seems to me like a pretty big assumption.

mosra:
Hello, thanks for the report, detailed description, commit comments and encouragement :)

It took me a while to untangle what is actually happening inside the framebuffer binding code and, ugh, what a mess did I do there. Sorry. The last straw was probably in cecb4863f7c786153513b88dfdb3d32f516e0c4f, where I tried to fix calling unitialized function pointer, but fixed it *wrong* and it's even worse now.

> Framebuffer binding on WebGL was raising `GL_INVALID_ENUM` because `0` was being passed to `glBindFramebuffer`, `glCheckFramebufferStatus` and friends.

This certainly shouldn't happen. I see that your second commit fixes that, thank you. The values from `FramebufferTarget` should be *completely* ignored when there is no read/draw framebuffer distinction so it shouldn't matter whether there is `0`, `GL_FRAMEBUFFER` or `0xDEADBABE`. That's why I'm not using the enum values at all inside the "Single" implementations but instead passing `GL_FRAMEBUFFER` directly. I also deliberately put a `0`/`1` there instead of legal value as a some sort of assertion that this value will not be used anywhere -- and that, in fact, helped catch this bug.

 > I'm not *entirely* certain if this is correct. Maybe the best solution here is to look at the supported extensions and try and figure out the correct framebuffer binding from there.

You are right, it's not. Let me explain -- on OpenGL ES 2.0 there are *four* different extensions that are providing separate read/draw binding points and it's done exactly as you hinted -- Magnum is looking for supported extensions [in initialization step](https://github.com/mosra/magnum/blob/23646e6690068ee268451de3bd7cb103c1298e75/src/Magnum/Implementation/FramebufferState.cpp#L132-L163) and then selects single/separate bind implementation based on whether they are supported or not, so deciding at compile time is not enough. `GL_FRAMEBUFFER_READ_APPLE` is used just because it was the first value I came across, but I could as well use `GL_FRAMEBUFFER_READ_NV` or `GL_FRAMEBUFFER_READ_ANGLE`, they all (fortunately) have the same value, so it doesn't matter.

So, to make it short, I'll cherry-pick your second commit, but will ignore the other one for the above reasons. I just need to check with WebGL 2, but I guess it behaves the same as GLES3 so it should "just work" there.

- - -

Re side note: I was not able to run the project in any of my browsers (Opera/Chrome/Firefox on Linux), Upon loading it in Firefox 44 starts eating whole processor core and dragging wav/mp3 there doesn't do anything. This is printed in the console:

> uncaught exception: SimulateInfiniteLoop <unknown>
> TypeError: context.createJavaScriptNode is not a function

I'll try the native version, maybe I'll have a better luck :) Also, I planned to put a section called "Notable projects" in the README and would fit very well in there, would you be okay with that?

LB--:
> Re side note: I was not able to run the project in any of my browsers (Opera/Chrome/Firefox on Linux), Upon loading it in Firefox 44 starts eating whole processor core and dragging wav/mp3 there doesn't do anything.

It works for me in Chrome 48, but I can't get it to play anything. (Where should we continue this discussion?)

mosra:
Cherry-picked in d60fa66b27567621ba9b7929331cfa6411d13a41. Let's continue the discussion in smspillaz/fmv#1 :)

smspillaz:
Thanks for the detailed analysis!

> You are right, it's not. Let me explain -- on OpenGL ES 2.0 there are four different extensions that are providing separate read/draw binding points and it's done exactly as you hinted -- Magnum is looking for supported extensions in initialization step and then selects single/separate bind implementation based on whether they are supported or not, so deciding at compile time is not enough. GL_FRAMEBUFFER_READ_APPLE is used just because it was the first value I came across, but I could as well use GL_FRAMEBUFFER_READ_NV or GL_FRAMEBUFFER_READ_ANGLE, they all (fortunately) have the same value, so it doesn't matter.

Yup, this makes sense. I remember I faced similar confusion when I was implementing the same thing in the compiz blur plugin a long time ago.

Merging the second commit should be sufficient. I suppose there isn't really a way to abstract out the difference between read/draw bindings on implementations that do and don't support multi-sampling. Which is a shame for library users, but that's GL's problem I suppose.

mosra:
Yeah, as it currently stands, the only place where the user needs the `FramebufferTarget` enum is in the `check()` function to check if the framebuffer is read for reading/drawing. All other places are doing this distinction internally, fortunately.

OpenGL is a mess, right? :)

PR #137, opened 2016-03-01
WGL context library build fixed

dlardi:


mosra:
Ow, sorry about that one.

Thank you for the fix!

PR #141, opened 2016-04-13
Add GlfwApplication

Squareys:
Hi @mosra !

As discussed via gitter, I am implement an `Application` class based on GLFW, which supports creation of Vulkan contexts and therefore acts as a basis for the Magnum Vulkan implementation.

This already compiles, but is not enabled on ci. Since creating Vulkan contexts will only be supported in GLFW 3.2, which is not quite released yet, we will need to have the ci build it from https://github.com/glfw/glfw.

I have not tested this yet, it might be completely unusable. Other than that, you may want to have a look at naming/doc in the Key enum. Also, the main loop maybe.

Regards,
Squareys (Jonathan Hale)

TODO:
 - [x] `exit()`
 - ~~`tickEvent()`~~
 - [x] More context/window flags?
 - [x] Run a sample application
 - [x] Add a bootstrap for https://github.com/mosra/magnum-bootstrap.
 - ~~@mosra shall I already add the context creation for Vulkan in this PR?~~ No.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -128,6 +128,7 @@ endif()
 
 # Platform-independent (almost) application libraries
 if(NOT CORRADE_TARGET_NACL AND NOT CORRADE_TARGET_ANDROID)
+    option(WITH_GLFWAPPLICATION "Build GlfwApplication library")
     cmake_dependent_option(WITH_GLUTAPPLICATION "Build GlutApplication library" OFF "NOT TARGET_GLES" OFF)
     option(WITH_SDL2APPLICATION "Build Sdl2Application library" OFF)
 endif()
--- a/doc/building.dox
+++ b/doc/building.dox
@@ -206,6 +206,7 @@ need at least one). Choose the one which suits your requirements and your
 platform best:
 
 -   `WITH_ANDROIDAPPLICATION` - @ref Platform::AndroidApplication "AndroidApplication"
+-   `WITH_GLFWAPPLICATION` - @ref Platform::GlfwApplication "GlfwApplication"
 -   `WITH_GLUTAPPLICATION` - @ref Platform::GlutApplication "GlutApplication"
 -   `WITH_GLXAPPLICATION` - @ref Platform::GlxApplication "GlxApplication"
 -   `WITH_NACLAPPLICATION` - @ref Platform::NaClApplication "NaClApplication"
--- a/modules/FindGLFW.cmake
+++ b/modules/FindGLFW.cmake
@@ -1,0 +1,67 @@
+#.rst:
+# Find GLFW
+# ---------
+#
+# Finds the GLFW library. This module defines:
+#
+#  GLFW_FOUND               - True if GLFW library is found
+#  GLFW::GLFW               - GLFW imported target
+#
+# Additionally these variables are defined for internal usage:
+#
+#  GLFW_LIBRARY             - GLFW library
+#  GLFW_INCLUDE_DIR         - Root include dir
+#
+
+#
+#   This file is part of Magnum.
+#
+#   Copyright © 2010, 2011, 2012, 2013, 2014, 2015, 2016
+#             Vladimír Vondruš <mosra@centrum.cz>
+#   Copyright © 2016 Jonathan Hale <squareys@googlemail.com>
+#
+#   Permission is hereby granted, free of charge, to any person obtaining a
+#   copy of this software and associated documentation files (the "Software"),
+#   to deal in the Software without restriction, including without limitation
+#   the rights to use, copy, modify, merge, publish, distribute, sublicense,
+#   and/or sell copies of the Software, and to permit persons to whom the
+#   Software is furnished to do so, subject to the following conditions:
+#
+#   The above copyright notice and this permission notice shall be included
+#   in all copies or substantial portions of the Software.
+#
+#   THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+#   IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+#   FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+#   THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+#   LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+#   FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+#   DEALINGS IN THE SOFTWARE.
+#
+
+find_library(GLFW_LIBRARY NAMES glfw glfw3)
+
+# Include dir
+find_path(GLFW_INCLUDE_DIR
+    NAMES glfw3.h
+    PATH_SUFFIXES GLFW)
+
+include(FindPackageHandleStandardArgs)
+find_package_handle_standard_args("GLFW" DEFAULT_MSG
+    GLFW_LIBRARY
+    GLFW_INCLUDE_DIR)
+
+if(NOT TARGET GLFW::GLFW)
+    add_library(GLFW::GLFW UNKNOWN IMPORTED)
+
+    # Work around BUGGY framework support on OSX
+    # https://cmake.org/Bug/view.php?id=14105
+    if(CORRADE_TARGET_APPLE AND ${GLFW_LIBRARY} MATCHES "\\.framework$")
+        set_property(TARGET GLFW::GLFW PROPERTY IMPORTED_LOCATION ${GLFW_LIBRARY}/GLFW)
+    else()
+        set_property(TARGET GLFW::GLFW PROPERTY IMPORTED_LOCATION ${GLFW_LIBRARY})
+    endif()
+
+    set_property(TARGET GLFW::GLFW PROPERTY
+        INTERFACE_INCLUDE_DIRECTORIES ${GLFW_INCLUDE_DIR})
+endif()
--- a/src/Magnum/Platform/GlfwApplication.cpp
+++ b/src/Magnum/Platform/GlfwApplication.cpp
@@ -150,8 +150,8 @@ int GlfwApplication::exec() {
     return 0;
 }
 
-void GlfwApplication::staticKeyEvent(GLFWwindow*, int key, int, int action, int) {
-    KeyEvent e(static_cast<KeyEvent::Key>(key));
+void GlfwApplication::staticKeyEvent(GLFWwindow*, int key, int, int action, int mods) {
+    KeyEvent e(static_cast<KeyEvent::Key>(key), {static_cast<InputEvent::Modifier>(mods)});
 
     if(action == GLFW_PRESS) {
         _instance->keyPressEvent(e);
@@ -160,13 +160,13 @@ void GlfwApplication::staticKeyEvent(GLFWwindow*, int key, int, int action, int)
     } /* we don't handle GLFW_REPEAT */
 }
 
-void GlfwApplication::staticMouseMoveEvent(GLFWwindow*, double x, double y) {
-    MouseMoveEvent e{Vector2i{Int(x), Int(y)}};
+void GlfwApplication::staticMouseMoveEvent(GLFWwindow* window, double x, double y) {
+    MouseMoveEvent e{Vector2i{Int(x), Int(y)}, KeyEvent::getCurrentGlfwModifiers(window)};
     _instance->mouseMoveEvent(e);
 }
 
-void GlfwApplication::staticMouseEvent(GLFWwindow*, int button, int action, int) {
-    MouseEvent e(static_cast<MouseEvent::Button>(button));
+void GlfwApplication::staticMouseEvent(GLFWwindow*, int button, int action, int mods) {
+    MouseEvent e(static_cast<MouseEvent::Button>(button), {static_cast<InputEvent::Modifier>(mods)});
 
     if(action == GLFW_PRESS) {
         _instance->mousePressEvent(e);
@@ -175,19 +175,36 @@ void GlfwApplication::staticMouseEvent(GLFWwindow*, int button, int action, int)
     } /* we don't handle GLFW_REPEAT */
 }
 
-void GlfwApplication::staticMouseScrollEvent(GLFWwindow*, double xoffset, double yoffset) {
-    MouseScrollEvent e(Vector2d{xoffset, yoffset});
+void GlfwApplication::staticMouseScrollEvent(GLFWwindow* window, double xoffset, double yoffset) {
+    MouseScrollEvent e(Vector2d{xoffset, yoffset}, KeyEvent::getCurrentGlfwModifiers(window));
     _instance->mouseScrollEvent(e);
 
     if(yoffset != 0.0) {
-        MouseEvent e1((yoffset > 0.0) ? MouseEvent::Button::WheelUp : MouseEvent::Button::WheelDown);
-        _instance->mousePressEvent(e);
+        MouseEvent e1((yoffset > 0.0) ? MouseEvent::Button::WheelUp : MouseEvent::Button::WheelDown, KeyEvent::getCurrentGlfwModifiers(window));
         _instance->mousePressEvent(e1);
     }
 }
 
 void GlfwApplication::staticErrorCallback(int, const char* description) {
     Error() << description;
+}
+
+GlfwApplication::InputEvent::Modifiers GlfwApplication::KeyEvent::getCurrentGlfwModifiers(GLFWwindow* window) {
+    static_assert(GLFW_PRESS == true && GLFW_RELEASE == false, "GLFW press and release constants do not correspond to bool values.");
+
+    Modifiers mods = (glfwGetKey(window, Int(Key::LeftShift)) || glfwGetKey(window, Int(Key::RightShift)))
+                     ? Modifiers{Modifier::Shift} : Modifiers{};
+    if(glfwGetKey(window, Int(Key::LeftAlt)) || glfwGetKey(window, Int(Key::RightAlt))) {
+        mods |= Modifier::Alt;
+    }
+    if(glfwGetKey(window, Int(Key::LeftCtrl)) || glfwGetKey(window, Int(Key::RightCtrl))) {
+        mods |= Modifier::Ctrl;
+    }
+    if(glfwGetKey(window, Int(Key::RightSuper))) {
+        mods |= Modifier::AltGr;
+    }
+
+    return mods;
 }
 
 void GlfwApplication::viewportEvent(const Vector2i&) {}
--- a/src/Magnum/Platform/GlfwApplication.cpp
+++ b/src/Magnum/Platform/GlfwApplication.cpp
@@ -178,6 +178,12 @@ void GlfwApplication::staticMouseEvent(GLFWwindow*, int button, int action, int)
 void GlfwApplication::staticMouseScrollEvent(GLFWwindow*, double xoffset, double yoffset) {
     MouseScrollEvent e(Vector2d{xoffset, yoffset});
     _instance->mouseScrollEvent(e);
+
+    if(yoffset != 0.0) {
+        MouseEvent e1((yoffset > 0.0) ? MouseEvent::Button::WheelUp : MouseEvent::Button::WheelDown);
+        _instance->mousePressEvent(e);
+        _instance->mousePressEvent(e1);
+    }
 }
 
 void GlfwApplication::staticErrorCallback(int, const char* description) {
--- a/src/Magnum/Platform/GlfwApplication.cpp
+++ b/src/Magnum/Platform/GlfwApplication.cpp
@@ -1,0 +1,208 @@
+/*
+    This file is part of Magnum.
+
+    Copyright © 2010, 2011, 2012, 2013, 2014, 2015, 2016
+              Vladimír Vondruš <mosra@centrum.cz>
+
+    Permission is hereby granted, free of charge, to any person obtaining a
+    copy of this software and associated documentation files (the "Software"),
+    to deal in the Software without restriction, including without limitation
+    the rights to use, copy, modify, merge, publish, distribute, sublicense,
+    and/or sell copies of the Software, and to permit persons to whom the
+    Software is furnished to do so, subject to the following conditions:
+
+    The above copyright notice and this permission notice shall be included
+    in all copies or substantial portions of the Software.
+
+    THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+    IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+    FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+    THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+    LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+    FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+    DEALINGS IN THE SOFTWARE.
+*/
+
+#include "GlfwApplication.h"
+
+#include <tuple>
+
+#include "Magnum/Version.h"
+#include "Magnum/Platform/Context.h"
+#include "Magnum/Platform/ScreenedApplication.hpp"
+
+namespace Magnum { namespace Platform {
+
+GlfwApplication* GlfwApplication::_instance = nullptr;
+
+#ifndef DOXYGEN_GENERATING_OUTPUT
+GlfwApplication::GlfwApplication(const Arguments& arguments): GlfwApplication{arguments, Configuration{}} {}
+#endif
+
+GlfwApplication::GlfwApplication(const Arguments& arguments, const Configuration& configuration): GlfwApplication{arguments, nullptr} {
+    createContext(configuration);
+}
+
+GlfwApplication::GlfwApplication(const Arguments& arguments, std::nullptr_t)
+    : _context{new Context{NoCreate, arguments.argc, arguments.argv}},
+      _needsRedraw(true)
+{
+    /* Save global instance */
+    _instance = this;
+
+    /* Init GLFW */
+    glfwSetErrorCallback(staticErrorCallback);
+
+    if(!glfwInit()) {
+        Error() << "Could not initialize GLFW";
+        std::exit(8);
+    }
+}
+
+void GlfwApplication::createContext() { createContext({}); }
+
+void GlfwApplication::createContext(const Configuration& configuration) {
+    if(!tryCreateContext(configuration)) std::exit(1);
+}
+
+bool GlfwApplication::tryCreateContext(const Configuration& configuration) {
+    CORRADE_ASSERT(_context->version() == Version::None, "Platform::GlfwApplication::tryCreateContext(): context already created", false);
+
+    static_assert(GLFW_TRUE == true && GLFW_FALSE == false, "GLFW does not have sane bool values.");
+
+    /* Window flags */
+    GLFWmonitor* monitor = nullptr; /* Needed for setting fullscreen */
+    if (configuration.windowFlags() >= Configuration::WindowFlag::Fullscreen) {
+        monitor = glfwGetPrimaryMonitor();
+        glfwWindowHint(GLFW_AUTO_ICONIFY, configuration.windowFlags() >= Configuration::WindowFlag::AutoIconify);
+    } else {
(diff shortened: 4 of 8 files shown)


modules/FindGLFW.cmake
```diff
@@ -0,0 +1,77 @@
...
+#   IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+#   FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+#   THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+#   LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+#   FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+#   DEALINGS IN THE SOFTWARE.
+#
+
+find_library(GLFW_LIBRARY NAMES glfw glfw3)
+set(GLFW_LIBRARY_NEEDED GLFW_LIBRARY)
```
mosra: This was just a workaround needed for SDL in Emscripten, where it was linked by default and thus no library was available. Let's assume the GLFW library is always there.

modules/FindGLFW.cmake
```diff
@@ -0,0 +1,77 @@
...
+
+# Include dir
+find_path(GLFW_INCLUDE_DIR
+    NAMES glfw3.h
+    PATH_SUFFIXES ${_GLFW_PATH_SUFFIXES})
+
+include(FindPackageHandleStandardArgs)
+find_package_handle_standard_args("GLFW" DEFAULT_MSG
+    ${GLFW_LIBRARY_NEEDED}
+    ${_GLFW_FRAMEWORK_LIBRARY_NAMES}
```
mosra: Also `${_GLFW_FRAMEWORK_LIBRARY_NAMES}` was there for SDL only because it depended on quite a lot iOS frameworks. No such case with GLFW I think, so let's remove that oo.

src/Magnum/Platform/GlfwApplication.h
```diff
@@ -0,0 +1,652 @@
...
+         */
+        int exec() {
+            while(!glfwWindowShouldClose(_window)) {
+                if(_needsRedraw) {
+                    drawEvent();
+                }
+                glfwPollEvents();
+            }
+            return 0;
+        }
```
mosra: Can you put the implementation into the `*.cpp` file?

src/Magnum/Platform/GlfwApplication.h
```diff
@@ -0,0 +1,652 @@
...
+*/
+class GlfwApplication::Configuration {
+    public:
+        /**
+         * @brief Context flag
+         *
+         * @see @ref Flags, @ref setFlags()
+         */
+        enum class Flag: int {
+            NoError = GLFW_CONTEXT_NO_ERROR,  /**< Create debug context */
```
mosra: Huh, GLFW supports `KHR_no_error` already? Wow. I don't even have the extension in the list yet :)

(The comment is misleading, though.)

mosra:
That was fast! :)

Let's start small, so:
- I would like to have a completely separate PR for Vulkan that will contain all the additions (it will be a long-running one, I guess)
- no `tickEvent()`, no context/window flags that are non-trivial to add

Here's a list of places you need to touch to mention the new application everywhere:
- put it in the list in `doc/cmake.dox`
- put it in the list in `doc/building.dox`
- `modules/FindMagnum.cmake` (once in docs, once in the regexp and once in the Application section)
- add an option to `CMakeLists.txt`

modules/FindGLFW.cmake
Squareys: Ah, I see, makes sense.

src/Magnum/Platform/GlfwApplication.h
Squareys: Will do! :)
- [ ]

src/Magnum/Platform/GlfwApplication.h
Squareys: Ah, right, sorry. I based the code off of GlutApplication, so this is a leftover comment.

modules/FindGLFW.cmake
Squareys: Uhh, yeah, remnant of the SDL find module.

- [ ]

Squareys:
Thanks for the quick reply!

> add an option to CMakeLists.txt

I forgot to add that file to the commit, consider it done :)

> put it in the list in doc/cmake.dox
Put it in the list in doc/building.dox

Right, almost forgot about those. Current goal is to have this done by Sunday at the latest.

src/Magnum/Platform/GlfwApplication.h
```diff
@@ -0,0 +1,649 @@
...
+        /** @brief sRGB-capable default framebuffer */
+        bool isSRGBCapable() const {
+            return _srgbCapable;
+        }
+
+        /**
+         * @brief Set sRGB-capable default framebuffer
+         * @return Reference to self (for method chaining)
+         */
+        Configuration& setSRGBCapable(bool enabled) {
```
Squareys: @mosra I wonder if this shouldn't be moved to the `Flag` enum instead? Analog one could make the change in Sdl2Application to make a switch easier. Or leave both this way? There are more such flags which could be either expressed one way or the orther.

mosra: Hmm. I also wondered about this. Currently the `Flag` enum maps directly to whatever platform-specific bit flags there are (so SDL has its own, GLX, WGL, CGL, EGL also have its own) and unfortunately "sRGB capable" is not in those bit flags on neither of those and one has to use special value/function for it. So it made more sense to me to put it separately instead of having a big switch for the `Flag` values to decide whether they belong into context flags or somewhere else.

Let's do it consistently with the rest first, we can change later.

Squareys:
@mosra Done!

I revisited the context and window flags aswell as key and button enums. I was missing some other stuff aswell as an implementation for scrolling, which is done via a separate event **(!)**, since GLFW supports scrolling not just via mouse, but also touchpad etc, which may also have horizontal scrolling.

Also, the boostrap project is done, but I need an existing `base-glfw` branch to create a pullrequest for it.

I therefore declare this pullrequest as ready for review :)

Squareys:
We should wait for https://github.com/glfw/glfw/issues/738 before merging this.

src/Magnum/Platform/GlfwApplication.h
```diff
@@ -0,0 +1,753 @@
...
+         * @see @ref WindowFlags, @ref setWindowFlags()
+         */
+        enum class WindowFlag: UnsignedShort {
+            Fullscreen = 1 << 0,   /**< Fullscreen window */
+            Resizeable = 1 << 1,   /**< Resizeable window */
+            Hidden = 1 << 2,       /**< Hidden window */
+            Maximized = 1 << 3,    /**< Maximized window */
+            Minimized = 1 << 4,    /**< Minimized window */
+            Floating = 1 << 5,     /**< Window floating above others, top-most */
+            AutoIconify = 1 << 6,  /**< Automatically iconify (minimize) if fullscreen window looses input focus */
```
mosra: loses, not looses ;)

src/Magnum/Platform/GlfwApplication.h
```diff
@@ -0,0 +1,753 @@
...
+            Numpad7 = GLFW_KEY_KP_7,            /**< Numpad 7 */
+            Numpad8 = GLFW_KEY_KP_8,            /**< Numpad 8 */
+            Numpad9 = GLFW_KEY_KP_9,            /**< Numpad 9 */
+            NumpadDecimal = GLFW_KEY_KP_DECIMAL,    /**< Numpad Decimal */
+            NumpadDivide = GLFW_KEY_KP_DIVIDE,      /**< Numpad Divide */
+            NumpadMultiply = GLFW_KEY_KP_MULTIPLY,  /**< Numpad Multiply */
+            NumpadSubtract = GLFW_KEY_KP_SUBTRACT,  /**< Numpad Subtract */
+            NumpadAdd = GLFW_KEY_KP_ADD,        /**< Numpad Add */
+            NumpadEnter = GLFW_KEY_KP_ENTER,    /**< Numpad Enter */
+            NumPadEqual = GLFW_KEY_KP_EQUAL,    /**< Numpad Equal */
```
mosra: Hmm... both SDL and GLFW name the numpad as `KP`, which might mean ... keypad? Never heard such name :D 

What about shorter prefix, just `Num`? For consistency with "normal" numbers, I would also name the numbers like `NumZero` etc.

src/Magnum/Platform/GlfwApplication.h
```diff
@@ -0,0 +1,753 @@
...
+            LeftShift = GLFW_KEY_LEFT_SHIFT,    /**< Left shift */
+            LeftCtrl = GLFW_KEY_LEFT_CONTROL,   /**< Left control */
+            LeftAlt = GLFW_KEY_LEFT_ALT,        /**< Left alt */
+            LeftSuper = GLFW_KEY_LEFT_SUPER,    /**< Left super */
+            RightShift = GLFW_KEY_RIGHT_SHIFT,  /**< Right shift */
+            RightCtrl = GLFW_KEY_RIGHT_CONTROL, /**< Right control */
+            RightAlt = GLFW_KEY_RIGHT_ALT,      /**< Right alt */
+            RightSuper = GLFW_KEY_RIGHT_SUPER,  /**< Right super */
```
mosra: Regarding modififers, I'm not sure if SDL2 emits a key event on shift/alt/ctrl keys or not. I think they are presented only as a modifier -- so if you press a key combination, SDL gives you a opportunity to ask whether any of the modifiers is pressed next to the "real" key.

src/Magnum/Platform/GlfwApplication.h
```diff
@@ -0,0 +1,753 @@
...
+    public:
+
+        /** @brief Scroll offset */
+        constexpr Vector2d offset() const { return _offset; }
+
+    private:
+        constexpr MouseScrollEvent(const Vector2d& offset): _offset(offset) {}
+
+        const Vector2d _offset;
+};
```
mosra: This means that `WheelUp`/`WheelDown` mouse buttons as treated in SDL2 would need to be handled with a separate event here?

I would like to have things consistent, is it possible to get the wheel also as a button? If not, then I should add similar `MouseScrollEvent` to Sdl2Application, if that's possible with SDL2.

src/Magnum/Platform/GlfwApplication.cpp
```diff
@@ -0,0 +1,213 @@
...
+    if ((configuration.windowFlags() & WindowFlag::Fullscreen) != NONE) {
+        monitor = glfwGetPrimaryMonitor();
+        glfwWindowHint(GLFW_AUTO_ICONIFY, ((configuration.windowFlags() & WindowFlag::AutoIconify) != NONE) ? GLFW_TRUE : GLFW_FALSE);
+    } else {
+        const Configuration::WindowFlags& flags = configuration.windowFlags();
+        glfwWindowHint(GLFW_RESIZABLE, ((flags & WindowFlag::Resizeable) != NONE) ? GLFW_TRUE : GLFW_FALSE);
+        glfwWindowHint(GLFW_VISIBLE, ((flags & WindowFlag::Hidden) != NONE) ? GLFW_FALSE : GLFW_TRUE);
+        glfwWindowHint(GLFW_MAXIMIZED, ((flags & WindowFlag::Maximized) != NONE) ? GLFW_TRUE : GLFW_FALSE);
+        glfwWindowHint(GLFW_ICONIFIED, ((flags & WindowFlag::Minimized) != NONE) ? GLFW_TRUE : GLFW_FALSE);
+        glfwWindowHint(GLFW_FLOATING, ((flags & WindowFlag::Floating) != NONE) ? GLFW_TRUE : GLFW_FALSE);
```
mosra: `Containers::EnumSet` has an `operator>=` that is equivalent to `(flags & A) == A`. Might reduce the verbosity here, I think.

modules/FindGLFW.cmake
```diff
@@ -0,0 +1,68 @@
...
+include(FindPackageHandleStandardArgs)
+find_package_handle_standard_args("GLFW" DEFAULT_MSG
+    ${GLFW_LIBRARY_NEEDED}
+    GLFW_INCLUDE_DIR)
+
+if(NOT TARGET GLFW::GLFW)
+    add_library(GLFW::GLFW UNKNOWN IMPORTED)
+
+    # Work around BUGGY framework support on OSX
+    # https://cmake.org/Bug/view.php?id=13765
```
mosra:
```cmake
# https://cmake.org/Bug/view.php?id=14105
```

Apologies -- I did a wrong copypaste here. Fixed in current `FindSDL2.cmake`, you probably based this module on an older version.

doc/building.dox
```diff
@@ -644,8 +645,8 @@ replace `localhost:8080` with your server name, replace `update-job` with
     java -jar ~/jenkins-cli.jar -s http://localhost:8080 update-job Magnum-GLTests < package/ci/jenkins-gltests.xml
 
 Build is done using **Ninja** build system and everything possible is enabled,
-thus you need also **SDL2**, **GLUT** and **OpenAL** libraries. It expects
-that **GCC** >= 4.9, 4.8, 4.7 and **Clang** are installed and there
+thus you need also **SDL2**, **GLFW**, **GLUT** and **OpenAL** libraries. It
+expects that **GCC** >= 4.9, 4.8, 4.7 and **Clang** are installed and there
```
mosra: This would imply that you have to update *all* packaging and CI scripts (which is, like, a thousand locations). Please revert this bit.

I'm not sure whether we want to require GLFW everywhere. I will think it through and do the required updates after merging.

CMakeLists.txt
```diff
@@ -129,6 +129,7 @@ endif()
 # Platform-independent (almost) application libraries
 if(NOT CORRADE_TARGET_NACL AND NOT CORRADE_TARGET_ANDROID)
     cmake_dependent_option(WITH_GLUTAPPLICATION "Build GlutApplication library" OFF "NOT TARGET_GLES" OFF)
+    cmake_dependent_option(WITH_GLFWAPPLICATION "Build GlfwApplication library" OFF "NOT TARGET_GLES" OFF)
```
mosra: I think GLFW supports GLES, so it should be looking just like this:
```cmake
option(WITH_GLFWAPPLICATION "Build GlfwApplication library" OFF)
```

modules/FindGLFW.cmake
```diff
@@ -0,0 +1,68 @@
...
+
+# Include dir
+set(_GLFW_PATH_SUFFIXES GLFW)
+find_path(GLFW_INCLUDE_DIR
+    NAMES glfw3.h
+    PATH_SUFFIXES ${_GLFW_PATH_SUFFIXES})
+
+include(FindPackageHandleStandardArgs)
+find_package_handle_standard_args("GLFW" DEFAULT_MSG
+    ${GLFW_LIBRARY_NEEDED}
```
mosra: One last remaining bit: should be `GLFW_LIBRARY`, not `${GLFW_LIBRARY_NEEDED}`.

modules/FindGLFW.cmake
```diff
@@ -0,0 +1,68 @@
...
+#   DEALINGS IN THE SOFTWARE.
+#
+
+find_library(GLFW_LIBRARY NAMES glfw glfw3)
+
+# Include dir
+set(_GLFW_PATH_SUFFIXES GLFW)
+find_path(GLFW_INCLUDE_DIR
+    NAMES glfw3.h
+    PATH_SUFFIXES ${_GLFW_PATH_SUFFIXES})
```
mosra: Also not needed to have the `${_GLFW_PATH_SUFFIXES}` variable here, just paste the value directly.

src/Magnum/Platform/GlfwApplication.cpp
```diff
@@ -0,0 +1,213 @@
...
+#include "Magnum/Version.h"
+#include "Magnum/Platform/Context.h"
+#include "Magnum/Platform/ScreenedApplication.hpp"
+
+namespace Magnum { namespace Platform {
+
+typedef GlfwApplication::Configuration::Flag Flag;
+typedef GlfwApplication::Configuration::Flags Flags;
+typedef GlfwApplication::Configuration::WindowFlag WindowFlag;
+typedef GlfwApplication::Configuration::WindowFlags WindowFlags;
```
mosra: Hmm... I see that the flags are used only in the `tryCreateContext()`, right? Then I would move these typedefs to the function block so they don't leak outside.

Also, maybe after the simplification the additional `Configuration::` prefix wouldn't do *that* much harm and so these typedefs won't be needed at all.

src/Magnum/Platform/GlfwApplication.h
Squareys: Oops, right :D

src/Magnum/Platform/GlfwApplication.h
Squareys: `Num` sounds fine to me! I just noticed that I even wrote `NumpadEnter`, but `NumPadEqual`. Will do.

src/Magnum/Platform/GlfwApplication.h
Squareys: I believe GLFW lets you do both, it passes modifiers to the key callback as well as calling the key callback when `Alt`, `Ctrl` or `Shift` are pressed...

src/Magnum/Platform/GlfwApplication.h
Squareys: The thing is that the scroll `offset`is a double, but for my simple mouse just returns values in `{(0.0, 1.0), (0.0, -1.0)}). This could be easily mapped to the SDL way of handling scrolling, but could break for more sophisticated scrolling devices.

SDL [does have something similar: SDL_MouseWheelEvent](https://wiki.libsdl.org/SDL_MouseWheelEvent)

src/Magnum/Platform/GlfwApplication.cpp
Squareys: *Uh*, thank god :D I was very unhappy with that code.

```
glfwWindowHint(GLFW_RESIZABLE, flags >= WindowFlag::Resizeable);
```

So beautiful! :dizzy_face:

doc/building.dox
Squareys: Alright, will do, thanks!

modules/FindGLFW.cmake
Squareys: Damn, sorry.

src/Magnum/Platform/GlfwApplication.h
mosra: Hmm, then a separate wheel event makes more sense. I think NaCl (Android?) has something similar. Now the question is: wheel event or scroll event? :D

LB--: I'd vote for `WheelEvent` if it directly relates to the signal from the mouse, or `ScrollEvent` if it takes into account the user's scroll speed setting in the operating system. But, touchpads complicate things. Why is naming things so hard?

Squareys: @LB-- What if it's a touchpad? Then it's obviously not a wheel, but you can still scroll.

nabijaczleweli: cornercase

mosra: Hmm... yes, thinking about this, `ScrollEvent` actually makes more sense because not all "mousey" devices have wheels (touchpads, Apple mouses) and you can still scroll with them.

On the other hand, I would keep the moused buttons named `WheelUp` and `WheelDown`, because that's what they really represent.

Squareys: Alright, so "ScrollEvent" and keep the buttons? (aka. add them for GlfwApplication, there is no key for it, so this would have to be done manually by checking offset.y() > 0/ < 0)

Shall I add the Scroll Event in SDL Application?

mosra: Yes, `ScrollEvent` and implement the buttons, if possible and easy to do.

No need to put it into `Sdl2Application` yet, we can do it later. Focus on the Vulkan stuff ;)

src/Magnum/Platform/GlfwApplication.h
Squareys: I guess, left todo is modifiers in key event?

src/Magnum/Platform/GlfwApplication.h
Squareys: Done https://github.com/mosra/magnum/pull/141/commits/887779544fc07b8d8e32ddaf51220a33cb7d65fc

src/Magnum/Platform/GlfwApplication.h
mosra: Yes. Seems to be way easier than with SDL -- they're provided directly in the last parameter in the key event callback.

src/Magnum/Platform/GlfwApplication.cpp
```diff
@@ -178,13 +178,24 @@ void GlfwApplication::staticMouseEvent(GLFWwindow*, int button, int action, int)
...
+        if(glfwGetKey(window, Int(KeyEvent::Key::LeftAlt)) == GLFW_PRESS || glfwGetKey(window, Int(KeyEvent::Key::RightAlt)) == GLFW_PRESS) {
+            mods |= InputEvent::Modifier::Alt;
+        }
+        if(glfwGetKey(window, Int(KeyEvent::Key::LeftCtrl)) == GLFW_PRESS || glfwGetKey(window, Int(KeyEvent::Key::RightCtrl)) == GLFW_PRESS) {
+            mods |= InputEvent::Modifier::Ctrl;
+        }
+        if(glfwGetKey(window, Int(KeyEvent::Key::RightSuper)) == GLFW_PRESS) {
+            mods |= InputEvent::Modifier::AltGr;
+        }
+        MouseEvent e1((yoffset > 0.0) ? MouseEvent::Button::WheelUp : MouseEvent::Button::WheelDown, mods);
```
Squareys: To have modifiers with all buttons, this monstrosity is necessary. I propose to deprecate Sdl2Application::MouseEvent::Button::Wheel*, since they are simulated anyway, in the same way in which they are simulated here [see this code](https://github.com/mosra/magnum/blob/master/src/Magnum/Platform/Sdl2Application.cpp#L372)

mosra: Hm... and one would think that ctrl/shift + scroll is pretty common to be handled sanely in all toolkits :(

Squareys: Well, it is... but not in the Scroll Event.

src/Magnum/Platform/GlfwApplication.cpp
```diff
@@ -79,7 +79,10 @@ bool GlfwApplication::tryCreateContext(const Configuration& configuration) {
         const Configuration::WindowFlags& flags = configuration.windowFlags();
         glfwWindowHint(GLFW_RESIZABLE, flags >= Configuration::WindowFlag::Resizeable);
         glfwWindowHint(GLFW_VISIBLE, !(flags >= Configuration::WindowFlag::Hidden));
-        glfwWindowHint(GLFW_MAXIMIZED, flags >= Configuration::WindowFlag::Maximized);
+        if(flags >= Configuration::WindowFlag::Maximized) {
+            /* FIXME: Workaround for https://github.com/glfw/glfw/issues/738 */
+            glfwWindowHint(GLFW_MAXIMIZED, GLFW_TRUE);
+        }
```
Squareys: @mosra I will remove the workaround. The bug has been fixed and GLFW_MAXIMIZED was added for 3.2, so the bug will not exist in any release of GLFW.

src/Magnum/Platform/GlfwApplication.cpp
Squareys: @mosra So what do you say? Are the `WheelUp` and `WheelDown` "Buttons" really needed when you also have a ScrollEvent in both SDL2 and GLFW? As mentioned above, those "Buttons" are simulated in `Sdl2Application` aswell, there is no 1:1 correspondence for them in SDL2 [=> MouseButtonEvent documentation](https://wiki.libsdl.org/SDL_MouseButtonEvent).

mosra: I don't see a problem with emulating `WheelUp` and `WheelDown`, that is just a single line in the above example and the same goes for `Sdl2Application`. (At least until we deprecate that and replace it with `ScrollEvent` everywhere. But later. I don't want to change everything now.)

To make myself clear: my concern was not the "emulation", but the crazy handling of scroll event *modifiers* in GLFW. But as there is apparently no other way, we have to live with it.

src/Magnum/Platform/GlfwApplication.cpp
mosra: Oh, okay, so ... the feature was not present at all in 3.1?

src/Magnum/Platform/GlfwApplication.cpp
Squareys: At the moment ScrollEvent does not have modifiers, merely KeyEvents and MouseEvent have them. Shall I add modifiers to ScrollEvent and MouseMoveEvent too?

mosra: If that's just a copypaste of the above, then why not. If that's too time consuming, then don't :)

src/Magnum/Platform/GlfwApplication.cpp
Squareys: Correct.

mosra: Ah, great :)

src/Magnum/Platform/GlfwApplication.cpp
```diff
@@ -193,6 +192,24 @@ void GlfwApplication::staticErrorCallback(int, const char* description) {
...
+    if(glfwGetKey(window, Int(Key::LeftCtrl)) || glfwGetKey(window, Int(Key::RightCtrl))) {
+        mods |= Modifier::Ctrl;
+    }
+    if(glfwGetKey(window, Int(Key::RightSuper))) {
+        mods |= Modifier::AltGr;
+    }
+
+    return mods;
+}
+
```
Squareys: @mosra I tried to clean this up a bit to make it look as least bad as possible. I think this should be okay.

mosra:
:+1: 

Since the bug was fixed, I think that workaround commit should be dropped. Also, can you rebase that fixup commit? I'll merge right after.

Squareys:
@mosra Done, thanks! :)

src/Magnum/Platform/GlfwApplication.cpp
```diff
@@ -0,0 +1,214 @@
...
+    } /* we don't handle GLFW_REPEAT */
+}
+
+void GlfwApplication::staticMouseScrollEvent(GLFWwindow*, double xoffset, double yoffset) {
+    MouseScrollEvent e(Vector2d{xoffset, yoffset});
+    _instance->mouseScrollEvent(e);
+
+    if(yoffset != 0.0) {
+        MouseEvent e1((yoffset > 0.0) ? MouseEvent::Button::WheelUp : MouseEvent::Button::WheelDown);
+        _instance->mousePressEvent(e);
```
mosra: This line seems somehow superfluous, there should be just one press event, right? Or am I missing something?

Squareys: Aaaargh, that was a bad rebase -_-
it should be `e1`

Squareys: Ah, that's one line further down. Still bad rebase, will fix in a sec

Squareys: Done.

Squareys: Wait. The modifiers are gone. Seems like I force pushed some outdated code.

mosra:
Merged! :sparkles:

Squareys:
Awesome, thanks! :tada:

src/Magnum/Platform/GlfwApplication.cpp
```diff
@@ -0,0 +1,208 @@
...
+        glfwWindowHint(GLFW_AUTO_ICONIFY, configuration.windowFlags() >= Configuration::WindowFlag::AutoIconify);
+    } else {
+        const Configuration::WindowFlags& flags = configuration.windowFlags();
+        glfwWindowHint(GLFW_RESIZABLE, flags >= Configuration::WindowFlag::Resizeable);
+        glfwWindowHint(GLFW_VISIBLE, !(flags >= Configuration::WindowFlag::Hidden));
+        glfwWindowHint(GLFW_MAXIMIZED, flags >= Configuration::WindowFlag::Maximized);
+        glfwWindowHint(GLFW_ICONIFIED, flags >= Configuration::WindowFlag::Minimized);
+        glfwWindowHint(GLFW_FLOATING, flags >= Configuration::WindowFlag::Floating);
+    }
+    glfwWindowHint(GLFW_FLOATING, configuration.windowFlags() >= Configuration::WindowFlag::Focused);
```
elmindreda: Should this be `GLFW_FOCUSED`?

Squareys: Yeah, good catch! Thanks :)

mosra: Just commited a fix in 563888e02ef82f752e1e906d8191ed49e7080eff. Thank you!

PR #157, opened 2016-06-13
Allow horizontal-only scrolling

staticfloat:
I have an app that I've been writing that makes use of two-finger scrolling on OSX using SDL.  Although SDL doesn't make it easy to do fancy things like inertial scrolling, it is still very useful to be able to scroll horizontally in user interfaces.

This API isn't perfect, as there is still a `WheelUp`/`WheelDown` flag being set.  Personally, I don't think that `WheelUp` and `WheelDown` should be part of the API, as horizontal scrolling is becoming a very common idiom in human-computer interaction, and as such the concept of the "mouse wheel" is being consumed by the more general concept of fully 2d scrolling.  I think you have a better idea of how to deal with those large API questions however, so I have only submitted a pull request to not dismiss events that have an `event.wheel.y` component of `0`.

### DIFF
--- a/src/Magnum/Platform/Sdl2Application.cpp
+++ b/src/Magnum/Platform/Sdl2Application.cpp
@@ -378,12 +378,11 @@ void Sdl2Application::mainLoop() {
                 event.type == SDL_MOUSEBUTTONDOWN ? mousePressEvent(e) : mouseReleaseEvent(e);
             } break;
 
-            case SDL_MOUSEWHEEL:
-                if(event.wheel.y != 0) {
-                    MouseEvent e(event.wheel.y > 0 ? MouseEvent::Button::WheelUp : MouseEvent::Button::WheelDown, {event.wheel.x, event.wheel.y}, 0);
-                    mousePressEvent(e);
-                } break;
-
+            case SDL_MOUSEWHEEL: {
+                MouseEvent e(event.wheel.y > 0 ? MouseEvent::Button::WheelUp : MouseEvent::Button::WheelDown, {event.wheel.x, event.wheel.y}, 0);
+                mousePressEvent(e);
+                break;
+            }
             case SDL_MOUSEMOTION: {
                 MouseMoveEvent e({event.motion.x, event.motion.y}, {event.motion.xrel, event.motion.yrel}, static_cast<MouseMoveEvent::Button>(event.motion.state));
                 mouseMoveEvent(e);


mosra:
Thank you! Merged in c817f8412e590b1f499261b2c7c37a5434ce0f07. 

Yes, as you say, the current API is weird (emulating wheel up/down "buttons" and position being relative scroll amount instead of, ahem, position). There should be a new scroll event (similar to what is now in `GlfwApplication`). It's on my TODO list.

src/Magnum/Platform/Sdl2Application.cpp
```diff
@@ -378,12 +378,11 @@ void Sdl2Application::mainLoop() {
...
 
-            case SDL_MOUSEWHEEL:
-                if(event.wheel.y != 0) {
-                    MouseEvent e(event.wheel.y > 0 ? MouseEvent::Button::WheelUp : MouseEvent::Button::WheelDown, {event.wheel.x, event.wheel.y}, 0);
-                    mousePressEvent(e);
-                } break;
-
+            case SDL_MOUSEWHEEL: {
+                MouseEvent e(event.wheel.y > 0 ? MouseEvent::Button::WheelUp : MouseEvent::Button::WheelDown, {event.wheel.x, event.wheel.y}, 0);
+                mousePressEvent(e);
```
Squareys: So... this will emit a `MouseEvent(WheelDown, {x, 0}, 0)` if `y == 0`, for both `x > 0` and `x < 0`? Is that intended? Maybe it would make sense to have `MouseEvent::Button::WheelRight` and  `MouseEvent::Button::WheelLeft`?

mosra: Aargh. Yes. Correct.

... probably the sanest thing to do is to introduce the scroll event and drop this. :D

staticfloat: Yep, that's what I was referring to as a non-ideal API. :)

I would be much happier to get rid of the MouseWheel distinction completely and move towards defining a more generic "scroll" event.

As a side question, Is there any way to get the smooth/inertial scrolling that most native OSX applications have inside of Magnum?  I imagine it might require using a different platform than SDL2, and of course would not be cross-platform, but it would still be very nice to have.

mosra: Hmm... I don't know, maybe some SDL2 touch event could be able to provide the acceleration vector? Or maybe GLFW knows it?

If you find something like that, please let me know :)

mosra: @staticfloat Proper scroll event is implemented in 2a77856df2e7d55ec587306f6184072865ba20b2.

staticfloat: Thanks @mosra, happy to report that it works perfectly.

I can furthermore report that GLFW does indeed do inertial scrolling on OSX, thanks for the tip!  I tested using [this test repository](https://github.com/staticfloat/scroll_test).

mosra: GLFW is great, isn't it :)

SDL2 could maybe handle it through `SDL_TouchFingerEvent`, but that's not exposed yet (and would add inconsistency across platforms).

PR #158, opened 2016-06-25
Ignore .DS_Store files for those on MacOS

staticfloat:
These files get littered around by Finder, make git output annoying when they're not ignored. :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ pkg
 *.kate-swp
 *.pkg.tar.xz
 CMakeLists.txt.user*
+.DS_Store


mosra:
Hmm, I think this should belong to a system-wide `.gitignore` file ([howto](http://stackoverflow.com/questions/7335420/global-git-ignore)) instead of specifying it in every project `.gitignore`.

That being said, I'm also guilty of putting `*.kate-swp` into the `.gitignore` file instead of having it ignored globally :)

staticfloat:
Fair enough.  That is definitely a better solution, if only us humans were as logical as the machines we work with.  ;)

PR #159, opened 2016-06-26
Add `Cmd` mapping for `KMOD_GUI`

staticfloat:
This maps to the Command key on MacOS machines, a very important key for those that wish to obey things like the typical close keystroke (CMD + W) or the typical save keystroke (CMD + S)

### DIFF
--- a/src/Magnum/Platform/Sdl2Application.h
+++ b/src/Magnum/Platform/Sdl2Application.h
@@ -975,6 +975,7 @@ class Sdl2Application::InputEvent {
             Ctrl = KMOD_CTRL,           /**< Ctrl */
             Alt = KMOD_ALT,             /**< Alt */
             AltGr = KMOD_MODE,          /**< AltGr */
+            Cmd = KMOD_GUI,             /**< Cmd */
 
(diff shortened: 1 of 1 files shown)


staticfloat:
Another option for the name is `Gui`.  It is quite annoying that (on MacOS at least) what GLFW calls `AltGr` is different than what SDL2 calls `AltGr`.

mosra:
I think I always saw the Windows/⌘ key named "Super". Could that naming work for you?

Seeing it being assigned to AltGr in GlfwApplication, which is wrong (at least on my Linux machine it maps to the Windows key). Will fix that.

On the other hand, AltGr doesn't seem to be reported at all, both SDL2 and GLFW are ignoring it completely, maybe due to my Czech keyboard layout.

mosra:
Fixed the GLFW mapping in 140f378739f9feeb946a8831c74278463820d519 and added the Super key to SDL2 application in e40533b841727a3c9a85737e7a5848b3f9c3671e.

staticfloat:
`Super` works great; functionality is now equivalent on both SDL2 and GLFW.  Thanks!

PR #160, opened 2016-06-26
Change GLFW keymappings to use actual GLFW defines

staticfloat:
Note that the `Plus` and `Percent` keys are not available without using the SHIFT key on the typical US English keyboard layout, and as such do not have dedicated `#define`'s in `GLFW.h`.


mosra:
Merged, thanks!

PR #165, opened 2016-07-24
Added Bezier curve

ashrko619:


src/Magnum/Math/CMakeLists.txt
```diff
@@ -25,6 +25,7 @@
 
 set(MagnumMath_HEADERS
     Angle.h
+	Bezier.h
```
LB--: Mismatched indentation

Squareys:
@ashrko619 First of all: Great job! This already looks very promising :+1: Thank you!

The CIs do not seem to be able to build your code for any other compiler than MinGW, it would be great if you could have a look at what error messages appeared and find out how to fix them `:)`

I will add some more comments inline and others may, too. 
Thank you very much!

ashrko619:
@Squareys thanks :) turns out MinGW works without including array and other compilers throw an error without it- weird :confused: and how to change the older commit message now?

LB--:
> and how to change the older commit message now?

You have to check out the commit, amend it, then rebase the later commit onto it and force your branch to it.

This sequence works on my machine:
```sh
git checkout affc989
git commit --amend # insert "Math: " at start
git cherry-pick c1dc033
git commit --amend # add space after first colon
git branch -f master HEAD
git checkout master
git log # just to double-check
git push --force
```

Squareys:
```
git rebase -i HEAD~2
# change all the 'p' in front of commits you want to reword to 'r'
# reword the commits
git push origin master -f
```

src/Magnum/Math/Bezier.h
```diff
@@ -0,0 +1,118 @@
...
+namespace Magnum { namespace Math {
+        /**
+        @brief Bezier
+        @tparam order       Order of Bezier curve
+        @tparam dimensions  Dimensions of the control points
+        @tparam T           Underlying data type
+
+        See <a href="https://en.wikipedia.org/wiki/B%C3%A9zier_curve">Bezier Curve</a>.
+        */
+        template<UnsignedInt order, UnsignedInt dimensions, class T> class Bezier {
```
Squareys: This seems to be indented too much. See http://mosra.cz/blog/corrade-doc/corrade-coding-style.html#corrade-coding-style-cpp-style for coding style documentation :)

src/Magnum/Math/Bezier.h
```diff
@@ -0,0 +1,118 @@
...
+             * @param t The interpolation factor
+             *
+             */
+            std::array<std::array<Vector<dimensions,T>, order + 1>, order + 1> calculateIntermediatePoints(Float t) const {
+                const auto n = order + 1;
+                std::array<std::array<Vector<dimensions,T>, n>, n> i_points;
+                for (UnsignedInt i = 0; i < n; ++i) {
+                    i_points[i][0] = _points[i];
+                }
+                for (UnsignedInt r = 1; r < n; ++r) {
```
Squareys: Also in the [coding style documentation](http://mosra.cz/blog/corrade-doc/corrade-coding-style.html#corrade-coding-style-cpp-style): magnum code uses `for(...)` rather than `for (...)`, same for `if` and `switch`.

src/Magnum/Math/Test/BezierTest.cpp
```diff
@@ -0,0 +1,81 @@
...
+                explicit BezierTest();
+
+                void testQuadratic();
+
+                void testCubic();
+
+            };
+
+            BezierTest::BezierTest() {
+                addTests({&BezierTest::testQuadratic});
```
Squareys: You can use `addTests` like this:
```
 addTests({&BezierTest::testQuadratic, 
           &BezierTest::testCubic});
```

src/Magnum/Math/Test/BezierTest.cpp
```diff
@@ -0,0 +1,81 @@
...
+    DEALINGS IN THE SOFTWARE.
+*/
+
+#include <Corrade/TestSuite/Tester.h>
+#include <array>
+#include "Magnum/Math/Bezier.h"
+
+namespace Magnum { namespace Math { namespace Test {
+
+            struct BezierTest: Corrade::TestSuite::Tester {
```
Squareys: For reference on indentation see https://github.com/mosra/magnum/blob/master/src/Magnum/Math/Test/DualTest.cpp#L34 for example.
(I know new code styles always take a bit to get used to :stuck_out_tongue_winking_eye: )

src/Magnum/Math/Test/BezierTest.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+            void BezierTest::testCubic() {
+                typedef Math::Vector<2, Float> Vector2;
+                Vector2 p0(0.0f, 0.0f), p1(10.0f, 15.0f), p2(20.0f, 4.0f), p3(5.0f, -20.0f);
+                std::array<Vector2, 4> points = {p0, p1, p2, p3};
+                CubicBezier2D<Float> bezier(points);
+                for(Float t = 0.0; t <= 1.0f; t += 0.01f) {
+                    Vector2 expected = cube(1-t)*p0
+                                       + 3*sqr(1-t)*t*p1
+                                       + 3*(1-t)*sqr(t) *p2
```
Squareys: From the magnum-doc [here](http://mosra.cz/blog/corrade-doc/corrade-coding-style.html#corrade-coding-style-cpp-style):
> Whitespace **should** be around operators with low precedence (namely +, -, <<, >> but not on unary -) [...] Whitespace **shouldn't** be around /, *, % to indicate they have precedence before addition and substraction.

Should therefore be `3*(1 - t)*sqr(t)*p2` for example.

Sorry for being so picky :sweat_smile:

src/Magnum/Math/Bezier.h
Squareys: I see you actually did that correctly up above :smile: You probably just missed it here.

src/Magnum/Math/Test/BezierTest.cpp
```diff
@@ -0,0 +1,81 @@
+/*
+    This file is part of Magnum.
+
+    Copyright © 2010, 2011, 2012, 2013, 2014, 2015, 2016
+              Vladimír Vondruš <mosra@centrum.cz>
```
Squareys: You can add yourself here ;)

[-> Reference](https://github.com/mosra/magnum/blob/master/src/Magnum/Audio/Context.h#L8)

src/Magnum/Math/Test/BezierTest.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+            };
+
+            BezierTest::BezierTest() {
+                addTests({&BezierTest::testQuadratic});
+                addTests({&BezierTest::testCubic});
+            }
+
+            template <class T>
+            inline T sqr(T t){ return t*t; }
```
Squareys: You could also use [Magnum::Math::pow(base, exponent)](http://mosra.cz/blog/magnum-doc/namespaceMagnum_1_1Math.html#a3f13a84c7a1c265727ec7c0c93908cf9)

src/Magnum/Math/Bezier.h
```diff
@@ -0,0 +1,118 @@
...
+
+            /**
+             * @brief Finds the point in the curve for a given interpolation factor
+             *        De Casteljau's algorithm is used.
+             * @param t The interpolation factor
+             */
+            Vector<dimensions, T> lerp (Float t) const {
+                auto i_points= calculateIntermediatePoints(t);
+                return i_points[0][order];
+            };
```
Squareys: This `;` is superfluous.

src/Magnum/Math/Bezier.h
```diff
@@ -0,0 +1,118 @@
...
+                for (UnsignedInt i = 0; i < n; ++i) {
+                    i_points[i][0] = _points[i];
+                }
+                for (UnsignedInt r = 1; r < n; ++r) {
+                    for (UnsignedInt i = 0; i < n - r; ++i) {
+                        i_points[i][r] = (1 - t) * i_points[i][r - 1] + t * i_points[i + 1][r - 1];
+                    }
+                }
+                return i_points;
+            };
```
Squareys: Again, extra `;`  ;)

src/Magnum/Math/Bezier.h
```diff
@@ -0,0 +1,118 @@
...
+        template<UnsignedInt dimensions, class T> using CubicBezier = Bezier <3, dimensions, T>;
+        template<class T> using QuadraticBezier2D = QuadraticBezier<2, T>;
+        template<class T> using QuadraticBezier3D = QuadraticBezier<3, T>;
+        template<class T> using CubicBezier2D = CubicBezier<2, T>;
+        template<class T> using CubicBezier3D = CubicBezier<3, T>;
+
+    }
+
+}
+#endif //Magnum_Math_Bezier_h
```
Squareys: Magnum code usually doesn't annotated `#endif`s, since these kinds of comments are likely to be missed during refactoring (and it's rather obvious what this endif does anyway)

src/Magnum/Math/Test/BezierTest.cpp
```diff
@@ -0,0 +1,81 @@
...
+                std::array<Vector2, 3> points = {p0, p1, p2};
+                QuadraticBezier2D<Float> bezier(points);
+                for(Float t = 0.0; t <= 1.0f; t += 0.01f) {
+                    Vector2 expected = sqr(1-t)*p0 + 2*(1-t)*t*p1 + sqr(t) *p2;
+                    CORRADE_COMPARE(bezier.lerp(t), expected);
+                }
+            }
+
+            void BezierTest::testCubic() {
+                typedef Math::Vector<2, Float> Vector2;
```
Squareys: You can move this typedef to the top of the namespace, since you need it in both tests.

src/Magnum/Math/Test/BezierTest.cpp
```diff
@@ -0,0 +1,81 @@
...
+                Vector2 p0(0.0f, 0.0f), p1(10.0f, 15.0f), p2(20.0f, 4.0f), p3(5.0f, -20.0f);
+                std::array<Vector2, 4> points = {p0, p1, p2, p3};
+                CubicBezier2D<Float> bezier(points);
```
Squareys: Would `CubicBezier2D<Float> bezier({p0, p1, p2, p3});` also work?

Squareys:
@ashrko619 Sorry for the delay! I was rather busy the last few weeks.

Thank you for fixing the CI compile errors! (The remaining errors should not be your fault.)
And also thanks for rewording the commits.

I had some more time to have a closer look at the code, compiled it and ran the tests, looks pretty good!
There were some formatting issues, and a bunch of minor things (some rather picky, sorry!). I commented in the code directly.

src/Magnum/Math/Test/BezierTest.cpp
mosra: Or, even better, [Math::pow<2>(base)](http://mosra.cz/blog/magnum-doc/namespaceMagnum_1_1Math.html#a9083d7f09f7846b593aff7afd01d0122), which gets compiled directly into `t*t` without expensive `std::pow()` calls.

src/Magnum/Math/Bezier.h
```diff
@@ -0,0 +1,118 @@
...
+             */
+            std::array<Bezier<order, dimensions, T >, 2> subdivide(Float t) const {
+                auto i_points = calculateIntermediatePoints(t);
+                std::array<Vector<dimensions, T>, order + 1> left, right;
+                for(UnsignedInt i=0; i<=order; ++i){
+                    left[i] = i_points[0][i];
+                }
+                for(UnsignedInt i = 0, j = order; j>=0; --j, ++i){
+                    right[i] = i_points[i][j];
+                }
```
mosra: While this would probably get optimized and unrolled by the compiler, I would suggest using `std::size_t` instead of `UnsignedInt` for the `i` variable. The reason is that on 64bit targets the 32bit iterating variable goes through quite a lot of conversions from/to machine-sized words and bounds checks, slowing down the code inside. There was one pretty cool article about that on the interwebs, but I can't seem to find it right now.

src/Magnum/Math/Bezier.h
```diff
@@ -0,0 +1,118 @@
...
+                for (UnsignedInt r = 1; r < n; ++r) {
+                    for (UnsignedInt i = 0; i < n - r; ++i) {
+                        i_points[i][r] = (1 - t) * i_points[i][r - 1] + t * i_points[i + 1][r - 1];
+                    }
+                }
+                return i_points;
+            };
+
+            std::array<Vector<dimensions, T >, order + 1> _points;
```
mosra: Would it be possible to use just plain C array here? The size is known at compile-time anyway, so we don't get much more by the additional wrapping done by `std::array` (except for maybe longer compilation times because of all the template instantiation).

```cpp
Vector<dimensions, T> _points[order + 1];
```

src/Magnum/Math/Test/BezierTest.cpp
```diff
@@ -0,0 +1,81 @@
...
+#include "Magnum/Math/Bezier.h"
+
+namespace Magnum { namespace Math { namespace Test {
+
+            struct BezierTest: Corrade::TestSuite::Tester {
+                explicit BezierTest();
+
+                void testQuadratic();
+
+                void testCubic();
```
mosra: picky suggestion: I would use just `quadratic()` and `cubic()` here, as the "test" is already in class name.

src/Magnum/Math/Bezier.h
```diff
@@ -0,0 +1,118 @@
...
+
+            /**
+             * @brief Divides a Bezier curve into two curves of same order having their own control points.
+             *        De Casteljau's algorithm is used.
+             * @param t The interpolation factor
+             *
+             * @return Array of two Bezier curves of the same order
+             */
+            std::array<Bezier<order, dimensions, T >, 2> subdivide(Float t) const {
+                auto i_points = calculateIntermediatePoints(t);
```
mosra: Coding style: please use `iPoints` instead of `i_points`. Similarly in other occurrences.

src/Magnum/Math/Bezier.h
```diff
@@ -0,0 +1,118 @@
...
+        @tparam T           Underlying data type
+
+        See <a href="https://en.wikipedia.org/wiki/B%C3%A9zier_curve">Bezier Curve</a>.
+        */
+        template<UnsignedInt order, UnsignedInt dimensions, class T> class Bezier {
+
+        public:
+
+            /** @brief Construct Bezier curve with the given array of control points */
+            explicit Bezier(const std::array<Vector<dimensions, T>, order+1> &points) {
```
mosra: Would it be possible for you to implement this similarly as in the `Vector::Vector(U...)` constructor? I.e. using variadic templated arguments, forward them to the `_points` array and `static_assert` that their count is right?

src/Magnum/Math/Bezier.h
```diff
@@ -0,0 +1,118 @@
...
+        @tparam order       Order of Bezier curve
+        @tparam dimensions  Dimensions of the control points
+        @tparam T           Underlying data type
+
+        See <a href="https://en.wikipedia.org/wiki/B%C3%A9zier_curve">Bezier Curve</a>.
+        */
+        template<UnsignedInt order, UnsignedInt dimensions, class T> class Bezier {
+
+        public:
+
```
mosra: Please also add a `implicit` `constexpr` default (parameter-less) constructor.

src/Magnum/Math/Test/BezierTest.cpp
```diff
@@ -0,0 +1,81 @@
...
+    IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+    FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+    THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+    LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+    FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+    DEALINGS IN THE SOFTWARE.
+*/
+
+#include <Corrade/TestSuite/Tester.h>
+#include <array>
```
mosra: I think the `<array>` include is not needed, since it is already included in `Bezier.h`.

mosra:
@ashrko619 Great job! :) Even better considering this is a first contribution to an unknown codebase :)

Hope you don't get scared away by the amount of comments @Squareys and me did ;) It's mainly just to align our coding style to the same view.

I put some minor style and API-related comments inline, other than I'm more than happy! (Even though it might not look like that -- I just comment a lot ;))

src/Magnum/Math/Bezier.h
ashrko619: What should the control points be if we have a constructor with no args :question:

src/Magnum/Math/Bezier.h
ashrko619: Why should the constructor be implicit :question:
Also if we have a constructor with variadic template args, how would I create Bezier in subdivide method. 
More specifically how to unpack a std::array ?

This is the first time am using variadic templates ;)

src/Magnum/Math/Bezier.h
mosra: Just default-constructed values (i.e. having the constructor completely empty). And please add a test for such case also.

src/Magnum/Math/Bezier.h
mosra: Implicit to be consistent with other Math classes (Vector, Quaternion, Range, ...). Or, in other words, it's obvious what the constructor does, so there is no need to *require* the user to make the type explicit.

Good question. Will look into it and give an idea later ;)

mosra: Okay, so thought about that a bit: what about using `Bezier<order, dimensions, T>` instead of `std::array<Vector<dimensions,T>, order + 1>` in the return type of `calculateIntermediatePoints()` so it would return `std::array<Bezier<order, dimensions, T>, order + 1>`? If you add `operator[]` accessor to the `Bezier` class, then it has basically the same API as `std::array` so it shouldn't be problem to use it and you also don't have to unpack it in any way to return the final `Bezier` instances.

ashrko619: That's a good idea :) I have come up with a solution to unpack the array using ```std::integer_sequence```. I am not sure how efficient this is https://gist.github.com/ashrko619/5244b9dbbb152a266767e6d3846f163b

What do you think would be a better solution? :)

mosra: The less amount of code the better, I think :) I see no problem in using Bezier in place of `std::array`, so if that works for you, then I would prefer that.

`std::integer_sequence` itself is in the standard library only since C++14, I have equivalent C++11 implementation [here](https://github.com/mosra/magnum/blob/master/src/Magnum/Math/BoolVector.h#L40-L55).

mosra:
Sorry for being silent for so long -- it's now cherry-picked in 6b7097497fd7938dd0aab7baf222e848eee99287. It needed some boilerplate additions such as debug operators etc. (which I didn't want to bother you with) in order to be properly integrated with the test and I hadn't time to do them until now.

Thanks a lot for your work :)

PR #166, opened 2016-07-28
FindSDL2: Properly handle target platform architecture on windows

Squareys:
Hi @mosra !

As discussed, here are the changes for the FindSDL2.cmake module. Let's see what the CI says.

Greetings,
Squareys.

PS: There seems to be a commit attached which I did not do... Did you revert master to a commit before per chance?

### DIFF
--- a/modules/FindSDL2.cmake
+++ b/modules/FindSDL2.cmake
@@ -43,14 +43,19 @@
 if(CORRADE_TARGET_EMSCRIPTEN)
     set(_SDL2_PATH_SUFFIXES SDL)
 else()
+    # Precompiled libraries for Windows are in x86/x64 subdirectories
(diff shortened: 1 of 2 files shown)


Squareys:
@mosra As soon as you declare this as fine, I will copy it over to magnum-examples and do a PR there also.

Squareys:
CIs seem to have properly found SDL2. There is an error on macosx, but that's the "cmake 3.6.0 already installed" one, which was there before.

What's with this commit of yours (946aab3) should I remove it?

mosra:
Heh, but GitHub reports that 946aab3 is here as well ... Weird.

mosra:
Happy with the contents, merged :)

PR #175, opened 2016-09-19
Extend WavAudioImporter to support more formats and non-ideal containers

Alicemargatroid:
A standard WAV header is _not_ guaranteed to be 44 bytes, for the following reasons:

1. The FMT chunk has no less than [three separate formats](http://www-mmsp.ece.mcgill.ca/documents/audioformats/wave/wave.html), of which WavAudioImporter supports only the simplest.
2. The first chunk after the FMT chunk _need not be a DATA chunk_; notably, any non-PCM WAV file is required to have a FACT chunk, which is almost always before the DATA chunk.
3. Reading to the end of the file after the DATA chunk is not guaranteed to be successful; chunks may occur in any section of the WAV container, and often are included at the end.

This means the previous technique of directly loading the header from the raw data only supports the simplest and most naive of WAV files, which represents around 80% of the files it should have been able to load. As well, a large number of valid WAV files could not be easily loaded.

This PR fixes these issues, allowing for a much larger amount of WAV files to be loaded, at only a minor increase in code complexity:

1. The RIFF/Wav header chunk and FMT chunk are dealt with independently, to allow for more flexible loading of non-PCM or non-standard formats.
2. As many sizes as possible are pulled from the chunkSize fields, rather than hard coded.
3. We search for the DATA chunk, and ignore any chunks we do not understand
4. We only read the data indicated, preventing clicks from chunks at the end of the file
5. We include loading for a number of non-PCM formats

We do not:
1. Provide loading for WAV extensions (not yet tested)
2. Provide loading for channels above stereo (not yet implemented)


Squareys:
I just checked out your branch to try it out with some .wav files I recently wanted to load and they appear to contain a `JUNK` chunk instead of the `fmt ` chunk, which seems to contain "nothing" and should be skipped. I will see if I can create a code snippet that handles these chunks and post it here.

src/MagnumPlugins/WavAudioImporter/WavHeader.h
```diff
@@ -31,30 +31,59 @@
 
 #include "Magnum/Types.h"
 
+#define WAVE_FORMAT_PCM 0x0001
```
Alicemargatroid: @Squareys @mosra I used defines as that's what they are in the original C code, and I wanted this to be as "close to original" as possible; we could go with making them entirely internal to the source file if you want, to make them more encapsulated.

Or an enum, or even const int's or something. I don't mind. I'd prefer defines for the legacy reason though.

src/MagnumPlugins/WavAudioImporter/WavImporter.cpp
```diff
@@ -43,74 +43,161 @@ bool WavImporter::doIsOpened() const { return _data; }
...
+        else if(formatChunk.numChannels == 2)
+            _format = Buffer::Format::StereoMuLaw;
+        else {
+            Error() << "Audio::WavImporter::openData(): unsupported channel count"
+                    << formatChunk.numChannels << "with" << formatChunk.bitsPerSample
+                    << "bits per sample";
+            return;
+        }
+    }
+    else if(formatChunk.audioFormat == WAVE_FORMAT_EXTENSIBLE) {
```
Alicemargatroid: @Squareys On it

src/Magnum/Audio/Buffer.h
```diff
@@ -83,7 +89,27 @@ class Buffer {
              *
              * @requires_al_extension Extension @al_extension{EXT,double}
              */
-            StereoDouble = AL_FORMAT_STEREO_DOUBLE_EXT
+            StereoDouble = AL_FORMAT_STEREO_DOUBLE_EXT,
+
+            Quad8 = AL_FORMAT_QUAD8,
```
Alicemargatroid: @Squareys I did add them to Extensions.h; did you miss them?

I can add some notes though, sure.

src/Magnum/Audio/Buffer.h
```diff
@@ -83,7 +89,27 @@ class Buffer {
...
+            Quad16 = AL_FORMAT_QUAD16,
+            Quad32 = AL_FORMAT_QUAD32,
+
+            Rear8 = AL_FORMAT_REAR8,
+            Rear16 = AL_FORMAT_REAR16,
+            Rear32 = AL_FORMAT_REAR32,
+
+            Five_1Chnl8 = AL_FORMAT_51CHN8,
```
Alicemargatroid: @Squareys @mosra Yeah, I had no clue what to go with here. Both of those feel awkward too, but I'll go with whatever. I just want surround sound, dang it!

Alicemargatroid:
@Squareys Are you serious? That's SUPER non-standard haha! The reason I didn't put FMT into the loop was for that reason.

src/Magnum/Audio/Buffer.h
Squareys: @Alicemargatroid I think we mean different headers: you did add them to `OpenAL/extensions.h`, yes, and that is correct, but you will also need to add them to [Magnum/Audio/Extension.h](https://github.com/mosra/magnum/blob/master/src/Magnum/Audio/Extensions.h#L81), which wraps the extensions in namespaces. This helps with code assists and type safety for example.

Squareys:
@Alicemargatroid 

> Are you serious?

Well, I was until now...? (I just want to be able to load my wav files, dang it! ;) ) The files I am trying to load have a JUNK chunk in front of the fmt. Since they were exported from Steinberg Cubase, I did not really question whether that is even allowed. On google I found that fmt being the first chunk is not guaranteed anyway... would you skip the JUNK chunk explicitly in this case? I don't know :)

> That's SUPER non-standard haha!

I am very unsure what you mean with "standard"... (excuse me, if I sound stupid) What is "standard" in this case and why would this not be recommended?

> The reason I didn't put FMT into the loop was for that reason.

I guess it does indeed make sense for fmt not to be in the loop, since this allows it to appear twice or not at all. Would the right way be to look at the header, skip all non fmt, look at fmt and then load all the data chunks? Can't even data chunks appear before fmt? I am a little confused about all of this, maybe that will be cleared up when I know what you meant with "standard".

Thanks in advance ;)

Alicemargatroid:
@Squareys Sorry for the confusion, I was referring to the presence of a JUNK chunk before the FMT chunk; the original RIFF specifications didn't even have a JUNK chunk, to the best of my knowledge, and no known official Microsoft tests that I could find allowed it. All resources I found said that FMT should come after the WAV chunk, but I suppose the JUNK chunk is there to enforce alignment (a reasonable-ish reason).

Your code looks fine; since FMT is not forced to be the second chunk, it makes sense to wiggle the code a bit to ensure we find FMT. I'll modify my own code to account for this shortly, then move onto a DrWav plugin.

Squareys:
> I suppose the JUNK chunk is there to enforce alignment (a reasonable-ish reason).

Yes, I read that, too.

> To reaffirm, the "non-standard" was referring to the .wav files you were talking about.

Alright, *phew* 😅

mosra:
Hello, sorry for being silent the whole week, the work is currently at its craziest peak yet and I literally wasn't doing anything else than just working, eating or sleeping. 

It's amazing how much investigation and attention to detail you put there. My original WavImporter code was just the "minimum viable product" and I also realized very soon that it's not able to load majority of WAV files. The documentation was silent about that limitation and that brought you quite a lot of work, sorry.

I'll now go over the code and try to use the new Pull Request Reviews feature on it :)

src/Magnum/Audio/Buffer.h
Alicemargatroid: @Squareys Got it, fixing now

src/MagnumPlugins/WavAudioImporter/WavHeader.h
mosra: I'd also prefer a (documented) enum to avoid polluting the global namespace. No need to create the Debug operator for it, since this is used only internally (unless you find a good reason to have it, e.g. usage in error messages). Also, since the `WavHeader.h` is internal header, not exposed to plugin users, there's no need to have them internal to the source file.

I'll check with `littleEndianInPlace()`, it should work on enums. If not, I'll make the required changes, because this is a valid use case.

src/MagnumPlugins/WavAudioImporter/WavHeader.h
```diff
@@ -31,30 +31,59 @@
...
+struct WavFactChunk {
+    RiffChunk chunk;
+    UnsignedInt dwSampleLength;     /**< @brief sample length */
+};
+#pragma pack()
+
+#pragma pack(1)
+/** @brief WAV 'data' header */
+struct WavDataChunk {
+    RiffChunk chunk;
```
mosra: Please put some documentation for every new field.

src/MagnumPlugins/WavAudioImporter/WavHeader.h
```diff
@@ -31,30 +31,59 @@
...
+#pragma pack()
 
-    char subChunk1Id[4];            /**< @brief `fmt ` characters */
-    UnsignedInt subChunk1Size;      /**< @brief 16 for PCM */
+#pragma pack(1)
+/** @brief WAV 'fmt' header */
+struct WavFormatChunk {
+    RiffChunk chunk;
     UnsignedShort audioFormat;      /**< @brief 1 = PCM */
     UnsignedShort numChannels;      /**< @brief 1 = Mono, 2 = Stereo */
```
mosra: I guess the documentation for `audioFormat` and `numChannels` is now outdated?

Also, along with the proposed change of `#define`s to `WavFormat` enum, I'd change the type of `audioFormat` to `WavFormat`, to avoid ugly casts in the code.

src/MagnumPlugins/WavAudioImporter/WavImporter.cpp
```diff
@@ -43,74 +43,161 @@ bool WavImporter::doIsOpened() const { return _data; }
...
-    _frequency = header.sampleRate;
+    _frequency = formatChunk.sampleRate;
 
     /** @todo Convert the data from little endian too */
     CORRADE_INTERNAL_ASSERT(!Utility::Endianness::isBigEndian());
 
     /* Copy the data */
-    _data = Containers::Array<char>(header.subChunk2Size);
-    std::copy(data.begin()+sizeof(WavHeader), data.end(), _data.begin());
+    const char* dataChunkPtr = reinterpret_cast<const char*>(++dataChunk);
```
mosra: I would prefer `dataChunk + 1` instead of `++dataChunk`, as it doesn't change the original value.

src/MagnumPlugins/WavAudioImporter/WavHeader.h
Squareys: It is used in an error message, which is why I mentioned it.

mosra: Right. Sorry, overlooked that.

The `Utility::Endianness` class should work on [`enum class` types without problems](https://github.com/mosra/corrade/commit/ed226f7519921db4cd8ca06af6be7e41a8617e2d).

src/MagnumPlugins/WavAudioImporter/WavImporter.cpp
Alicemargatroid: Done

src/MagnumExternal/OpenAL/extensions.h
```diff
@@ -52,6 +52,39 @@ extern "C" {
...
+#endif
+
+/* AL_EXT_ALAW */
+#ifndef AL_EXT_ALAW
+#define AL_EXT_ALAW 1
+#define AL_FORMAT_MONO_ALAW_EXT                  0x10016
+#define AL_FORMAT_STEREO_ALAW_EXT                0x10017
+#endif
+
+#ifndef AL_EXT_MCFORMATS
```
Alicemargatroid: Done

Alicemargatroid:
@Squareys I've added support for files containing chunks between WAV and FMT, based on your mono8junk.wav. Can you verify it works against some of your other files?

src/MagnumPlugins/WavAudioImporter/WavHeader.h
Alicemargatroid: WavImporter does not currently support more than 2 channels (still). The DrWavAudioImporter, however, will.

Alicemargatroid:
@mosra That should be everything besides the enum thing; do you know how to force an enum to be of type and size UnsignedShort? I think keeping it as defined in C is still a good idea, and in no way exposes any unnecessary defines (I've tucked them away into the source file), but I'll do it if required to accept the PR.

Please check over everything again, when you have time; I may have missed something! 
Thank you!

mosra:
Merged in 1897e36b19ea5dc597971e3a5ff0825c49fae078. Thanks a lot!

PR #177, opened 2016-09-30
WavAudioImporter: Add tests to improve coverage

Alicemargatroid:
Extends test coverage to all test files.

I've been extending the tests for DrFlacAudioImporter and the new DrWavAudioImporter, and these provide baseline comparisons.


src/MagnumPlugins/WavAudioImporter/Test/WavImporterTest.cpp
```diff
@@ -172,6 +172,10 @@ void WavImporterTest::invalidFactChunk() {
 
     CORRADE_COMPARE(importer.format(), Buffer::Format::Mono16);
     CORRADE_COMPARE(importer.frequency(), 22050);
+
+    CORRADE_COMPARE_AS(importer.data().slice(0, 8),
```
mosra: Can you use `prefix(8)` instead? Shorter to write :)

Similarly elsewhere.

src/MagnumPlugins/WavAudioImporter/Test/WavImporterTest.cpp
```diff
@@ -172,6 +172,10 @@ void WavImporterTest::invalidFactChunk() {
...
+        Containers::Array<char>::from(-27, -11, -1, -9, 24, -6, 127, -5).slice(0, 8),
```
mosra: There's some unfortunate limitation in finding the common type in the testsuite that I have to fix, but didn't find out how yet so it's kinda annoying to use, sorry.

In this case, can you do the conversion explicitly, i.e.

```cpp
Containers::ArrayView<const char>{Containers::Array<char>::from(-27, -11, -1, -9, 24, -6, 127, -5)}
```

instead of the `slice()`? Similarly elsewhere.

src/MagnumPlugins/WavAudioImporter/Test/WavImporterTest.cpp
```diff
@@ -278,6 +312,10 @@ void WavImporterTest::mono32f() {
 
     CORRADE_COMPARE(importer.format(), Buffer::Format::MonoFloat);
     CORRADE_COMPARE(importer.frequency(), 48000);
+
+    CORRADE_COMPARE_AS(importer.data().slice(0, 16),
+        Containers::Array<char>::from(0, 0, 0, 0, 108, 57, -103, 59, 3, 63, 42, 60, -33, -81, -120, 60).slice(0, 16),
+        TestSuite::Compare::Container);
```
mosra: For the 16-bit, 32-bit and 64-bit formats it might make sense to compare the data as actual types and not as their byte representation. That will also account for potential endianness issues. Currently you have to do the casting manually (a helper function is on my TODO list):

```cpp
Containers::Array<char> data = importer.data();
CORRADE_VERIFY(data.size() >= 4*sizeof(Float));
CORRADE_COMPARE(Containers::ArrayView<Float>{reinterpret_cast<Float*>(data.data()), 4},
    Containers::Array<Float>::from(...),
    TestSuite::Compare::Container);
```

src/Magnum/Audio/Buffer.h
```diff
@@ -58,28 +58,28 @@ class Buffer {
...
-             * @brief 8-bit unsigned mono [μ-Law Compressed Sound Format](https://en.wikipedia.org/wiki/Μ-law_algorithm)
+             * @brief 8-bit mono [μ-Law Compressed Sound Format](https://en.wikipedia.org/wiki/Μ-law_algorithm)
              *
              * @requires_al_extension Extension @al_extension{EXT,MULAW}
              */
             MonoMuLaw = AL_FORMAT_MONO_MULAW_EXT,
 
             /**
-             * @brief 16-bit signed [μ-Law Compressed Sound Format](https://en.wikipedia.org/wiki/Μ-law_algorithm)
+             * @brief 8-bit interleaved [μ-Law Compressed Sound Format](https://en.wikipedia.org/wiki/Μ-law_algorithm)
```
mosra: I was also unsure about "singed"/"unsigned"/"interleaved" for the quad formats so the info there is quite random -- i couldn't find anything about those formats. Do you have any idea?

Alicemargatroid: I asked over at OpenAL Soft: https://github.com/kcat/openal-soft/issues/76

They are all 8 bit (stereo being two interleaved 8 bit samples).

src/MagnumPlugins/WavAudioImporter/Test/WavImporterTest.cpp
Alicemargatroid: @mosra That is, frankly, substantially longer than say:

    Containers::Array<char>::from(-27, -11, -1, -9, 24, -6, 127, -5).prefix(8)

Which also looks much more like:

    importer.data().prefix(8)

src/MagnumPlugins/WavAudioImporter/Test/WavImporterTest.cpp
Alicemargatroid: @mosra I don't really think this would help with endian issues, given floating point endianness does not need to be the same as the rest of the systems endianness!

src/MagnumPlugins/WavAudioImporter/Test/WavImporterTest.cpp
Alicemargatroid: @mosra Done

src/MagnumPlugins/WavAudioImporter/Test/WavImporterTest.cpp
mosra: Wow. That's a new thing to me. I hope it's as common as mixed endian, though :D

src/MagnumPlugins/WavAudioImporter/Test/WavImporterTest.cpp
mosra: Okay, leave it as is. Ideally it should work with just

```cpp
Containers::Array<char>::from(-27, -11, -1, -9, 24, -6, 127, -5)
```

but i'm not quite there yet.

src/MagnumPlugins/WavAudioImporter/Test/WavImporterTest.cpp
Alicemargatroid: @mosra http://www.ietf.org/rfc/ien/ien137.txt

It used to be a lot more prevalent, but I work on machines that use different endianness for floats and ints. I don't know why.

Alicemargatroid:
@mosra That should be everything; please tell me if any more changes are necessary.

mosra:
Cherry-picked as 921a11564d59699948d9c1f42ea8e95a31277fde and d3239b6be62319bfa1e45524363cdec21c12bb50. Thanks a lot!

Minor request for next time: could you prefix commit messages with name of the "module" that the commit belongs to? In this case it was either `Audio:` or `WavAudioImporter`. Makes Git history reading way easier.

Alicemargatroid:
@mosra Can do!

PR #185, opened 2016-11-22
Math: Add rangeFrustum and pointFrustum intersection implementions

Squareys:
Hi @mosra !

I finally got around to extracting the `Math::Geometry::Intersection::*` code from the magnum-extras/octree pullrequest. This is a initial sketch, I will finish it up in the next few minutes (before I let it lying around for half a year again) or tomorrow.

Regards, Squareys.

**TODOs**:
 - [x] Tests for new functions in `Intersection`
 - [x] Latex formulas
 - [x] Frustum test

### DIFF
--- a/src/Magnum/Math/Frustum.h
+++ b/src/Magnum/Math/Frustum.h
@@ -1,0 +1,98 @@
+#ifndef Magnum_Math_Frustum_h
+#define Magnum_Math_Frustum_h
+/*
+    This file is part of Magnum.
+
+    Copyright © 2010, 2011, 2012, 2013, 2014, 2015, 2016
+              Vladimír Vondruš <mosra@centrum.cz>
+    Copyright © 2016 Jonathan Hale <squareys@googlemail.com>
+
+    Permission is hereby granted, free of charge, to any person obtaining a
+    copy of this software and associated documentation files (the "Software"),
+    to deal in the Software without restriction, including without limitation
+    the rights to use, copy, modify, merge, publish, distribute, sublicense,
+    and/or sell copies of the Software, and to permit persons to whom the
+    Software is furnished to do so, subject to the following conditions:
+
+    The above copyright notice and this permission notice shall be included
+    in all copies or substantial portions of the Software.
+
+    THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+    IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+    FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+    THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+    LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+    FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+    DEALINGS IN THE SOFTWARE.
+*/
+
+/** @file
+ * @brief Class @ref Magnum::Math::Frustum
+ */
+
+#include <Corrade/configure.h>
+#include <Corrade/Containers/ArrayView.h>
+#include <Corrade/Utility/Debug.h>
+
+#include "Magnum/Math/Matrix4.h"
+
+namespace Magnum { namespace Math {
+
+/**
+@brief Camera frustum
+
+*/
+template<class T> class Frustum {
+    public:
+
+        /**
+         * @brief Create a frustum from projection matrix
+         */
+        static Frustum<T> fromMatrix(const Matrix4<T>& m) {
+            return Frustum{
+                    mvp.row(3) + mvp.row(0),
+                    mvp.row(3) - mvp.row(0),
+                    mvp.row(3) + mvp.row(1),
+                    mvp.row(3) - mvp.row(1),
+                    mvp.row(3) + mvp.row(2),
+                    mvp.row(3) - mvp.row(2)
+                };
+        }
+
+        /**
+         * @brief Construct frustum from frustum planes
+         */
+        constexpr /*implicit*/ Frustum(const Vector4<T>& left, const Vector4<T>& right, const Vector4<T>& bottom, const Vector4<T>& top, const Vector4<T>& near, const Vector4& far): _data{left, right, bottom, top, near, far} {}
+
+        /**
+         * @brief Raw data
+         * @return One-dimensional array of length `24`.
+         */
+        T* data() { return _data[0].data(); }
+
+        /** @overload */
+        constexpr const T* data() const { return _data[0].data(); }
+
+        /**
+         * @brief The frustum planes
+         *
+         * In order left (index `0`), right (index `1`), bottom (index `1`),
+         * top (index `3`), near (index `4`), far (index `5`).
+         */
+        constexpr Corrade::Containers::StaticArrayView<6, Vector4<T>> planes() const { return _data; }
+
+        /**
+         * @brief Plane at given index
+         *
+         * In order left (index `0`), right (index `1`), bottom (index `1`),
+         * top (index `3`), near (index `4`), far (index `5`).
+         */
+        constexpr Vector4<T> operator[](std::size_t i) const { return _data[i]; }
+
+    private:
+        Vector4<T> _data[6];
+};
+
+}}
+
+#endif
--- a/src/Magnum/Math/Frustum.h
+++ b/src/Magnum/Math/Frustum.h
@@ -35,6 +35,7 @@
 #include <Corrade/Utility/Debug.h>
 
 #include "Magnum/Math/Matrix4.h"
+#include "Magnum/Math/Vector4.h"
 
 namespace Magnum { namespace Math {
 
@@ -50,19 +51,19 @@ template<class T> class Frustum {
          */
         static Frustum<T> fromMatrix(const Matrix4<T>& m) {
             return Frustum{
-                    mvp.row(3) + mvp.row(0),
-                    mvp.row(3) - mvp.row(0),
-                    mvp.row(3) + mvp.row(1),
-                    mvp.row(3) - mvp.row(1),
-                    mvp.row(3) + mvp.row(2),
-                    mvp.row(3) - mvp.row(2)
+                    m.row(3) + m.row(0),
+                    m.row(3) - m.row(0),
+                    m.row(3) + m.row(1),
+                    m.row(3) - m.row(1),
+                    m.row(3) + m.row(2),
+                    m.row(3) - m.row(2)
                 };
         }
 
         /**
          * @brief Construct frustum from frustum planes
          */
-        constexpr /*implicit*/ Frustum(const Vector4<T>& left, const Vector4<T>& right, const Vector4<T>& bottom, const Vector4<T>& top, const Vector4<T>& near, const Vector4& far): _data{left, right, bottom, top, near, far} {}
+        constexpr /*implicit*/ Frustum(const Vector4<T>& left, const Vector4<T>& right, const Vector4<T>& bottom, const Vector4<T>& top, const Vector4<T>& near, const Vector4<T>& far): _data{left, right, bottom, top, near, far} {}
 
         /**
          * @brief Raw data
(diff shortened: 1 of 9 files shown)


src/Magnum/Math/Geometry/Distance.h
```diff
@@ -162,6 +162,34 @@ class Distance {
...
          */
         template<class T> static T lineSegmentPointSquared(const Vector3<T>& a, const Vector3<T>& b, const Vector3<T>& point);
+
+        /**
+         * @brief Distance of point from plane, scaled by the length of the planes normal.
+         * @param point the point
+         * @param plane the plane
+         * @returns a value `> 0` if point is on positive half space,
+         *                  `< 0` if point is on negative half space,
+         *                  `0` if point is on the plane.
```
mosra: I think the obvious parameter names don't need to be explicitly documented, then I would just convert the `@return` into a plain text paragraph (Doxygen would break the indentation anyway).

Similarly for the function below.

src/Magnum/Math/Geometry/Distance.h
```diff
@@ -162,6 +162,34 @@ class Distance {
...
+        /**
+         * @brief Distance of point from plane.
+         * @param point Point
+         * @param plane Plane
+         * @returns `> 0` if point is in front of plane,
+         *          `< 0` if point is behind the plane,
+         *          `0` if point is on the plane.
+         */
+        template<class T> static T pointPlaneUnnormalized(const Vector3<T>& point, const Vector4<T>& plane) {
+            return pointPlaneUnnormalized<T>(point, plane)/plane.xyz().length();
```
mosra: Isn't this infinite recursion? :)

src/Magnum/Math/Geometry/Distance.h
```diff
@@ -162,6 +162,34 @@ class Distance {
...
+         * @param plane the plane
+         * @returns a value `> 0` if point is on positive half space,
+         *                  `< 0` if point is on negative half space,
+         *                  `0` if point is on the plane.
+         *
+         * More efficient than @ref pointPlane() when the planes normal is already
+         * normalized or when testing on which half space the point lies.
+         */
+        template<class T> static T pointPlane(const Vector3<T>& point, const Vector4<T>& plane) {
+            return (plane.xyz()*point).sum() + plane.w();
```
mosra: I would name this `pointPlaneNormalized()` and the other just `pointPlane()` and then add an assertion for normalizedness here. (To make the correct-though-slower the default.)

src/Magnum/Math/Geometry/Distance.h
```diff
@@ -162,6 +162,34 @@ class Distance {
...
+        }
+
+        /**
+         * @brief Distance of point from plane.
+         * @param point Point
+         * @param plane Plane
+         * @returns `> 0` if point is in front of plane,
+         *          `< 0` if point is behind the plane,
+         *          `0` if point is on the plane.
+         */
```
mosra: Besides the above I would reference the other function as a faster alternative to this one.

src/Magnum/Math/Geometry/Distance.h
```diff
@@ -162,6 +162,34 @@ class Distance {
...
+         *
+         * More efficient than @ref pointPlane() when the planes normal is already
+         * normalized or when testing on which half space the point lies.
+         */
+        template<class T> static T pointPlane(const Vector3<T>& point, const Vector4<T>& plane) {
+            return (plane.xyz()*point).sum() + plane.w();
+        }
+
+        /**
+         * @brief Distance of point from plane.
```
mosra: No period after `@brief` docs, please ;) (Doxygen adds one on its own and only somewhere.)

src/Magnum/Math/Geometry/Intersection.h
```diff
@@ -123,8 +125,74 @@ class Intersection {
...
+            return true;
+        }
+
+        /**
+         * @brief Intersection of a range and a camera frustum
+         * @param box Range
+         * @param frustum Frustum planes with normals pointing outwards
+         * @return `true` if the box intersects with the camera frustum.
+         */
+        template<class T> bool rangeFrustum(const Range3D<T>& box, Containers::StaticArrayView<6, const Vector4<T>> frustum);
```
mosra: I thought about this for quite some time -- from what I've seen on various places, there are currently multiple ways to represent the frustum planes:

- `Vector4[6]` (used above)
- `std::vector<Vector4>` (used in the shadows example)
- `std::array<Vector4, 6>`
- ...
- `Containers::StaticArrayView<6, Vector4>` (used in this function)

Neither of these is clearly a winner and converting the representations in order to make it possible to call a function that takes one particular kind is annoying, sometimes impossible. Because of that, I'd like to have a clearly defined `Math::Frustum` type for representing frustums in the engine instead of having a bunch of free functions and letting the user spend costly brain cycles reinventing the storage format every time:

```cpp
namespace Magnum { namespace Math {

template<class T> class Frustum {
    public:
        constexpr /*implicit*/ Frustum(const Vector4<T>& plane0, const Vector4<T>& plane1, const Vector4<T>& plane2, const Vector4<T>& plane3, const Vector4<T>& plane4, const Vector4<T>& plane5): _data{plane0, plane1, plane2, plane3, plane4, plane5} {}
        
        explicit Frustum(const Matrix4<T>& projectionMatrix): _data{...} {
            // I guess such conversion is already done somewhere? is an inverse operation possible?
        }
        
        T* data() { return _data[0].data(); }
        constexpr const T* data() const { return _data[0].data(); }
        
        constexpr Containers::StaticArrayView<6, Vector4<T>> planes() const { return _data; }
        
        constexpr Vector4<T> operator[](std::size_t i) const { return _data[i]; }

        Containers::StaticArray<6, Vector3<T>> corners() const {
            // maybe later? the code is in the shadows example
        }

    private:
        Vector4<T> _data[6];
};

}}
```

There is a bunch of "boilerplate" stuff missing (like debug operators, typedefs in root namespace, conversion constructors etc.), I can add that myself later.

Originally I wanted to put the point/box intersection routines inside the class, but after thinking about that a bit, I'm not really sure -- what do you think? Pros of having member function is that it's all very self-contained (one doesn't need to include another header), cons are that the class might explode later with stuff that not everybody needs (intersection of frustum and lines, AABBs, OBBs, spheres, intersecting stereo frustums with objects...). So I would keep them here I think.

Also, because I assume this functionality will get used quite a lot in quite tight loops, I imagine there could be some additions for batch processing (I guess you remember the [famous Mike Acton review](https://macton.smugmug.com/Other/2008-07-15-by-Eye-Fi/n-xmKDH/i-BrHWXdJ)). So for example a `cull()` function, working similarly to `std::remove()` (receiving mutable array view of boxes, reordering them, returning count of boxes that actually intersect the frustum) and then [optimizing the hell out of it](https://fgiesen.wordpress.com/2010/10/17/view-frustum-culling/). But... not *really* requesting that now, a TODO for later :)

src/Magnum/Math/Geometry/Test/DistanceTest.cpp
```diff
@@ -156,6 +158,20 @@ void DistanceTest::lineSegmentPoint3D() {
                     Constants::sqrt2());
 }
 
+void DistanceTest::pointPlaneUnnormalized() {
+    Vector3 point{1.0f, 1.0f, 1.0f};
+    Vector4 plane{2.0f, 2.0f, 2.0f, 0.0f};
```
mosra: I would use a different value here so the difference between normalized and not normalized implementation is properly tested.

src/Magnum/Math/Geometry/Distance.h
Squareys: In some cases you do not need normalized values, e.g. when just trying to find out whether the point is in front or behind the plane. Hence, forcing normalization here would prohibit that use case.

src/Magnum/Math/Geometry/Distance.h
Squareys: Yeah, forgot to rename something here, it seems

src/Magnum/Math/Geometry/Intersection.h
Squareys: I agree with having the `Frustum` class. And I also agree with 

> So I would keep them [the intersection functions] here [In Intersections.h] I think.

since it is one consistent place to look for intersection functions. If `pointLine, pointPlane, pointAABB etc` were all in that namespace, but `pointFrustum` not, that would cause a world of confusion.

Regarding batch processing: I will skip that for now (you could put it in Projects :) ), since that will most likely be an additional method.

src/Magnum/Math/Geometry/Distance.h
Squareys: Well, it is a plain text paragraph (not a list in case that is what you meant)? It's just easier to read in code this way, I find.

src/Magnum/Math/Frustum.h
```diff
@@ -0,0 +1,97 @@
...
+    IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+    FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+    THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+    LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+    FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+    DEALINGS IN THE SOFTWARE.
+*/
+
+/** @file
+ * @brief Class @ref Magnum::Math::Frustum.
```
mosra: No dot please (sorry) :D

src/Magnum/Math/Geometry/Test/CMakeLists.txt
```diff
@@ -25,3 +25,7 @@
 
 corrade_add_test(MathGeometryDistanceTest DistanceTest.cpp LIBRARIES MagnumMathTestLib)
 corrade_add_test(MathGeometryIntersectionTest IntersectionTest.cpp LIBRARIES MagnumMathTestLib)
+
+set_property(TARGET
+    MathGeometryDistanceTest
+    APPEND PROPERTY COMPILE_DEFINITIONS "CORRADE_GRACEFUL_ASSERT")
```
mosra: Would `target_compile_definitions()` work here also?

Squareys: No idea, I copied that from the other math tests... will try!

src/Magnum/Math/Frustum.h
Squareys: That dot came from angle! ;)

Squareys: ... where there is actually a sentence to be completed. What's the rule here? :)

src/Magnum/Math/Geometry/Test/CMakeLists.txt
mosra: Ah, there it was because I was setting one property for multiple targets -- easier than writing `target_compile_definitions()` ten times.

src/Magnum/Math/Geometry/Intersection.h
```diff
@@ -123,8 +126,83 @@ class Intersection {
...
+        template<class T> static bool pointFrustum(const Vector3<T>& point, const Frustum<T>& frustum);
+
+        /**
+         * @brief Intersection of a range and a camera frustum
+         * @return `true` if the box intersects with the camera frustum.
+         *
+         * Counts for each plane of the frustum how many points of the box lie in
+         * front of the plane (outside of the frustum). If none, the box must lie
+         * entirely outside of the frustum and there is no intersection.
+         * Else, the box is considered as intersecting, even if it is merely corners
```
Squareys: I wonder if it really is "merely the corners overlapping" 🤔

src/Magnum/Math/Geometry/Intersection.h
```diff
@@ -123,8 +126,83 @@ class Intersection {
...
+        if(corners == 8) {
+            ++planes;
+        }
+    }
+
+    if(planes == 6) {
+        return true;
+    }
+
+    // potentially check corners here to avoid false positives!
```
mosra: Is this related to the

> I wonder if it really is "merely the corners overlapping" :thinking:

comment above? Should this be a TODO?

Squareys: Yes, it is related. I don't know whether it should be a TODO. Calculating the Corners of the frustum is non-trivial, so not really efficient. For the Octree stuff it's not required and probably more efficient this way anyway.

I am not entirely sure what this means for accuracy: As far as I understand this will merely include the "corner of frustum overlapping with corner of box" case as an intersection. I will do some research on this and find out.

mosra:
In the process of merging this now -- I just have like 15 TODOs for it, so it will take some time :) Hope that's okay.

mosra:
Cherry-picked in 0062fbf78efca78ced4e9ea59d7d682c47e3c89b and 1f995d1ab2e44dd2cbfd5f5361cd7e031bd6c836.

Thanks a lot (and sorry for the huge delay) :+1:

Squareys:
Thanks! :) Sorry for the huge list of TODOs this opened ;)

mosra:
It was not your fault :D

Issue #101, opened 2015-06-17
Complete animation support


mosra:
This is one thing I wanted to do since the very beginning. I'm slowly implementing various bits here and there. Making this public so we can track the (non)progress better and to have a list of requirements in case someone is willing to help.

-   [ ] Interpolation
    -   [x] N-dimensional linear interpolation (`Math::lerp()`)
    -   [ ] N-dimensional Beziers with interpolation and subdivision (do we need more features?)
    -   [x] Quaternion linear/spherical interpolation (`Math::lerp()`/`Math::slerp()`)
    -   [ ] Dual quaternion linear/spherical interpolation ([original paper](http://dcgi.felk.cvut.cz/home/zara/papers/TCD-CS-2006-46.pdf))
    -   [ ] Screw interpolation (can't find the source now, but it's about decomposing the transformation into axis and angle and then linearly interpolating both)
    -   more obscure interpolation types (logarithmic)?
    -   no, I don't want to have rotation matrix interpolation with renormaliazation (for more information reread the above paper)
    -   other curve types? (OpenGEX has TCB, is it widely used outside of 3Ds Max?)
-   [ ] Class for storing animation clips -- list of *keyframes* (time/value pairs) and animation duration
    -   Templated on value type (interpolating 1D, 2D, 3D vectors, quaternions etc.)
    -   Ability to quickly calculate the value at given time using given interpolator, ability to extrapolate
    -   Preferably stateless
-   [ ] Animation class for storing animation state for given clip
    -   Templated on clip value type
    -   Something similar to what's in `SceneGraph` now, but without the scene graph and virtual calls
    -   Start/stop/pause, play count etc.
    -   Non-linear time interpolation (OpenGEX needs that)
-   [ ] Support in importers
    -   [ ] Transformation animation API -- currently all transformations are combined together into one 4x4 matrix, need to access them also separately if some are animated
    -   [ ] Skinning data API -- list of bones, bone weights in meshes, bone objects etc.
    -   [ ] Skin animation API
    -   [ ] Transformation animation import in OpenGEX
    -   [ ] Skinning data import in OpenGEX
    -   [ ] Skin animation import in OpenGEX
    -   others (light/camera property, texture coordinate animation, color animation) later?
    -   import from other formats? (no, I'm not touching COLLADA ever again)
-   [ ] Basic (transformation) animation example -- preferably a scene imported from a file (or not? in case of more obscure usages?)
-   [ ] Skeletal animation example -- also imported from a file

[...]

PR #191, opened 2017-01-22
[WIP] Keyframe-based animation support

mosra:
~~Result of the weekend hackaton. Fun!~~ Now revived after one and half a year. TODO:

- [x] Basic keyframe picking + interpolation algorithm
- [x] Track and TrackView APIs
  - [x] `duration()` returning calculated duration of a clip, returning `Range1D`, suggesting `Range::join()` for combining (what to do with no keyframes?)
- [x] Correct interpolation of booleans
- [x] ~~at least basic Hemite Spline interpolation, needed to verify design for complex interpolators~~ postponed, design has been verified through a WIP implementation
- [x] ~~A Clip class that handles playing, pausing etc.~~ will be part of Player API
- [x] ~~An Animator~~ A Player class that handles playing multiple tracks, play/pause/stop, repeating
  - [x] ~~what about multiple clips?~~ nope, use multiple players
- [x] Trade::AnimationData
- [x] Trade::ObjectData with separate TRS instead of a matrix

What's postponed to later has been moved to #101.

### DIFF
--- a/doc/changelog.dox
+++ b/doc/changelog.dox
@@ -40,6 +40,11 @@ See also:
 
 @subsection changelog-latest-new New features
 
+@subsubsection changelog-latest-new-animation Animation library
+
+-   New experimental @ref Animation library for keyframe-based animation
+    playback
+
 @subsubsection changelog-latest-new-math Math library
 
 -   Added @ref Math::Intersection::rangeFrustum(),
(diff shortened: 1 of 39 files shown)


mosra:
Merged into master now, finally :tada:

PR #193, opened 2017-02-20
Math: minmax() for a pointer range.

mosra:
Someone please tell me how to do the scalar/vector overload so it works on vector subclasses as well :(


Squareys:
You can use `Math::min()` and `Math::max()`, no? That will handle all the types for you.

PR #194, opened 2017-02-25
Use glfw cmake config if available

Squareys:
Hi @mosra !

As discussed, for static glfw library it can happen that X11 dependency is not linked aswell. The solution we discussed was to use the glfw cmake config files usually installed and fall back to the previous method when that is not available.

This is still WIP, because I did not test if this works yet. You may already have a quick look at it, if you want.

Regards, Jonathan.

### DIFF
--- a/modules/FindGLFW.cmake
+++ b/modules/FindGLFW.cmake
@@ -2,12 +2,14 @@
 # Find GLFW
 # ---------
 #
-# Finds the GLFW library. This module defines:
+# Finds the GLFW library using its cmake config if that exists, otherwise
+# falls back to finding it manually. This module defines:
 #
 #  GLFW_FOUND               - True if GLFW library is found
 #  GLFW::GLFW               - GLFW imported target
 #
-# Additionally these variables are defined for internal usage:
+# Additionally, in case the config was not found, these variables are defined
+# for internal usage:
 #
 #  GLFW_LIBRARY             - GLFW library
 #  GLFW_INCLUDE_DIR         - Root include dir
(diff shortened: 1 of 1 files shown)


Squareys:
@mosra Tested this on windows (built magnum WITH_GLFWAPPLICAITON and made sure Glfw was found without setting `GLFW_INCLUDE_DIR`, which means it must have used the config). We don't really have a test for it though, as in, the library is never linked anywhere.

Squareys:
Alright, tested this by compiling the glfw bootstrap application on Windows and removed that one commit as dicussed via gitter.

Ready to review/merge 👍

modules/FindGLFW.cmake
```diff
@@ -39,44 +41,71 @@
...
-    # the tree and then complains that GLFW::GLFW target was not found,
-    # which it shouldn't. This is reproducible with the base bootstrap
-    # project *and* when using Magnum as CMake subproject. Works with both
-    # 2.8.12 and 3.1, so I'm assuming this is a CMake 3.0 bug. The
-    # workaround is to make the target GLOBAL so it's propagated upwards
-    # the tree unconditionally. For some reason, UNKNOWN targets can't be
-    # marked as GLOBAL, so I'm  biting the bullet and saying the library is
-    # shared -- CMake 3.0 is only on Debian Jessie now and I'm assuming GLFW
-    # comes from system package, which *should be* shared. Hopefully this won't
-    # bite back in the future.
```
mosra: Hmm. Why this is no longer here? Original commit with more details: https://github.com/mosra/magnum/commit/477bf4c9587c6f4c338438602317ed5d4020d1d8

Squareys: Probably lost during a rebase 🙈

Squareys: Fixed and rebased.

mosra:
I merged this (and pushed to master) in b70ac5fa51d7c7074273df5f691c58d616c3ede5, but then immediately discovered that it doesn't work at all for me on Arch (yet again I'm merging too fast, sigh). On Arch the system package has only a NOCONFIG IMPORTED_CONFIGURATION (i.e., neither Debug nor Release) and so the finding failed because there was no library to link to.

In order to fix that I threw away most of the patch and just made `GLFW::GLFW` an `INTERFACE` target that has `glfw` in `INTERFACE_LINK_LIBRARIES`. That's much simpler and also correctly propagates *all* possible target properties (such as compiler flags, which were not propagated by the above patch). The only downside of this is that `INTERFACE` doesn't work on CMake 2.8.12, but people stuck on 2.8.12 (Ubuntu 14.04) probably also don't have GLFW 3 in their repos (and if they build it manually, they can also just download a newer CMake while at it). In case of 2.8.12 it does the usual Find module business, without trying to look for GLFW config.

See e59dcd236a99bc04a29976da999fc0c288a22b98 for details.

Issue #148, opened 2016-06-04
Dropppping unmaintained code


mosra:
Assuming nobody uses the following parts anymore, I propose deprecating and then dropping them:

- [x] `Shapes` namespace. Poorly designed, sparsely implemented and of no real use. This would make #5 and also the `DebugTools::ShapeRenderer` class obsolete. Affects the Bullet and Audio example, which would need to be reworked without these (shouldn't be hard, we could also make use of instancing there) -- deprecated in 8efc6b39e902d01dfac904163ad90385d3007013
  - [x] adapt the Bullet example -- mosra/magnum-examples@11f917f9a99dd16257055d7b64490f123aef87a3
  - [x] adapt the Audio example -- mosra/magnum-examples@8a8b06ebf7a0d2929fc3c19a9500182c17698139
  - [x] deprecate `BulletIntegration::convertShape()` -- mosra/magnum-integration@968fa40c7fe58ae9efced185f91e47d0992589d9
  - [x] Drop everything -- mosra/magnum@aca02f52f550c98c632358b2e5bb8ecda3853b6d, mosra/magnum-integration@e2be55b0a2f65d97b74cf8d9acae10f1ee0cbd93
  - [x] Remove `@collisionoperator` or whatever is it named from Doxyfile and Coding Style docs
- [x] `Platform::GlutApplication` class. Is there any platform that has GLUT but doesn't have SDL2? The class is not receiving any updates anymore and I'm not sure if it even works. -- deprecated in d70f75a179f05920da5dcf09c61dd268476328bf
  - [x] Drop it (#195)
- [x] `ColladaImporter` plugin. Depends on outdated `Qt4`, supports only some features because COLLADA files are impossible to parse and I don't plan to extend it anymore. I propose replacing it with general-purpose importer based on `Assimp` (#146) -- deprecated in mosra/magnum-plugins@35ee1a2665917c0246b02b108086a7ade88d1f0f
  - [x] Drop it -- mosra/magnum-plugins@ced67fce5791fea1ce4b8f477351e4423de86ca8
- [x] The `compatibility` branch, along with support for GCC 4.4, 4.5, 4.6 and MSVC 2013.

wivlaro:
I still use collada importer. But can probably move to opengex at some
point. I think assimp is quite good iirc, only concern would be it being a
bit lowest-common-feature set.

Squareys:
> Affects bullet example, which would need to be reworked

[...]

PR #195, opened 2017-02-25
Removed glutapplication class and references

Bluer01:
This refers to issue #148 and removes both the header and source file for glutapplication and also removes any mention found of glutapplication from other files.


mosra:
After a period of deprecation, finally dropped in 035f644769742d00af800f33af2b586a389608c2. And, even though two years later, you get [a credit](https://github.com/mosra/magnum/blob/035f644769742d00af800f33af2b586a389608c2/doc/credits.dox#L102-L103) for it, @Bluer01 :)

PR #196, opened 2017-03-20
Fix href to OpenGL reference generated by Doxygen.

emgre:
The doc was moved to khronos.org and the URL schema was changed. This is just a minor change, but these links are very handy while developing.

By the way, Magnum is a really great library. Keep up the good work!

Émile Grégoire


mosra:
Thank you a lot!

I discovered that also all other links to extensions etc. were being redirected, so I did a bigger overhaul of all the GL specification links, it's now all a bit cleaner inside.

Upload the updated docs right now.

emgre:
There are still redirects from http to https for fn_gl and fn_gl2. My bad...

mosra:
Will update that as well, thanks for the heads up :)

PR #197, opened 2017-03-23
Fix issue in Primitives::Cylinder::solid() with cap ends

emgre:
Fixed issue in Primitives::Cylinder::solid() where the last ring of faces wouldn't be created when using cap ends because of wrong offset.

See discussion here: https://gitter.im/mosra/magnum/archives/2017/03/23


mosra:
I merged and patched the tests. Thanks a lot for your effort! Also first time I see Gitter archives, *very* useful! :)

PR #200, opened 2017-04-01
Modernize Magnum

alicemargatroid:
This is a minor commit that meticulously modernizes Magnum with the following changes:

1. No moving elements that are better simply copied
2. push_back -> emplace_back when appropriate
3. Use defaulted constructor/destructors, not trivial ones
4. Use more modern initialization
5.  Remove unnecessary quantifiers / add ones where appropriate

Feel free to take what you like and leave what you don't!

### DIFF
--- a/src/MagnumExternal/OpenGL/GL/flextGL.cpp
+++ b/src/MagnumExternal/OpenGL/GL/flextGL.cpp
@@ -30,7 +30,7 @@ FLEXTGL_EXPORT void(APIENTRY *flextglMultiDrawArraysIndirectCountARB)(GLenum, GL
 FLEXTGL_EXPORT void(APIENTRY *flextglMultiDrawElementsIndirectCountARB)(GLenum, GLenum, GLintptr, GLintptr, GLsizei, GLsizei) = nullptr;
 
 /* GL_ARB_robustness */
-FLEXTGL_EXPORT GLenum(APIENTRY *flextglGetGraphicsResetStatusARB)(void) = nullptr;
+FLEXTGL_EXPORT GLenum(APIENTRY *flextglGetGraphicsResetStatusARB)() = nullptr;
 FLEXTGL_EXPORT void(APIENTRY *flextglGetnCompressedTexImageARB)(GLenum, GLint, GLsizei, void *) = nullptr;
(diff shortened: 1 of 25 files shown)


src/MagnumExternal/OpenGL/GL/flextGL.cpp
```diff
@@ -30,7 +30,7 @@ FLEXTGL_EXPORT void(APIENTRY *flextglMultiDrawArraysIndirectCountARB)(GLenum, GL
 FLEXTGL_EXPORT void(APIENTRY *flextglMultiDrawElementsIndirectCountARB)(GLenum, GLenum, GLintptr, GLintptr, GLsizei, GLsizei) = nullptr;
 
 /* GL_ARB_robustness */
-FLEXTGL_EXPORT GLenum(APIENTRY *flextglGetGraphicsResetStatusARB)(void) = nullptr;
+FLEXTGL_EXPORT GLenum(APIENTRY *flextglGetGraphicsResetStatusARB)() = nullptr;
```
mosra: These files are autogenerated by flextGL, so this is just what the GL XML spec files say. I could change flextGL to don't put `void` there, but I don't think it's worth the effort :)

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
```diff
@@ -53,7 +53,7 @@ void ignoreLine(std::istream& in) {
     in.ignore(std::numeric_limits<std::streamsize>::max(), '\n');
 }
 
-template<std::size_t size> Math::Vector<size, Float> extractFloatData(std::string str, Float* extra = nullptr) {
+template<std::size_t size> Math::Vector<size, Float> extractFloatData(const std::string& str, Float* extra = nullptr) {
```
mosra: Wow, why am I always doing the stupidest things in the hottest code paths? :D Thanks, this makes the plugin suddenly ∞ times faster as well.

mosra:
Thank you! Much appreciated.

I took almost all of it except the changes to auto-generated GL headers and function loaders.

mosra:
Oh and congrats for having number 200! :) :tada: :sparkles:

PR #203, opened 2017-05-16
ci: Explicitly set travis distro to trusty (Fix travis builds)

Squareys:
Hello everyone,

I just found the magnum builds to be failing on travis, because all of the sudden cmake is only version 2.8.7 there. I am guessing it may be using packages ffrom ubuntu percise instead of trusty, maybe because of default that changed on travis-ci.

I figured explicitly definiting `distro: trusty` may help. Let's see what the build says.

Cheers, Jonathan.

### DIFF
--- a/package/ci/travis-emscripten.sh
+++ b/package/ci/travis-emscripten.sh
@@ -42,8 +42,8 @@ cmake .. \
     -DCMAKE_TOOLCHAIN_FILE="../toolchains/generic/Emscripten.cmake" \
     -DEMSCRIPTEN_PREFIX=$(echo /usr/local/Cellar/emscripten/*/libexec) \
     -DCMAKE_BUILD_TYPE=Release \
-    -DCMAKE_CXX_FLAGS_RELEASE="-DNDEBUG -O1" \
-    -DCMAKE_EXE_LINKER_FLAGS_RELEASE="-O1" \
+    -DCMAKE_CXX_FLAGS_RELEASE="-DNDEBUG -O2" \
+    -DCMAKE_EXE_LINKER_FLAGS_RELEASE="-O2" \
     -DCMAKE_INSTALL_PREFIX=$HOME/deps \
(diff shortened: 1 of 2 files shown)


Squareys:
Instead, linux builds now fail with:
```
CMake Error at CMakeLists.txt:269 (message):

  The RandR library and headers were not found
```
while building glfw dependency. They weren't able solve this issue yet either, their build is also failing because of the cmake version.

Squareys:
I found [this travis blog post](https://blog.travis-ci.com/2017-05-04-precise-image-updates), which writes about an image update that happened around-ish the time the cronjobs started failing. The latest commit tries going back to the previous images by specifying ```group: deprecated-2017Q1```

Squareys:
Tadaaa 👏 🎉 , looking good again.

Squareys:
Note: We may need to add `libxi-dev` package when upgrading to a newer glfw version in the future, see https://github.com/glfw/glfw/pull/1015.

Squareys:
Android builds are now failing because of: https://github.com/travis-ci/travis-ci/issues/5370

> The Trusty image currently lacks Android capabilities.

mosra:
Thanks a lot for all the investigation!

I *really* want to retain the Android builds, so I guess for them I need to stay on 12.04 + download cmake some other way. The Emscripten build is broken since Homebrew updated to 1.37.10 -- I have to investigate that locally, probably some misoptimization again.

Is it okay if I take your commits and continue on this myself?

Squareys:
> Is it okay if I take your commits and continue on this myself?

Of course! (As discussed via gitter)
You should remove 2fdc897 though.

mosra:
Cherry-picked in 3268c5d4304826dc5feda1e777669950cb0ebf0c with some modifications in order to get the Android build running as well. Thanks a lot!

Squareys:
No problem, thank you! :)

PR #205, opened 2017-06-08
[WIP] Rework ObjImporter to support more features

Squareys:
Hello everybody!

I started reimplementing the ObjImporter pretty much a year ago for a University project, since I needed to import .obj files with materials.

While back then I only needed to support the features that I needed for those specfic .obj files, I decided to try to also implement support for all the previously supported features to hopefully get this into upstream someday.

**Added features include:**
 * Materials (`usemtl` and `mtllib` keywords)
 * Groups (`g` keyword)
 * Mixed primitive meshes
 * Zero-copy parsing (almost at least?)

Currently alot of tests are failing. The new implementation seems to be especially bad at failing for bad files still. There is quite alot to do still and may take another couple of days for me to get into a reviewable state.

Cheers, Jonathan.


Squareys:
Oh, I guess I should proabably rebase the branch sometime again :D

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
```diff
@@ -28,81 +28,338 @@
...
+            endpos = endpos.suffix(i);
+        }
     }
 
-    if(data.size() == size+1) {
-        /* This should be obvious from the first if, but add this just to make
-           Clang Analyzer happy */
-        CORRADE_INTERNAL_ASSERT(extra);
+    return std::make_tuple(indices, endpos);
+}
```
mosra: I think this could be all made much less complex and less prone to file errors by doing just a single pass over the data instead of first trying to find the delimiter using a highly-parametrized `findNext()` and then trying to parse everything before that as a number, because that way you can very well end up with half of a broken file trying to be parsed as a number ;) Or, on the other hand, index list like `378baaad/1/5` would be parsed as 378, 1, 5 without any error reported because the `baaad` part gets skipped by `findNext()`.

So, the parsing could look like this (very similar to what the ODDL parser is doing):

1. Ignore all whitespace from the front of the line
2. Branch on first non-whitespace character
3. If it's a `f`, for example, this is a face
4. Ignore all whitespace until next character
5. Because this is `f` line, next should be a number. Eat all numeric chars and parse them as a number. If there are none, end with an error.
6. Next character after the parsed number should be either a space or `/`, branch based on that. If it's not, end with an error.
7. Again expect numeric chars next, parse them.
8. After a valid sequence of face indices is parsed, next should be just whitespace until EOL. If not, emit an error (for example -- "too many vertices per face" or "garbage at EOL" etc.).
9. Ignore `\r` or `\n` characters.
10. Start from 1 again.

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
```diff
@@ -115,387 +372,575 @@ ObjImporter::~ObjImporter() = default;
 
 auto ObjImporter::doFeatures() const -> Features { return Feature::OpenData; }
 
-void ObjImporter::doClose() { _file.reset(); }
+bool ObjImporter::doIsOpened() const { return _in; }
 
-bool ObjImporter::doIsOpened() const { return !!_file; }
+void ObjImporter::doClose() { _in = nullptr; }
```
mosra: Why are these two swapped? Unneeded change.

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
```diff
@@ -115,387 +372,575 @@ ObjImporter::~ObjImporter() = default;
...
-    std::unique_ptr<std::istream> in{new std::ifstream{filename, std::ios::binary}};
-    if(!in->good()) {
-        Error() << "Trade::ObjImporter::openFile(): cannot open file" << filename;
-        return;
-    }
-
-    _file.reset(new File);
-    _file->in = std::move(in);
-    parseMeshNames();
+    _fileRoot = filename.substr(0, filename.find_last_of('/')+1);
```
mosra: Uh, there is a bunch of Directory functions for this ;)

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
```diff
@@ -28,81 +28,338 @@
...
+    Int specularTexture = -1;
+};
+
+struct ObjObject;
+struct ObjGroup;
+struct ObjMesh;
+
+struct ObjMeshData {
+    ObjGroup& group;
+    int materialId; /* from 'usemtl' keyword */
```
mosra: Would be *really* nice if you could use Magnum typedefs for basic types (`Int`, `Float`) .. consistency is king ;)

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
```diff
@@ -28,81 +28,338 @@
...
+// Different description: Get string until next '\n' and return ArrayView of data
+// after it.
+std::tuple<ArrayView<char>, ArrayView<char>> nextLine(const ArrayView<char>& pos) {
+    int i = Math::min(findNext(pos, '\n'), findNext(pos, '\r'));
+    if(i == -1)         {
+        i = pos.size();
+        return std::make_tuple(pos.prefix(i), pos.suffix(i));
+    }
+    return std::make_tuple(pos.prefix(i), pos.suffix(i+1));
+}
```
mosra: What about using `std::pair` for two-element tuples? The semantics is the same (`std::tie()` works there as well), the compiler should have less work and you can use just `{}` instead of `std::make_tuple()`

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
```diff
@@ -28,81 +28,338 @@
...
+        std::tie(word, pos) = nextWord(pos);
+        v[i] = strToFloat(word);
     }
 
-    return output;
+    return pos;
 }
 
-template<class T> void reindex(const std::vector<UnsignedInt>& indices, std::vector<T>& data) {
+template<class T> std::vector<T> reindex(const std::vector<UnsignedInt>& indices, std::vector<T>& data) {
```
mosra: I would then make the `data` parameter a `const` reference.

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
```diff
@@ -115,387 +372,575 @@ ObjImporter::~ObjImporter() = default;
...
+            }
+            mesh = meshData->points;
+        } else if(sectionPrimitive == 'l') {
+            if(!meshData->lines) {
+                // TODO: C++ 17
+                _state->meshes.emplace_back(new ObjMesh{*meshData, MeshPrimitive::Lines});
+                meshData->lines = _state->meshes.back().get();
+            }
+            mesh = meshData->lines;
+        } else { /* sectionPrimitive == 'f' */
```
mosra: Just to be sure, I'd put an explicit `if(sectionPrimitive == 'f')` here and an additional `else CORRADE_ASSERT_UNREACHABLE();` at the end.

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
```diff
@@ -115,387 +372,575 @@ ObjImporter::~ObjImporter() = default;
...
+        }
+    };
+
+    while(!pos.empty()) {
 
         /* Comment line */
-        if(_file->in->peek() == '#') {
-            ignoreLine(*_file->in);
+        if(pos[0] == '#') {
+            finishSection();
```
mosra: Um. Comment finishes a section? What's a section, anyway? Is it a separate mesh?

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
```diff
@@ -28,81 +28,338 @@
...
     }
 
-    if(data.size() == size+1) {
-        /* This should be obvious from the first if, but add this just to make
-           Clang Analyzer happy */
-        CORRADE_INTERNAL_ASSERT(extra);
+    return std::make_tuple(indices, endpos);
+}
+
+template<UnsignedInt D>
```
mosra: lowercase names for non-type template parameters, please .. and in this case saying `dimensions` would make things a bit clearer :)

also would be nice to have this together with function name on one line and `parseVector` instead of `getVector` for better descriptive name :)

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
```diff
@@ -115,387 +372,575 @@ ObjImporter::~ObjImporter() = default;
...
 
-            /* This is the name of first mesh */
-            if(thisIsFirstMeshAndItHasNoData) {
-                thisIsFirstMeshAndItHasNoData = false;
+        /* Vertex position */
+        if(keyword == "v") {
+            // TODO C++17: Could be one slick line with emplace_back ref return.
+            Vector3 v;
+            pos = getVector<3>(pos, v);
+            _state->positions.push_back(v);
```
mosra: What about ~~`getVector`~~ `parseVector()` returning `Vector3` and modifying `pos` via a reference? Then it could fit on one line ;)

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
```diff
@@ -115,387 +372,575 @@ ObjImporter::~ObjImporter() = default;
...
+                ensureGroup();
+                meshData = &group->meshDataForMaterial(materialIndex);
+            } // else: usemtl did not result in material switch, no need to create new mesh
+
+        /* Ignore unsupported keywords, error out on unknown keywords */
+        } else {
+            Warning() << "Trade::ObjImporter::parse(): unknown keyword:" << keyword;
+        }
+
+        if(sectionEnd) {
```
mosra: This boolean is set to `false` only on one place, couldn't the state be inferred from something else? For example a zero count of vertices/faces/... in current section?

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
```diff
@@ -115,387 +372,575 @@ ObjImporter::~ObjImporter() = default;
...
+            finishSection();
         }
 
         /* Ignore the rest of the line */
-        ignoreLine(*_file->in);
+        line = pos = skipWhitespaces(ignoreLine(pos));
```
mosra: I'm confused. Does this ignore the rest of the line or also a bunch of whitespace on next line? Whitespace on line start is ignored above already.

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
```diff
@@ -115,387 +372,575 @@ ObjImporter::~ObjImporter() = default;
...
+            std::tie(name, pos) = nextWord(pos);
+            group->name = std::string{name.data(), name.size()};
+
+        /* Load a material library */
+        } else if(keyword == "mtllib") {
+            ArrayView<char> word;
+            pos = skipWhitespaces(pos);
+            std::tie(word, pos) = nextWord(pos);
+
+            parseMaterialLibrary(word);
```
mosra: From what I understand, just opening the file basically parses all of it and bails out if there is an error in any part of the file or if an external `*.mtl` file can't be found, even though I might not ever need that.

Isn't there a way to first just do a quick look over the file, remember mapping between names and IDs and do the whole numeric data parsing only on-demand when a particular mesh/object/material is requested?

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
```diff
@@ -115,387 +372,575 @@ ObjImporter::~ObjImporter() = default;
...
+        /* Ignore the rest of the line */
+        pos = skipWhitespaces(ignoreLine(pos));
+    }
+}
 
-                primitive = MeshPrimitive::Triangles;
+UnsignedInt ObjImporter::doMesh3DCount() const { return _state->meshes.size(); }
 
-            } else CORRADE_ASSERT_UNREACHABLE(); /* LCOV_EXCL_LINE */
+UnsignedInt ObjImporter::doMaterialCount() const { return _state->materials.size(); }
```
mosra: Just for consistency with other importers I would group all `doMesh*()` functions together, all `doMaterial*()` functions together, etc.

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
```diff
@@ -115,387 +372,575 @@ ObjImporter::~ObjImporter() = default;
...
+    }
+
+    return -1;
+}
+
+std::string ObjImporter::doObject3DName(UnsignedInt id) {
+    const size_t numMeshes = _state->meshes.size();
+    if(id >= numMeshes) {
+        return _state->meshlessObjects[id - numMeshes];
+    }
```
mosra: :thinking:  what about inverting this and having the names just for objects instead of just for meshes? Then there would be no need for `meshlessObjects` -- if index of an object would be larger than size of `meshes`, it would mean there's no corresponding mesh.

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
```diff
@@ -115,387 +372,575 @@ ObjImporter::~ObjImporter() = default;
...
+    return _state->meshes[id]->name(); /* Intentional, objects are just meshes + material */
+}
+
+std::unique_ptr<ObjectData3D> ObjImporter::doObject3D(UnsignedInt id) {
+    const size_t numMeshes = _state->meshes.size();
+    if(id > numMeshes) {
+        return std::unique_ptr<ObjectData3D>{new ObjectData3D{{}, {}, &_state->meshlessObjects[id - numMeshes]}};
```
mosra: The state class is not exposed to public headers at all -- does it make sense to include the pointer to it?

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
```diff
@@ -115,387 +372,575 @@ ObjImporter::~ObjImporter() = default;
...
+    arrays.emplace_back(positionIndices);
+    if(hasNormals) arrays.emplace_back(normalIndices);
+    if(hasTexCoords) arrays.emplace_back(textureCoordinateIndices);
+    indices = MeshTools::combineIndexArrays(arrays);
+
+    /* Reindex data arrays */
+    try {
+        positionLayers.push_back(reindex(positionIndices, _state->positions));
+        if(hasNormals) normalLayers.push_back(reindex(normalIndices, _state->normals));
+        if(hasTexCoords) texCoordLayers.push_back(reindex(textureCoordinateIndices, _state->texCoords));
```
mosra: Sorry, the code was bad to begin with :/

I would just remove the `throw 0` from `reindex()` and return zero-sized vector in that case, then you can get rid of the `try` block completely.

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
```diff
@@ -115,387 +372,575 @@ ObjImporter::~ObjImporter() = default;
...
     }
 
-    return MeshData3D{*primitive, std::move(indices), {std::move(positions)}, std::move(normals), std::move(textureCoordinates), {}, nullptr};
+    return std::unique_ptr<AbstractMaterialData>(mat);
 }
 
+std::optional<ImageData2D> ObjImporter::doImage2D(UnsignedInt id) {
+    CORRADE_ASSERT(manager(), "Trade::ObjImporter::image2D(): the plugin must be instantiated with access to plugin manager in order to open image files", {});
+
+    std::unique_ptr<AbstractImporter> imageImporter = manager()->loadAndInstantiate("TgaImporter"); // probably AnyImageImporter would be the way to go here...
```
mosra: Design hole. Sorry about this. Putting this on my TODO list.

src/MagnumPlugins/ObjImporter/Test/Test.cpp
```diff
@@ -406,7 +427,6 @@ void ObjImporterTest::wrongFloat() {
 
     std::ostringstream out;
     Error redirectError{&out};
-    CORRADE_VERIFY(!importer.mesh3D(id));
```
mosra: `CORRADE_EXPECT_FAIL` at least, please don't remove tests just because they currently fail ;)

src/MagnumPlugins/ObjImporter/Test/Test.cpp
```diff
@@ -716,7 +761,6 @@ void ObjImporterTest::unsupportedKeyword() {
     CORRADE_COMPARE(data->positions(0), (std::vector<Vector3>{
         {0.0f, 1.0f, 2.0f}
     }));
-    CORRADE_COMPARE(data->indices(), std::vector<UnsignedInt>{0});
```
mosra: Why?

src/MagnumPlugins/ObjImporter/Test/Test.cpp
```diff
@@ -702,6 +722,31 @@ void ObjImporterTest::wrongTextureCoordinateIndexCount() {
...
+    CORRADE_VERIFY(data);
+    CORRADE_COMPARE(data->primitive(), MeshPrimitive::Triangles);
+    CORRADE_COMPARE(data->positionArrayCount(), 1);
+    CORRADE_COMPARE(data->positions(0), (std::vector<Vector3>{
+        {Vector3(1.72414, 18.9233, -3.20162),
+         Vector3(2.74428, -0.499733, -3.50576),
+         Vector3(-1.92235, -0.846268, 2.9722),
+         Vector3(1.72414, 18.9233, -3.20162),
+         Vector3(-1.92235, -0.846268, 2.9722),
+         Vector3(2.43556, 18.8755, 2.23745)}
```
mosra: `f` suffix please, then you can just use `{}` without the explicit class name ;)

src/MagnumPlugins/ObjImporter/Test/Test.cpp
```diff
@@ -311,30 +333,25 @@ void ObjImporterTest::textureCoordinatesNormals() {
...
     }));
 }
 
 void ObjImporterTest::emptyFile() {
     ObjImporter importer;
     CORRADE_VERIFY(importer.openFile(Utility::Directory::join(OBJIMPORTER_TEST_DIR, "emptyFile.obj")));
-    CORRADE_COMPARE(importer.mesh3DCount(), 1);
-}
-
-void ObjImporterTest::unnamedMesh() {
```
mosra: Why is this removed? Then there's no test case that tests that `mesh3DForName("")` returns -1

src/MagnumPlugins/ObjImporter/Test/Test.cpp
```diff
@@ -83,6 +82,8 @@ struct ObjImporterTest: TestSuite::Tester {
     void missingNormalIndices();
     void missingTextureCoordinateIndices();
 
+    void multiMaterialObject();
+
```
mosra: Is there a test case for the actual material *data* import?

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
```diff
@@ -115,387 +372,575 @@ ObjImporter::~ObjImporter() = default;
...
-        indices = std::move(positionIndices);
-        for(UnsignedInt i: indices) if(i >= positions.size()) {
-            Error() << "Trade::ObjImporter::mesh3D(): index out of range";
-            return std::nullopt;
-        }
+        mat->specularTexture() = objMat.specularTexture;
     }
 
-    return MeshData3D{*primitive, std::move(indices), {std::move(positions)}, std::move(normals), std::move(textureCoordinates), {}, nullptr};
+    return std::unique_ptr<AbstractMaterialData>(mat);
```
mosra: If the material would be fully parsed in this function instead of in `parse()`, you wouldn't need the temporary `objMat` structure I think.

I hope I'm not talking about impossible things here, though ;)

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
```diff
@@ -115,387 +372,575 @@ ObjImporter::~ObjImporter() = default;
...
+    while(!pos.empty()) {
 
-    try { while(_file->in->good() && _file->in->tellg() < end) {
-        /* Ignore comments */
-        if(_file->in->peek() == '#') {
-            ignoreLine(*_file->in);
+        /* Comment line */
+        if(pos[0] == '#') {
+            pos = ignoreLine(pos);
+            pos = skipWhitespaces(pos);
```
mosra: Again, this is done right in the next loop iteration, no need to do this here as well.

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
```diff
@@ -115,387 +372,575 @@ ObjImporter::~ObjImporter() = default;
...
+        section = {nullptr};
+        minSectionPrimitives = 0;
+    };
+
+    auto finishObject = [&]{
+        if(object && object->groups.empty()) {
+            /* Create dummy mesh for this object so that it gets loaded as ObjectData */
+            _state->meshlessObjects.push_back(object->name);
+        }
+    };
```
mosra: I'm a bit unsure about these inline lambdas -- i'm personally always trying to do such cleanup code only once in the parsing loop, because otherwise the corner-case testing gets quite out of hand. But I didn't stare at the code long enough yet to see how that could be done :)

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
```diff
@@ -28,81 +28,338 @@
 #include <fstream>
 #include <limits>
 #include <sstream>
+#include <tuple>
+#include <stdlib.h>
```
Squareys: I think that was because of [strtol](http://www.cplusplus.com/reference/cstdlib/strtol/) and strtof

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
```diff
@@ -28,81 +28,338 @@
...
+#include "Magnum/Math/Vector3.h"
 #include "Magnum/Math/Color.h"
 #include "Magnum/Trade/MeshData3D.h"
+#include "Magnum/Trade/MeshObjectData3D.h"
+#include "Magnum/Trade/ImageData.h"
+#include "Magnum/Trade/PhongMaterialData.h"
+
+#include "MagnumPlugins/TgaImporter/TgaImporter.h"
+
+using namespace Corrade::Containers;
```
Squareys: :thinking: But it makes alot of things more convenient... Aaaand it's a "private" file? Obviously explicitness is oftenly nicer, but is there any other reason for this?

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
```diff
@@ -28,81 +28,338 @@
...
+
+        if((pos[i] == '\n' && termByNewline) || (pos[i] == ' ' && termByWhitespace)) {
+            return -1;
+        }
+    }
+    return -1;
+}
+
+// find next non-whitespace char and return suffix at that point.
+// param newlineIsWhitespace if "true", '\n' and '\r' are skipped also
```
Squareys: I know! :D It is a temporary comment, so I don't forget to put decent wording for other people there.

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
```diff
@@ -28,81 +28,338 @@
...
+        if(c != ' ' && c != '\t' && c != '\0' && (!newlineIsWhitespace || (c != '\n'  && c != '\r'))) {
+            return pos.suffix(i);
+        }
+    }
+    return {};
+}
+
+// returns suffix after next '\n'
+ArrayView<char> ignoreLine(const ArrayView<char> pos) {
+    return pos.suffix(findNext(pos, '\n') + 1);
```
Squareys: Well, then the equivalent to pos is returned, because there is not such thing as a next ' \n'. The code expects this function to work that way and I even thought about this because of a test file not having an newl. at eof. (not the one you pointed out, I think)

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
Squareys: ¯\_(ツ)_/¯ Will revert 😄

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
Squareys: This will not be easy, will stare at the code for a while and tell you what I think after.

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
Squareys: A section is a block of lines with the same primitive keyword to be parsed in doMesh3D later. e.g.

```
p 1
p 2
# section ended at beginning of this line, excluding the comment, next beginns beginning of next line
f 1/..
f 2/..
# section end again
f 1/..
f 2/..
```
which makes the parsing of the indices easier later.

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
Squareys: No, only skips until after the next `\n` Which is why I then skip preceding whitespaces of the next line.

> Whitespace on line start is ignored above already.

Not that I know of? Only for `#`, this line is duplicated above, yes. The other ignore whitespaces is ignoring after the keyword, e.g. `p·····2` will lead to keyword `f` and `pos == "2"`

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
Squareys: Again, no, the ignore Whitespaces ignores words before the keyword (`nextWord()` expects first character to be non-whitespace, otherwise will return empty string, indicating there is no word at the beginning), versus the other one skips whitespaces *after* the keyword.

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
Squareys: Names for objects vs names for objects with meshes are retrieved differently, because for objects with meshes, the name is generated from the intermediate `ObjMesh` structure, dependent on primitive, group and object name and whether there are other primitives in the same group and other groups etc, for potentially ommiting those from the mesh/object name.

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
Squareys: One day it may allow someone to replicate the structure by hand from the code and reinterpret_cast it ;P 

Nah, but seriously, I don't think it does any harm, right? After all, if anybody *really* wants to do the above, then we don't provide any API promises in that sense anyway.

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
Squareys: Very nice, I didn't think about this much, but the try catch really annoyed me

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
Squareys: Sure, I guess it would be enough to only parse the material names in `parse()` and maybe store their location in `ArrayView<const char>`s instead

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
Squareys: Would it be possible to move Any*Importers into the magnum repository? Would that even make sense?

src/MagnumPlugins/ObjImporter/Test/multiMaterial.mtl
```diff
@@ -0,0 +1,8 @@
+newmtl mat_0
+Ka 1 1 1
+map_Ka Textures/mat_0.tga
+
+
+newmtl mat_1
+Ka 1 1 1
+map_Ka Textures/mat_1.tga
```
Squareys: But then we never test files missing newlines at end?
Just saying, I think I found a bug in `parse()`, because another file was missing that. I can extract that into separate files, though, which are explicitly named `missingNewlineAtEof.obj/.mtl`

src/MagnumPlugins/ObjImporter/Test/Test.cpp
Squareys: Good point.

src/MagnumPlugins/ObjImporter/Test/Test.cpp
Squareys: I was actually using CORRADE_SKIP alot in the AssimpImporter, to also provide the reason why it fails, no idea what I meant to do here... :thinking:

src/MagnumPlugins/ObjImporter/Test/Test.cpp
Squareys: Uh oh, this was still some hasty code from back when I was working on this for University, gah 😅

Squareys:
Thanks for the initial review!

> I hope I didn't put you off with those 36 comments, though -- sorry about that.

Never! I am always happy to learn! :D 

A couple of things were pretty obvious, though, these things would have gotten resolved in an autonomous code review also 😉 I don't think commenting on code stlye/formatting makes too much sense yet, since I didn't do that final cleanup pass myself yet.

I will have time to apply the requested changes over the weekend :)

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
mosra: Yes, that's basically what I was thinking.

The move will bring some confusion/burden to users and packagers, though...

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
mosra: Sorry :)

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
mosra: Ah. Right, confused it with this one.

So this means there can't be whitespace before the first keyword in the file but can be for all the other? Is that desired behavior? To make the code clearer, I would touch just a single line in every loop iteration, not going over -- so ignoring prefix whitespace at the beginning and ignoring suffix whitespace at the end.

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
mosra: Maybe Wikipedia has it wrong, but I saw some prefix whitespace in the code samples there: https://en.wikipedia.org/wiki/Wavefront_.obj_file#File_format

I personally don't see why there should be such a restriction of no whitespace at line beginning.

src/MagnumPlugins/ObjImporter/Test/multiMaterial.mtl
mosra: I think this problem would go away completely with the single-pass parser design I outlined above -- if there would be no newline at the end, the parser loop would just not ignore anything at the end and then not go into another loop iteration.

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
Squareys: I never said it was a restriction, the parser will parse `······\t····p 1` just fine with my current implementation?

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
mosra: Yes, but why a comment in the middle of a `f` lines would split them into two sections?

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
mosra: > the parser will parse `······\t····p·1` just fine with my current implementation?

My point is that this wouldn't work if this would be the very first line (as far as I can see), so I'm suggesting putting the `skipWhitespaces()` as the very first thing in the loop, instead of very last ;)

Squareys: Hmm, alright, yeah, I seem to have removed the first skipWhitespace call somewhere, but what you say makes sense

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
Squareys: Then I can consistently parse only what I need later and expect it to be exactly what I need and nothing else.

src/MagnumPlugins/ObjImporter/ObjImporter.h
```diff
@@ -86,14 +88,28 @@ class MAGNUM_OBJIMPORTER_EXPORT ObjImporter: public AbstractImporter {
...
+        MAGNUM_OBJIMPORTER_LOCAL UnsignedInt doImage2DCount() const;
+        MAGNUM_OBJIMPORTER_LOCAL std::optional<ImageData2D> doImage2D(UnsignedInt id);
+
+        MAGNUM_OBJIMPORTER_LOCAL void parse();
+        MAGNUM_OBJIMPORTER_LOCAL void parseMaterialLibrary(Containers::ArrayView<char> libname);
 
-        std::unique_ptr<File> _file;
+        Containers::Array<char> _in;
+        std::unique_ptr<struct ImporterState> _state;
+        std::string _fileRoot;
```
Squareys: ✔️ done

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
Squareys: ✔️ done

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
Squareys: ✔️ done in over 60 instances :P

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
```diff
@@ -28,81 +28,338 @@
...
+struct ObjMeshData {
+    ObjGroup& group;
+    int materialId; /* from 'usemtl' keyword */
+
+    ObjMesh* points = nullptr;
+    ObjMesh* lines = nullptr;
+    ObjMesh* faces = nullptr;
+
+    ObjMeshData()=delete;
+    ObjMeshData(ObjGroup& g, int matId): group(g), materialId(matId) {}
```
Squareys: ✔️ done

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
```diff
@@ -28,81 +28,338 @@
...
+    std::vector<Vector3> positions;
+    std::vector<Vector2> texCoords;
+    std::vector<Vector3> normals;
 };
 
 namespace {
 
-void ignoreLine(std::istream& in) {
-    in.ignore(std::numeric_limits<std::streamsize>::max(), '\n');
+int strToInt(const ArrayView<char> str) {
```
Squareys: ✔️ done (*many* consts added, basically everywhere...)

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
Squareys: ✔️ done (Cleanly formulated all the helper documentations and checked for Captial letter at comment starts everywhere)

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
Squareys: ✔️ done (Documented)

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
Squareys: ✔️ done (wow, this looks sooo much better now 😍 )

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
Squareys: ✔️ done

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
Squareys: ✔️ done

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
Squareys: ✔️ done (`Utility::Directory::path(filename)`)

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
Squareys: ✔️ done

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
Squareys: ✔️ done

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
```diff
@@ -115,387 +372,575 @@ ObjImporter::~ObjImporter() = default;
...
+            ArrayView<char> word;
+            pos = skipWhitespaces(pos);
+            std::tie(word, pos) = nextWord(pos);
+
+            parseMaterialLibrary(word);
+
+        /* Set current material and add a new mesh for it */
+        } else if(keyword == "usemtl") {
+        ArrayView<char> word;
+        pos = skipWhitespaces(pos);
```
Squareys: ✔️ done

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
Squareys: ✔️ done

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
Squareys: ✔️ done

src/MagnumPlugins/ObjImporter/Test/multiMaterial.mtl
Squareys: ✔️ done

src/MagnumPlugins/ObjImporter/Test/Test.cpp
Squareys: Well, for the empty file, we now don't load anything at all...
I could put a mesh there, and then check the `mesh3DForName("")`. But why should it be returning -1 anyway?

src/MagnumPlugins/ObjImporter/Test/Test.cpp
Squareys: ✔️ done (restored)

src/MagnumPlugins/ObjImporter/Test/Test.cpp
Squareys: ✔️ done

src/MagnumPlugins/ObjImporter/Test/Test.cpp
Squareys: No clue.
✔️ done (restored)

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
mosra: It's not used on *that* many places ;)

My reasoning: if a namespace is annoyingly long to type, it indicates a problem with the API naming -- it should be shortened. Using `using namespace` would somewhat hide that problem, but not solve it, and also makes the code a bit more opaque for people that don't know what belongs to the `used` namespace. I would say that in this case the namespace is still bearable ;)

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
mosra: Okay -- if this behaves in a way that adding a comment line *does not* affect how the parsed data look like, I'm fine with that :)

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
mosra: Could it be documented that there is some non-`nullptr` value returned, just to remove any potential confusion?

src/MagnumPlugins/ObjImporter/Test/Test.cpp
mosra: :D

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
```diff
@@ -28,81 +28,344 @@
...
-struct ObjImporter::File {
-    std::unordered_map<std::string, UnsignedInt> meshesForName;
-    std::vector<std::string> meshNames;
-    std::vector<std::tuple<std::streampos, std::streampos, UnsignedInt, UnsignedInt, UnsignedInt>> meshes;
-    std::unique_ptr<std::istream> in;
+
+/* Not using PhongMaterialData, since we may want to set color and texture to
+ * later decide which flags we set. We do not know whether we have a texture
+ * or color beforehand. */
+struct ObjMaterial {
```
mosra: Oh, forgot to say: could the internal `Obj*` structs be wrapped in an anonymous namespace as well to prevent symbol leakages? Thank you.

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
```diff
@@ -28,81 +28,344 @@
...
+            name += "%" + out.str();
+        }
+
+        return name;
+    }
+};
+
+
+/* The state of the imported generated by openData() */
+struct ImporterState {
```
mosra: Thank you for putting it all in one -- could it be `ObjImporter::State` or something so it doesn't appear in the global `Trade` namespace? There is a leftover `struct File` forward-declaration in the header also, so you could just reuse that line.

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
mosra: Hmm. I saw the `object3DName()` and `mesh3DName()` internally use the same mapping structures, so I thought the name retrieval wouldn't be *that* different?

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
Squareys: Previously the state `unique_ptr` was reset in `doOpen*()`, but those two added members `in` and `fileRoot` shouldn't be, because they are set before the state is reset. Hence two options:
Either `unique_ptr<File>` with a new struct `File` in the Importer state struct (which can be reset independently from the importer state)  or revert back to having the two members outside of this struct.

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
Squareys: ✔️ done

src/MagnumPlugins/ObjImporter/ObjImporter.cpp
mosra: Took me a while to realize what's going on, sorry for the late reply.

> but those two added members in and fileRoot shouldn't be, because they are set before the state is reset

I think they don't need to be. One option:

```cpp
void ObjImporter::doOpenFile(const std::string& filename) {
    AbstractImporter::doOpenFile(filename);
    _state->fileRoot = Utility::Directory::path(filename);
}

void ObjImporter::doOpenData(Containers::ArrayView<const char> data) {
    _state.reset(new ImporterState);
    _state->in = Containers::Array<char>{data.size()};
    std::copy(data.begin(), data.end(), _state->in.begin());
    parse();
}
```

The above assumes that `parse()` doesn't need to access the `_fileRoot` parameter (which it shouldn't, if you defer the material parsing to later, as I suggested). Another option could be this, where the `_state` is only populated if it's not already there:

```cpp
void ObjImporter::doOpenFile(const std::string& filename) {
    _state.reset(new ImporterState);
    _state->fileRoot = Utility::Directory::path(filename);
    AbstractImporter::doOpenFile(filename);
}

void ObjImporter::doOpenData(Containers::ArrayView<const char> data) {
    if(!_state) _state.reset(new ImporterState);
    _state->in = Containers::Array<char>{data.size()};
    std::copy(data.begin(), data.end(), _state->in.begin());
    parse();
}
```

The `AbstractImporter::open*()` function implementation is [unconditionally calling](https://github.com/mosra/magnum/blob/2384d5cccdb020342ad9ad6d40d700c7586acbc1/src/Magnum/Trade/AbstractImporter.cpp#L56) `close()` before calling the `doOpen()` implementations, so you can be always sure that if `_state` is already populated, it's yours and not some stale one from a previous file.

sariug:
Hello,

I am trying to open an obj file. Tried it with 3 different version of it. For same obj file:
- I tried opening the original
- Opened in MeshLab and saved again **with** Vertex+color information
- Opened in MeshLab and saved again **without** Vertex+color information

None of the worked nicely with Obj viewer plugin.
Errors I have gotten are, respectively:
-Trade::ObjImporter::mesh3D(): error while converting numeric data
-Trade::ObjImporter::mesh3D(): invalid float array size
-Opening file deneme2.obj; Importing mesh 0 ; Cannot load the mesh, skipping

Do you have any other suggestions to handle these issue as well as is there any expected improvement in this plugin ? or any other suggestions to deal with it ? 

The obj file looks like : 
image


Regards,

PR #209, opened 2017-06-10
[appveyor] Added vs2017 to the build images

sigman78:


mosra:
Thank you very much for this! 

Cherry-picked with minor changes in e287c3c410a9dfa35d2e054c63d4504b4d5f6ffc, 24183ce3ba9e224e8c8c02e772c99bfd4b8ab678 and 04c7ffdf1b3c7d148526a2791201949468d86b54.

Compared to MSVC 2015 a lot less workarounds is needed and the builds seem to be a bit faster, in particular all `constexpr` functionality in Math classes works properly the same way as in GCC/Clang.

Issue #208, opened 2017-06-10
BoolVector bit setting bug


Crisspl:
I think you've got a little bug [here](https://github.com/mosra/magnum/blob/master/src/Magnum/Math/BoolVector.h#L125). If `value` is `false`, then i-th bit won't be set to 0. I was wondering if you just meant this method to set to 1 a bit (probably *set* can be understood like this in context of bits), but then presence of `value` argument doesn't make sense to me, so I think it's a bug. I was looking for a proper solution and [there](https://stackoverflow.com/questions/47981/how-do-you-set-clear-and-toggle-a-single-bit) I found it.

mosra:
Fixed with #210. Thanks a lot!

PR #210, opened 2017-06-10
Fixed bit setting bug in BoolVector::set

Crisspl:
There's also an issue which I created before I realised it'd be probably better to create a PR: https://github.com/mosra/magnum/issues/208
Anyway - if `value` arg is `false` and desired bit is set to 1, then the bit will still remain 1 (won't change value to 0). The solution I found [here](https://stackoverflow.com/questions/47981/how-do-you-set-clear-and-toggle-a-single-bit), ~~but I didn't use *best answer* - [the one](https://stackoverflow.com/questions/47981/how-do-you-set-clear-and-toggle-a-single-bit#comment-46654671) from the comments seemed more reliable to me~~.

### DIFF
--- a/src/Magnum/Math/BoolVector.h
+++ b/src/Magnum/Math/BoolVector.h
@@ -122,7 +122,7 @@ template<std::size_t size> class BoolVector {
 
         /** @brief Set bit at given position */
         BoolVector<size>& set(std::size_t i, bool value) {
-            _data[i/8] ^= (-Byte(value) ^ _data[i/8]) & (0x01 << i);
+            _data[i/8] ^= (-Byte(value) ^ _data[i/8]) & (0x01 << i%8);
             return *this;
         }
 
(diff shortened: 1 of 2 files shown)


Crisspl:
Commited once more. I didn't test my previous solution and it was wrong.

mosra:
Thank you! Oops, yes, what I did there is definitely wrong.

But I think I would exchange the cleverness for a bit more readability:

```cpp
value ? _data[i/8] |=  (1 << i%8) : 
        _data[i/8] &= ~(1 << i%8);
```

Thanks for adding the the test case as well. Will squash the commits and merge it with the above change, if that's okay for you :)

Crisspl:
Ok, whatever you think is the best :)

mosra:
Merged in 4f746ce7299e39551f639e23c0e0b7828f06a08c and f1f34df6a0a9b7a74f5d9773d7de121eb89def5f. Thank you!

PR #216, opened 2017-08-18
Audio::Error/Audio::Context::Error enums and some emscripten specifics 

Squareys:
Hi @mosra !

As promised, I started porting the audio example. I made a minor enhancement which now prints a decent error when `alcCreateContext` fails because Emscripten does not support specifying attributes.

Cheers, Jonathan.

### DIFF
--- a/src/Magnum/Audio/Context.cpp
+++ b/src/Magnum/Audio/Context.cpp
@@ -228,7 +228,16 @@ bool Context::tryCreateContext(const Configuration& config) {
         attributes[last++] = config.refreshRate();
     }
 
+#ifndef CORRADE_TARGET_EMSCRIPTEN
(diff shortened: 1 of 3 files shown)


mosra:
Merged manually as 30de8726081fbf9edb3e7501397c298cc4a845c6 and 20110358b0c2ba06214d5ddfebabc7f74ac36d53. The Error enum was already a part of `Audio::Renderer` and the other had no matching public API, so I'm just printing the stringified `ALC_*` enums instead of adding another indirection. I also made the Emscripten patch into just a warning, saying that the parameters are ignored at the moment (instead of failing hard), in order to eliminate unnecessary patching of user code.

Thanks for the contribution! :)

PR #217, opened 2017-09-01
Replaced Implementation::Sequence with std::index_sequence etc.

Crisspl:
Have seen a `@todo` tag about this somewhere in code and just found some time to take care of it :)

### DIFF
--- a/src/Magnum/Array.h
+++ b/src/Magnum/Array.h
@@ -32,7 +32,7 @@
 #include <type_traits>
 
 #include "Magnum/Magnum.h"
-#include "Magnum/Math/BoolVector.h" /* for Math::Implementation::Sequence */
+#include "Magnum/Math/BoolVector.h" /* for Math::Implementation::repeat() */
 
 namespace Magnum {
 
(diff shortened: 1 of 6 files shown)


mosra:
Thanks for the contribution! This however depends on having C++14 support and unfortunately some of the compilers I need to support don't know it yet. I think at least Travis CI will complain.

I'll keep this here until support for those compilers is dropped.

Hmm, looking closer, the trick with removing the need for "reverse sequence" is neat, I'll integrate that one (and credit you) :)

Crisspl:
Just saw build logs :/ I assumed, that Magnum supports only c++14-aware compilers (probably because of repository description).

mosra:
Integrated the `GenerateReverseSequence` part in 98f920527e2cd4e8d6f0c9ddb7ea9ebbf1d7843b.

Crisspl:
Nice, thanks for assigning me :)

mosra:
The `Sequence` helpers were moved to Corrade in mosra/corrade@bfee14a031c780de8ebff350155e7912910cb85a and significantly optimized in mosra/corrade@0b828147d692a472846dad16f730bc0d844886b7 so there's not much advantage left in switching to `std::index_sequence`. And Magnum isn't going to move to C++14 anytime soon, which makes this PR not relevant anymore.

PR #218, opened 2017-09-26
SDL2 Support when building for Emscripten

Squareys:
Hi @mosra !

As mentioned, I am in the midst of moving the Sdl2Application to SDL2 for Emscripten (which removes alot of preprocessor branches).

Cheers, Jonathan

### DIFF
--- a/modules/FindMagnum.cmake
+++ b/modules/FindMagnum.cmake
@@ -534,6 +534,10 @@ foreach(_component ${Magnum_FIND_COMPONENTS})
                 find_package(SDL2)
                 set_property(TARGET Magnum::${_component} APPEND PROPERTY
                     INTERFACE_LINK_LIBRARIES SDL2::SDL2)
+                if(CORRADE_TARGET_EMSCRIPTEN)
+                    # TODO: give me INTERFACE_LINK_OPTIONS or something, please
+                    set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -s USE_SDL=2")
+                endif()
 
             # (Windowless) GLX application dependencies
             elseif(_component STREQUAL GlxApplication OR _component STREQUAL WindowlessGlxApplication)
--- a/src/Magnum/Platform/Sdl2Application.cpp
+++ b/src/Magnum/Platform/Sdl2Application.cpp
@@ -100,7 +100,6 @@ bool Sdl2Application::tryCreateContext(const Configuration& configuration) {
     SDL_GL_SetAttribute(SDL_GL_FRAMEBUFFER_SRGB_CAPABLE, configuration.isSRGBCapable());
     #endif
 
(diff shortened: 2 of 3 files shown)


modules/FindMagnum.cmake
```diff
@@ -534,6 +534,10 @@ foreach(_component ${Magnum_FIND_COMPONENTS})
                 find_package(SDL2)
                 set_property(TARGET Magnum::${_component} APPEND PROPERTY
                     INTERFACE_LINK_LIBRARIES SDL2::SDL2)
+                if(CORRADE_TARGET_EMSCRIPTEN)
+                    # TODO: give me INTERFACE_LINK_OPTIONS or something, please
+                    set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -s USE_SDL=2")
```
mosra: I need to verify that this doesn't bloat Emscripten apps that are not using SDL2. Apart from that, I have no other comment. Thanks for the contribution!

src/Magnum/Platform/Sdl2Application.cpp
```diff
@@ -100,8 +100,6 @@ bool Sdl2Application::tryCreateContext(const Configuration& configuration) {
     SDL_GL_SetAttribute(SDL_GL_FRAMEBUFFER_SRGB_CAPABLE, configuration.isSRGBCapable());
     #endif
 
-    /** @todo Remove when Emscripten has proper SDL2 support */
-    #ifndef CORRADE_TARGET_EMSCRIPTEN
     /* Set context version, if user-specified */
     if(configuration.version() != Version::None) {
```
mosra: This line doesn't compile for me:

```
../src/Magnum/Platform/Sdl2Application.cpp:104:22: error: no member named 'version' in 'Magnum::Platform::Sdl2Application::Configuration'
    if(configuration.version() != Version::None) {
```

I force-pushed a rebased version on your branch, waiting for the CI to say what's up. Can you look into it after?

Also, while at it, there's also `MouseEvent::clickCount()` that's disabled on Emscripten currently (shouldn't be anymore if using SDL2) and the `isTextInputActive()` function has a probably-no-longer-necessary Emscripten workaround in it (you have to verify that the linker doesn't fail).

Sorry for the premature "no other comment" comment, then. These are things that can be seen only after I pull locally, not from the diff :D

Squareys: Alright, thanks for the review! I will find some time today to finish that up 👍

Squareys:
Thanks! 👍

mosra:
Just for the record: reverted back to the previous state in 6ae21f23b5e916a34a3b18e67d24e67191d2d2d3, because the size increase even in WASM build was over half a megabyte (i.e., size of the triangle example went from 600 kB to 1.2 MB), which I don't find acceptable.

Will revisit this later, if needed.

PR #220, opened 2017-11-02
Solid circle primitive is missing a triangle

Guillaume227:
Adding one extra vertex in solid circle primitive in order to fill missing triangle.

This is what I see with the following code:
```
		auto const circle = Primitives::Circle::solid(6);
		
		_vertexBuffer.setData(circle.positions(0), BufferUsage::StaticDraw);
		_mesh.setPrimitive(circle.primitive())
			.setCount(circle.positions(0).size())
			.addVertexBuffer(_vertexBuffer, 0, Shaders::Flat2D::Position{})

```

image

### DIFF
--- a/src/Magnum/Primitives/Circle.cpp
+++ b/src/Magnum/Primitives/Circle.cpp
@@ -44,7 +44,7 @@ Trade::MeshData2D Circle::solid(UnsignedInt segments) {
 
     /* Points on circle */
     const Rad angleIncrement(Constants::tau()/segments);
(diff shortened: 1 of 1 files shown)


mosra:
Hi,

sorry, somehow missed the GitHub notifications so I'm answering late. Thanks a lot for the fix, I'm stupid :) I apparently misunderstood how triangle fan primitive works (I thought it closes itself) and never bothered to test properly.

Applied your patch as 059f1396b039be4564369382cb0dada4b3b11282 (updating the unit test as well, so the builds pass).

PR #225, opened 2018-01-17
Add openState in abstract importer

costashatz:
As discussed in gitter, this PR adds `openState` capability to the `AbstractImporter`. This can be used to directly set the state of the importer (i.e., to load from memory an already loaded file etc.).

I didn't try to add any doc in the `openState` function, as @mosra most probably knows better what to write to be consistent..


src/Magnum/Trade/AbstractImporter.h
```diff
@@ -496,6 +496,8 @@ class MAGNUM_EXPORT AbstractImporter: public PluginManager::AbstractManagingPlug
          */
         const void* importerState() const;
 
+        void openState(const void* state);
```
Squareys: There needs to be a piece of documentation here to explain what `openState` does 😉

mosra: Similarly to the `*.cpp` file I would put this function right next to the `openData()`/`openFile()` declaration. Then you could just copypaste the docs from `openData()` and change relevant portions (mention `Feature::OpenState` etc.). That's it. There's then a few cases where `openData()`/`doOpanData()` is mentioned above in the main class docs, would be great if you'd mention the new functions there as well, but I can do that later, too.

src/Magnum/Trade/AbstractImporter.h
```diff
@@ -821,6 +823,9 @@ class MAGNUM_EXPORT AbstractImporter: public PluginManager::AbstractManagingPlug
 
         /** @brief Implementation for @ref importerState() */
         virtual const void* doImporterState() const;
+
+        /** @brief Implementation for @ref openState() */
+        virtual void doOpenState(const void* state);
```
mosra: Third and last comment: I would *also* put this next to the `doOpenData()`/`doOpenFile()` functions.

costashatz:
> Okay, I think this is perfect now :)

Ahh..I just saw. Maybe we could document the `filePath` argument, or not?

mosra:
I'll do that together with the other docs updates :)

mosra:
Cherry-picked as 4d7f955d55f0b0e3d087106a840c2ecf2547c90e. Thank you!

PR #226, opened 2018-01-21
Culling math

Squareys:
Hi @mosra !

As discussed per gitter, I started working on some culling math for a project at university. This pullrequest will add cleanly tested implementations for:

 - [x] `Math::Geometry::Intersection::sphereFrustum`
 - [x] Improve `Math::Geometry::Intersection::boxFrustum`
     - [x] Deprecate and rename to `Math::Geometry::Intersection::rangeFrustum`
     - [x] Split and add to`Math::Geometry::Intersection::aabbFrustum`
 - [x] `Math::Geometry::Intersection::pointCone`
 - [x] `Math::Geometry::Intersection::pointDoubleCone`
 - [x] `Math::Geometry::Intersection::sphereCone`
 - [x] `Math::Geometry::Intersection::rangeCone`
 - ~~(Maybe `Math::Geometry::Intersection::triangleCone`, even though it's not really useful for culling)~~

Cheers, Jonathan

___

~~Plus batch intersection methods for:~~ (not part of this PR, progress backed up in [separate branch](https://github.com/Squareys/magnum/tree/culling-math-batch) )
 - [x] ~~Range - Frustum~~
 - ~~Range - Cone~~
 - [x] ~~Sphere - Frustum~~ Batching does not seem to yield improvements with array of structures
 - ~~Sphere - Cone~~

~~Should I put them into a separate file? And/or separate `Batch::` namespace?~~ Discussed: `IntersectionBatch.h/.cpp` and same namespace.

### DIFF
--- a/src/Magnum/Math/Geometry/Intersection.h
+++ b/src/Magnum/Math/Geometry/Intersection.h
@@ -5,7 +5,7 @@
 
     Copyright © 2010, 2011, 2012, 2013, 2014, 2015, 2016, 2017, 2018
               Vladimír Vondruš <mosra@centrum.cz>
-    Copyright © 2016 Jonathan Hale <squareys@googlemail.com>
+    Copyright © 2016, 2018 Jonathan Hale <squareys@googlemail.com>
 
     Permission is hereby granted, free of charge, to any person obtaining a
     copy of this software and associated documentation files (the "Software"),
@@ -33,7 +33,9 @@
 #include "Magnum/Math/Frustum.h"
 #include "Magnum/Math/Geometry/Distance.h"
 #include "Magnum/Math/Range.h"
+#include "Magnum/Math/Vector2.h"
 #include "Magnum/Math/Vector3.h"
+#include "Magnum/Math/Matrix4.h"
 
 namespace Magnum { namespace Math { namespace Geometry { namespace Intersection {
 
@@ -143,7 +145,7 @@ template<class T> inline T planeLine(const Vector3<T>& planePosition, const Vect
 }
 
 /**
-@brief Intersection of a point and a camera frustum
+@brief Intersection of a point and a frustum
 @param point    Point
 @param frustum  Frustum planes with normals pointing outwards
 
@@ -155,19 +157,259 @@ points distance from the plane is negative) using @ref Distance::pointPlaneScale
 template<class T> bool pointFrustum(const Vector3<T>& point, const Frustum<T>& frustum);
 
 /**
-@brief Intersection of an axis-aligned box and a camera frustum
+@brief Intersection of a range and a frustum
+@param range    Range
+@param frustum  Frustum planes with normals pointing outwards
+
+Returns @cpp true @ce if the box intersects with the frustum.
+
+Uses the "p/n-vertex" approach: First converts the @ref Range3D into a representation
+using center and extent which allows using the following condition for whether the
+plane is intersecting the box: @f[
+     \begin{array}{rcl}
+         d & = & \boldsymbol c \cdot \boldsymbol n \\
+         r & = & \boldsymbol c \cdot \text{abs}(\boldsymbol n) \\
+         d + r & < & -w
+     \end{array}
+@f]
+
+for plane normal @f$ \boldsymbol n @f$ and determinant @f$ w @f$.
+
+@see @ref aabbFrustum()
+*/
+template<class T> bool rangeFrustum(const Range3D<T>& range, const Frustum<T>& frustum);
+
+#ifdef MAGNUM_BUILD_DEPRECATED
+/**
+@brief @copybrief rangeFrustum()
+@deprecated Use @ref rangeFrustum() instead.
+*/
+template<class T> CORRADE_DEPRECATED("use rangeFrustum() instead") bool boxFrustum(const Range3D<T>& box, const Frustum<T>& frustum) {
+    return rangeFrustum(box, frustum);
+}
+#endif
+
+/**
+@brief Intersection of an axis-aligned box and a frustum
 @param box      Axis-aligned box
 @param frustum  Frustum planes with normals pointing outwards
 
-Returns @cpp true @ce if the box intersects with the camera frustum.
+Returns @cpp true @ce if the box intersects with the frustum.
 
-Counts for each plane of the frustum how many points of the box lie in front of
-the plane (outside of the frustum). If none, the box must lie entirely outside
-of the frustum and there is no intersection. Else, the box is considered as
-intersecting, even if it is merely corners of the box overlapping with corners
-of the frustum, since checking the corners is less efficient.
+Uses the same method as @ref rangeFrustum() "rangeFrustum()", but does not need to
+convert to center/extents representation.
 */
-template<class T> bool boxFrustum(const Range3D<T>& box, const Frustum<T>& frustum);
+template<class T> bool aabbFrustum(const Vector3<T>& center, const Vector3<T>& extents, const Frustum<T>& frustum);
+
+/**
+@brief Intersection of a sphere and a frustum
+@param center   Sphere center
+@param radius   Sphere radius
+@param frustum  Frustum planes with normals pointing outwards
+
+Returns @cpp true @ce if the sphere intersects the frustum.
+
+Checks for each plane of the frustum whether the sphere is behind the plane (the
+points distance larger than the sphere's radius) using @ref Distance::pointPlaneScaled().
+*/
+template<class T> bool sphereFrustum(const Vector3<T>& center, T radius, const Frustum<T>& frustum);
+
+/**
+@brief Intersection of a point and a cone
+@param p        The point
+@param origin   Cone origin
+@param normal   Cone normal
+@param angle    Apex angle of the cone
+
+Returns @cpp true @ce if the point is inside the cone.
+
+Precomputes a portion of the intersection equation from @p angle and calls
+@ref pointCone(const Vector3&, const Vector3&, const Vector3&, T)
+*/
+template<class T> bool pointCone(const Vector3<T>& p, const Vector3<T>& origin, const Vector3<T>& normal, Rad<T> angle);
+
+/**
+@brief Intersection of a point and a cone using precomputed values
+@param p            The point
+@param coneOrigin   Cone origin
+@param coneNormal   Cone normal
+@param tanAngleSqPlusOne Precomputed portion of the cone intersection equation:
+                         @cpp T(1) + Math::pow(Math::tan(angle/T(2)), T(2)) @ce
+
+Returns @cpp true @ce if the point is inside the cone.
+*/
+template<class T> bool pointCone(const Vector3<T>& p, const Vector3<T>& coneOrigin, const Vector3<T>& coneNormal, T tanAngleSqPlusOne);
+
+/**
+@brief Intersection of a point and a double cone
+@param p            The point
+@param coneOrigin   Cone origin
+@param coneNormal   Cone normal
+@param coneAngle    Apex angle of the cone
+
+Returns @cpp true @ce if the point is inside the double cone.
+
+Precomputes a portion of the intersection equation from @p angle and calls
+@ref pointDoubleCone(const Vector3&, const Vector3&, const Vector3&, T)
+*/
+template<class T> bool pointDoubleCone(const Vector3<T>& p, const Vector3<T>& coneOrigin, const Vector3<T>& coneNormal, Rad<T> coneAngle);
+
+/**
+@brief Intersection of a point and a double cone using precomputed values
+@param p            The point
+@param coneOrigin   Cone origin
+@param coneNormal   Cone normal
+@param tanAngleSqPlusOne Precomputed portion of the cone intersection equation:
+                         @cpp T(1) + Math::pow(Math::tan(angle/T(2)), T(2)) @ce
+
+Returns @cpp true @ce if the point is inside the double cone.
+
+Uses the result of precomputing @f$ x = \tan^2{\theta} + 1 @f$.
+*/
+template<class T> bool pointDoubleCone(const Vector3<T>& p, const Vector3<T>& coneOrigin, const Vector3<T>& coneNormal, T tanAngleSqPlusOne);
+
+/**
+@brief Intersection of a sphere and a cone view
+@param sphereCenter   Center of the sphere
+@param radius         Radius of the sphere
+@param coneView       View matrix with translation and rotation of the cone
+@param coneAngle      Cone opening angle
+
+Returns @cpp true @ce if the sphere intersects the cone.
+
+Transforms the sphere center into cone space (using the cone view matrix) and
+performs sphere-cone intersection with the zero-origin -Z axis-aligned cone.
+
+Precomputes a portion of the intersection equation from @p angle and calls
+@ref sphereConeView(const Vector3<T>&, T, const Matrix4<T>&, T, T, T)
+*/
+template<class T> bool sphereConeView(const Vector3<T>& sphereCenter, T radius, const Matrix4<T>& coneView, Rad<T> coneAngle);
+
+/**
+@brief Intersection of a sphere and a cone view
+@param sphereCenter Sphere center
+@param radius   Sphere radius
+@param coneView View matrix with translation and rotation of the cone
+@param sinAngle Precomputed sine of half the cone's opening angle: @cpp Math::sin(angle/T(2)) @ce
+@param cosAngle Precomputed cosine of half the cone's opening angle: @cpp Math::cos(angle/T(2)) @ce
+@param tanAngle Precomputed tangens of half the cone's opening angle: @cpp Math::tan(angle/T(2)) @ce
+
+Returns @cpp true @ce if the sphere intersects the cone.
+*/
+template<class T> bool sphereConeView(const Vector3<T>& sphereCenter, T radius, const Matrix4<T>& coneView, T sinAngle, T cosAngle, T tanAngle);
+
+/**
+@brief Intersection of a sphere and a cone
+@param sphereCenter   Sphere center
+@param radius Sphere  Sphere radius
+@param coneOrigin     Cone origin
+@param coneNormal     Cone normal
+@param coneAngle      Cone opening angle (@f$ 0 < \text{angle} < \pi @f$).
+
+Returns @cpp true @ce if the sphere intersects with the cone.
+
+Offsets the cone plane by @f$ -r\sin{\theta} \cdot \boldsymbol n @f$ (with  @f$ \theta @f$
+the cone's half-angle) which separates two half-spaces:
+In front of the plane, in which the sphere cone intersection test is equivalent
+to testing the sphere's center against a similarly offset cone (which is equivalent
+the cone with surface expanded by @f$ r @f$ in surface normal direction), and
+behind the plane, where the test is equivalent to testing whether the origin of
+the original cone intersects the sphere.
+
+Precomputes a portion of the intersection equation from @p angle and calls
+@ref sphereCone(const Vector3<T>& sphereCenter, T, const Vector3<T>&, const Vector3<T>&, T, T)
+*/
+template<class T> bool sphereCone(const Vector3<T>& sphereCenter, T radius, const Vector3<T>& coneOrigin, const Vector3<T>& coneNormal, Rad<T> coneAngle);
+
+/**
+@brief Intersection of a sphere and a cone using precomputed values
+@param sphereCenter Sphere center
+@param radius       Sphere radius
+@param coneOrigin   Cone origin
+@param coneNormal   Cone normal
+@param sinAngle     Precomputed sine of half the cone's opening angle:
+                    @cpp Math::sin(angle/T(2)) @ce
+@param tanAngleSqPlusOne Precomputed portion of the cone intersection equation:
+                         @cpp Math::pow(Math::tan(angle/T(2)), 2) + 1 @ce
+
+Returns @cpp true @ce if the sphere intersects with the cone.
+*/
+template<class T> bool sphereCone(const Vector3<T>& sphereCenter, T radius, const Vector3<T>& coneOrigin, const Vector3<T>& coneNormal, T sinAngle, T tanAngleSqPlusOne);
+
+/**
+@brief Intersection of an axis aligned bounding box and a cone
+@param center      Center of the AABB
+@param extents     (Half-)extents of the AABB
+@param coneOrigin  Cone origin
+@param coneNormal  Cone normal
+@param coneAngle   Cone opening angle
+
+Returns @cpp true @ce if the axis aligned bounding box intersects the cone.
+
+On each axis finds the intersection points of the cone's axis with infinite planes
+obtained by extending the two faces of the box that are perpendicular to that axis.
+
+The intersection points on the planes perpendicular to axis @f$ a \in {0, 1, 2} @f$
+are given by @f[
+    \boldsymbol i = \boldsymbol n \cdot \frac{(\boldsymbol c_a - \boldsymbol o_a) \pm \boldsymbol e_a}{\boldsymbol n_a}
+@f]
+
+with normal @f$ n @f$, cone origin @f$ o @f$, box center @f$ x @f$ and box extents @f$ e @f$.
+
+The points on the faces that are closest to this intersection point are the closest
+to the cone's axis and are tested for intersection with the cone using
+@ref pointCone(const Vector3<T>&, const Vector3<T>&, const Vector3<T>&, const T) "pointCone()".
+
+As soon as an intersecting point is found, the function returns @cpp true @ce.
+If all points lie outside of the cone, it will return @cpp false @ce.
+
+Precomputes a portion of the intersection equation from @p angle and calls
+@ref aabbCone(const Vector3<T>&, const Vector3<T>&, const Vector3<T>&, const Vector3<T>&, T)
+*/
+template<class T> bool aabbCone(const Vector3<T>& center, const Vector3<T>& extents, const Vector3<T>& origin, const Vector3<T>& normal, Rad<T> angle);
+
+/**
+@brief Intersection of an axis aligned bounding box and a cone using precomputed values
+@param center Center of the AABB
+@param extents (Half-)extents of the AABB
+@param origin Cone origin
+@param normal Cone normal
+@param tanAngleSqPlusOne Precomputed portion of the cone intersection equation:
+                         @cpp T(1) + Math::pow(Math::tan(angle/T(2)), T(2)) @ce
+
+Returns @cpp true @ce if the axis aligned bounding box intersects the cone.
+*/
+template<class T> bool aabbCone(const Vector3<T>& center, const Vector3<T>& extents, const Vector3<T>& origin, const Vector3<T>& normal, T tanAngleSqPlusOne);
+
+/**
+@brief Intersection of a range and a cone
+
+@param range        Range
+@param coneOrigin   Cone origin
+@param coneNormal   Cone normal
+@param coneAngle    Cone opening angle
+
+Returns @cpp true @ce if the range intersects the cone.
+
+Converts the range into center/extents representation and passes it on to
+@ref aabbCone(const Vector3<T>&, const Vector3<T>&, const Vector3<T>&, const Vector3<T>&, T) "aabbCone()".
+*/
+template<class T> bool rangeCone(const Range3D<T>& range, const Vector3<T>& coneOrigin, const Vector3<T>& coneNormal, const Rad<T> angle);
+
+/**
+@brief Intersection of a range and a cone using precomputed values
+@param range        Range
+@param coneOrigin   Cone origin
+@param coneNormal   Cone normal
+@param tanAngleSqPlusOne Precomputed portion of the cone intersection equation:
+                         @cpp T(1) + Math::pow(Math::tan(angle/T(2)), T(2)) @ce
+
+Returns @cpp true @ce if the range intersects the cone.
+
+Converts the range into center/extents representation and passes it on to
+@ref aabbCone(const Vector3<T>&, const Vector3<T>&, const Vector3<T>&, const Vector3<T>&, T) "aabbCone()".
+*/
+template<class T> bool rangeCone(const Range3D<T>& range, const Vector3<T>& coneOrigin, const Vector3<T>& coneNormal, const T tanAngleSqPlusOne);
 
 template<class T> bool pointFrustum(const Vector3<T>& point, const Frustum<T>& frustum) {
     for(const Vector4<T>& plane: frustum.planes()) {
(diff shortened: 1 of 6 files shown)


src/Magnum/Math/Geometry/Intersection.h
```diff
@@ -155,6 +155,19 @@ of the frustum, since checking the corners is less efficient.
...
+@param center   Sphere center
+@param radius   Sphere radius
+@param frustum  Frustum planes with normals pointing outwards
+
+Returns `true` if the sphere intersects the frustum.
+
+Checks for each plane of the frustum whether the sphere is behind the plane (the
+points distance larger than the sphere's radius) using @ref Distance::pointPlaneScaled().
+*/
+template<class T> bool sphereFrustum(const Vector3<T>& center, const T radius, const Frustum<T>& frustum);
```
mosra: no need to have `const T radius` in the declaration -- only in the definition

src/Magnum/Math/Geometry/Intersection.h
```diff
@@ -188,6 +201,89 @@ template<class T> bool boxFrustum(const Range3D<T>& box, const Frustum<T>& frust
...
+        /* The sphere is in front of one of the frustum planes (normals point
+           outwards) */
+        if(Distance::pointPlaneScaled<T>(center, plane) < -radiusSq)
+            return false;
+    }
+
+    return true;
+}
+
+/**
```
mosra: Could the functions below be moved up in front of the long function bodies to be together with the documented function declarations? Reason: people reading docs straight from header files would not need to scroll this far down.

src/Magnum/Math/Geometry/Intersection.h
```diff
@@ -188,6 +201,89 @@ template<class T> bool boxFrustum(const Range3D<T>& box, const Frustum<T>& frust
...
+@brief Intersection of a point and a cone
+@param p        The point
+@param origin   Origin of the cone
+@param normal   Normal of the cone
+@param angle    Apex angle of the cone
+
+Returns `true` if the point is inside the cone.
+
+@see pointCone(Vector3, Vector3, Vector3, T) @todo: That ref is never going to work...
```
mosra: Does `@see @ref pointCone(const Vector3<T>&, const Vector3<T>&, T)` work? Yes, it's a bit shitty, but what can we do...

Similarly elsewhere.

src/Magnum/Math/Geometry/Intersection.h
```diff
@@ -188,6 +201,89 @@ template<class T> bool boxFrustum(const Range3D<T>& box, const Frustum<T>& frust
...
+/**
+@brief Faster version of intersection of a point and a double cone
+@param p        The point
+@param origin   Origin of the cone
+@param normal   Normal of the cone
+@param tanAngleSqPlusOne Precomputed portion of the cone intersection equation: `Math::tan(angle/2)+1`
+
+Returns `true` if the point is inside the double cone.
+
+Uses the result of precomputing @f$x = \tan^2{\theta} + 1@f$.
```
mosra: I think not having space around `@f$` might trigger some Doxygen bugs (or syntax higlighting issues in some editors). Also the space around makes it a bit more readable.

src/Magnum/Math/Geometry/Test/IntersectionTest.cpp
```diff
@@ -140,6 +151,55 @@ void IntersectionTest::boxFrustum() {
...
+    /* Sphere inside */
+    CORRADE_VERIFY(Intersection::sphereFrustum({5.5f, 5.5f, 5.5f}, 1.5f,  frustum));
+    /* Sphere outside */
+    CORRADE_VERIFY(!Intersection::sphereFrustum({0.0f, 0.0f, 100.0f}, 0.5f, frustum));
+}
+
+void IntersectionTest::pointCone() {
+    const Vector3 center{0.1f, 0.2f, 0.3f};
+    const Vector3 normal{0.0f, 1.0f, 0.0f};
+    const Deg<Float> angle{72.0f};
```
mosra: Could you add a typedef to just `Deg` in the typedef section above and use that instead, for consistency?

src/Magnum/Math/Geometry/Intersection.h
```diff
@@ -188,6 +201,89 @@ template<class T> bool boxFrustum(const Range3D<T>& box, const Frustum<T>& frust
...
+
+    return lenA >= 0 && c.dot() <= lenA*lenA*tanAngleSqPlusOne;
+}
+
+/**
+@brief Faster version of intersection of a point and a double cone
+@param p        The point
+@param origin   Origin of the cone
+@param normal   Normal of the cone
+@param tanAngleSqPlusOne Precomputed portion of the cone intersection equation: `Math::tan(angle/2)+1`
```
mosra: It's possible to wrap lines in `@param` blocks, doesn't need to be all on one line.

The m.css Doxygen theme makes it possible to have inline code highlighting, so you can write `@cpp Math::tan(angle/2) + 1 @ce` to have the syntax highlighted (it will also link to Math::tan() once I implement that ;)).

src/Magnum/Math/Geometry/Intersection.h
```diff
@@ -188,6 +201,89 @@ template<class T> bool boxFrustum(const Range3D<T>& box, const Frustum<T>& frust
...
+
+Returns `true` if the point is inside the double cone.
```
mosra: I'm such a crazy person that I usually highlight even this: `@cpp true @ce` (it'll make it appear bold, in case you are wondering).

You are free to say "no" to this, if such extra work offends you :D

mosra:
Oh, regarding the batch things: I think a separate `IntersectionBatch.h`/`IntersectionBatch.cpp` file would be okay, no need to create another namespace (the namespace chain is long enough already :D).

src/Magnum/Math/Geometry/Intersection.h
Squareys: As I said often times before... I use vim 😈

Squareys:
@mosra I would like to keep the benchmarks for `sphereCone` and `boxFrustum` separate to allow executing them separately. Does that make sense? I though about calling them along the lines of `IntersectionBoxFrustumBenchmark`.

src/Magnum/Math/Geometry/Test/IntersectionBenchmark.cpp
```diff
@@ -0,0 +1,156 @@
...
+
+#include "Magnum/Math/Geometry/Intersection.h"
+#include "Magnum/Math/Angle.h"
+#include "Magnum/Math/Matrix4.h"
+
+#include <random>
+#include <tuple>
+#include <utility>
+
+namespace Magnum { namespace Math { namespace Geometry { namespace Benchmark {
```
mosra: Please keep this in the `Test` namespace, there's a lot of implicit behavior depending on that already (one is, for example, making Doxygen ignore all Test namespaces and not requiring documentation inside these, so you don't need to provide documentation blocks like below).

src/Magnum/Math/Geometry/Test/IntersectionBenchmark.cpp
```diff
@@ -0,0 +1,156 @@
...
+    void sphereConeView();
+
+    std::vector<std::pair<Vector3, Float>> _spheres;
+    std::vector<std::tuple<Vector3, Vector3, Rad>> _cones;
+};
+
+IntersectionBenchmark::IntersectionBenchmark() {
+    addBenchmarks({&IntersectionBenchmark::sphereConeNaive,
+                   &IntersectionBenchmark::sphereCone,
+                   &IntersectionBenchmark::sphereConeView}, 1000);
```
mosra: Wouldn't `100` be enough? I.e., don't make this run more than ~5 seconds on the CI and ~1 second on your machine. It's always possible to increase the iteration count using [--repeat-every](http://doc.magnum.graphics/corrade/classCorrade_1_1TestSuite_1_1Tester.html#TestSuite-Tester-command-line), if you want less variance in measured times.

Also, I think decreasing this number but increasing the loop count inside `CORRADE_BENCHMARK(n)` would help with the variance too -- so, for example 25 here and 50 there. You have to experiment, though :)

src/Magnum/Math/Geometry/Test/IntersectionBenchmark.cpp
```diff
@@ -0,0 +1,156 @@
...
+    for(int i = 0; i < 100; ++i) {
+        _spheres.emplace_back(Vector3{pd(g), pd(g), pd(g)}, rd(g));
+    }
+}
+
+void IntersectionBenchmark::sphereConeNaive() {
+    volatile bool b = false;
+    CORRADE_BENCHMARK(10) {
+        for(auto& cone : _cones) {
+            for(auto& sphere : _spheres) {
```
mosra: code golf (making lines shorter) -- maybe like this?

```cpp
CORRADE_BENCHMARK(10) for(auto& cone: _cones) for(auto& sphere: _spheres) {
    b = b ^ ...
}
```

Also, if you can, coding style -- `for(a: b)` instead of `for(a : b)`. There's never a space before `:`. Similarly in other places.

src/Magnum/Math/Geometry/Test/IntersectionBenchmark.cpp
```diff
@@ -0,0 +1,156 @@
...
+
+#include <Corrade/TestSuite/Tester.h>
+
+#include "Magnum/Math/Geometry/Intersection.h"
+#include "Magnum/Math/Angle.h"
+#include "Magnum/Math/Matrix4.h"
+
+#include <random>
+#include <tuple>
+#include <utility>
```
mosra: Coding style: ordering includes: system headers first, then dependency headers (with `<>`), then project-local headers (`""`). Thank you :smiley:

src/Magnum/Math/Geometry/Intersection.h
```diff
@@ -147,14 +149,122 @@ template<class T> bool pointFrustum(const Vector3<T>& point, const Frustum<T>& f
...
+/**
+ * @brief Intersection of a sphere and a cone
+ * @param sphereCenter Sphere center
+ * @param radius Sphere radius
+ * @param origin Origin of the cone
+ * @param normal Cone normal
+ * @param angle Cone opening angle (0 < angle < pi).
+ *
+ * Returns `true` if the sphere intersects with the cone.
+ */
```
mosra: No need to bother with the ` * ` at line starts if the comment block is not indented in any way. Also, `@cpp true @ce` (sorry :D)

src/Magnum/Math/Geometry/Test/IntersectionBoxFrustumBenchmark.cpp
```diff
@@ -0,0 +1,121 @@
...
+    void boxFrustum();
+
+    std::vector<Range3D> _boxes;
+    const Frustum _frustum{
+        {1.0f, 0.0f, 0.0f, 0.0f},
+        {-1.0f, 0.0f, 0.0f, 10.0f},
+        {0.0f, 1.0f, 0.0f, 0.0f},
+        {0.0f, -1.0f, 0.0f, 10.0f},
+        {0.0f, 0.0f, 1.0f, 0.0f},
+        {0.0f, 0.0f, -1.0f, 10.0f}};
```
mosra: Hmm. Couldn't this be written in a shorter, more descriptive way using `Matrix4::orthographicProjection({2.0f, 2.0f}, 0.0f, 10.0f)` or something?

mosra:
> I would like to keep the benchmarks for sphereCone and boxFrustum separate to allow executing them separately

No need to do that, the test suite is capable of handling this on its own. See the [command-line / environment options](http://doc.magnum.graphics/corrade/classCorrade_1_1TestSuite_1_1Tester.html#TestSuite-Tester-command-line), particularly the `--skip-tests` / `--skip-benchmarks` options.

src/Magnum/Math/Geometry/Test/IntersectionBenchmark.cpp
Squareys: That is not a documentation block ;-)

src/Magnum/Math/Geometry/Test/IntersectionBoxFrustumBenchmark.cpp
Squareys: Well, `Frustum::fromMatrix(Matrix4::orthographicProjection({2.0f, 2.0f}, 0.0f, 10.0f))` maybe.

src/Magnum/Math/Geometry/IntersectionBatch.h
```diff
@@ -0,0 +1,100 @@
...
+    LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+    FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+    DEALINGS IN THE SOFTWARE.
+*/
+
+/** @file
+ * @brief Namespace @ref Magnum::Math::Geometry::Intersection
+ */
+
+#include <bitset>
```
mosra: Leftover? :)

src/Magnum/Math/Geometry/IntersectionBatch.h
```diff
@@ -0,0 +1,100 @@
...
+
+#include "Corrade/Containers/ArrayView.h"
```
mosra: Use `<>` for dependency includes, please

src/Magnum/Math/Geometry/Intersection.h
```diff
@@ -167,25 +323,246 @@ template<class T> bool pointFrustum(const Vector3<T>& point, const Frustum<T>& f
...
+
+    const T sinAngle = Math::sin(angle / T(2)); /* precomputable */
+    const T tanAngleSquaredPlusOne = T(1)+Math::pow<T>(Math::tan<T>(angle/T(2)), T(2));
+
+    return sphereCone(sCenter, radius, origin, normal, sinAngle, tanAngleSquaredPlusOne);
+}
+
+template<class T> bool sphereCone(
+        const Vector3<T>& sCenter, const T radius,
+        const Vector3<T>& origin, const Vector3<T>& normal, const T sinAngle, const T tanAngleSquaredPlusOne) {
```
mosra: Not sure about the indentation here. Maybe everything on one line? Or indent by 4 spaces and put the `{` on a new line?

Similarly above.

src/Magnum/Math/Geometry/Intersection.h
```diff
@@ -147,14 +149,168 @@ template<class T> bool pointFrustum(const Vector3<T>& point, const Frustum<T>& f
...
+@ref pointCone(const Vector3<T>&, const Vector3<T>&, const Vector3<T>&, const T)
+
+As soon as an intersecting point is found, the function returns @cpp true @ce.
+If all points lie outside of the cone, it will return @cpp false @ce.
+
+Returns @cpp true @ce if the axis aligned bounding box intersects the cone.
+*/
+template<class T> bool aabbCone(const Vector3<T>& center, const Vector3<T>& extents, const Vector3<T>& origin, const Vector3<T>& normal, const Rad<T> angle);
+
+template<class T> bool aabbCone(const Vector3<T>& center, const Vector3<T>& extents, const Vector3<T>& origin, const Vector3<T>& normal, const T tanAngleSquaredPlusOne);
```
mosra: Since the documentation is a very important part of this, it would be great if you could generate it and verify it says what it should say.

For the overloads, a nice thing could be to say how all the extra parameters are precomputed and also referencing them from the convenience versions.

src/Magnum/Math/Geometry/Intersection.h
```diff
@@ -167,25 +323,246 @@ template<class T> bool pointFrustum(const Vector3<T>& point, const Frustum<T>& f
...
+        if(Distance::pointPlaneScaled<T>(center, plane) < -radiusSq)
+            return false;
+    }
+
+    return true;
+}
+
+
+template<class T> bool pointCone(const Vector3<T>& p, const Vector3<T>& origin, const Vector3<T>& normal, const Deg<T> angle) {
+    const T x = T(1)+Math::pow<T>(Math::tan<T>(angle/T(2)), T(2));
```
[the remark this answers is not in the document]
Squareys: I am unsure why I added those (I usually wouldn't), will check!

src/Magnum/Math/Geometry/Intersection.h
```diff
@@ -167,25 +323,246 @@ template<class T> bool pointFrustum(const Vector3<T>& point, const Frustum<T>& f
...
+
+template<class T> bool pointDoubleCone(const Vector3<T>& p, const Vector3<T>& origin, const Vector3<T>& normal, const Deg<T> angle) {
+    const T x = T(1)+Math::pow<T>(Math::tan<T>(angle/T(2)), T(2));
+
+    return pointDoubleCone(p, origin, normal, x);
+}
+
+template<class T> bool pointDoubleCone(const Vector3<T>& p, const Vector3<T>& origin, const Vector3<T>& normal, const T tanAngleSqPlusOne) {
+    const Vector3<T> c = p - origin;
+    const T lenA = dot(c, normal);
```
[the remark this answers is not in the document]
Squareys: Dot (c,normal) is not the squared length of a vector (I believe you may be confusing it with `c.dot()` for example).

`a` is a quantity that is not clearly defined here (I think by bachelors thesis may clear up what `a` is. It's the length of c projected onto the normal or something like that, if that makes sense.

src/Magnum/Math/Geometry/Intersection.h
Squareys: > Or indent by 4 spaces and put the { on a new line?

I like that option 👍

src/Magnum/Math/Geometry/IntersectionBatch.h
Squareys: Yeah, leftover. Though about using bitset for the result, but it turns out to be compile-time defined size (or have to use dynamic_bitset)

Squareys:
@mosra This is *finally* ready for full review from my side! 👍 

I squashed everything together and rebased onto master. Looking forward to the review.
Cheers, Jonathan.

src/Magnum/Math/Geometry/Intersection.h
```diff
@@ -140,20 +142,246 @@ points distance from the plane is negative) using @ref Distance::pointPlaneScale
...
+The intersection points on the planes perpendicular to axis @f$ a \in {0, 1, 2} @f$
+are given by @f[
+    \boldsymbol i = \boldsymbol n \cdot \frac{(\boldsymbol c_a - \boldsymbol o_a) \pm \boldsymbol e_a}{\boldsymbol n_a}
+@f]
+
+with normal @f$ n @f$, cone origin @f$ o @f$, box center @f$ x @f$ and box extents @f$ e @f$.
+
+The points on the faces that are closest to this intersection point are the closest
+to the cone's axis and are tested for intersection with the cone using
+@ref pointCone(const Vector3<T>&, const Vector3<T>&, const Vector3<T>&, const T) "pointCone".
```
mosra: ... `"pointCone()"`? I.e., so it's clear it refers to a function. 

Similarly in other cases.

src/Magnum/Math/Geometry/Intersection.h
```diff
@@ -140,20 +142,246 @@ points distance from the plane is negative) using @ref Distance::pointPlaneScale
...
+
+As soon as an intersecting point is found, the function returns @cpp true @ce.
+If all points lie outside of the cone, it will return @cpp false @ce.
```
mosra: At first I thought this would give a "false negative" in case the cone is so narrow that it fits inside the AABB without touching the edges. But that's handled by the above mentioned plane intersection, right? So maybe update the sentence to make it clear this case is handled correctly.

src/Magnum/Math/Geometry/Intersection.h
```diff
@@ -166,28 +394,206 @@ template<class T> bool pointFrustum(const Vector3<T>& point, const Frustum<T>& f
...
-        bool cornerHit = 0;
+template<class T> bool rangeFrustum(const Range3D<T>& box, const Frustum<T>& frustum) {
+    /* Convert to center/extent, avoiding division by 2 and instead comparing
+       to 2*-plane.w() later */
+    const Vector3<T> center = box.min() + box.max();
+    const Vector3<T> extent = box.max() - box.min();
 
-        for(UnsignedByte c = 0; c != 8; ++c) {
-            const Vector3<T> corner = Math::lerp(box.min(), box.max(), Math::BoolVector<3>{c});
+    for(const Vector4<T>& plane: frustum.planes()) {
```
mosra: Why is this not just calling into the other `rangeFrustum()` overload? The code is the same.

The `box` param should be `range` to be consistent with the declaration above.

src/Magnum/Math/Geometry/Intersection.h
```diff
@@ -166,28 +394,206 @@ template<class T> bool pointFrustum(const Vector3<T>& point, const Frustum<T>& f
...
+        }
     }
 
-    /** @todo potentially check corners here to avoid false positives */
+    return true;
+}
+
+template<class T> bool sphereFrustum(
+    const Vector3<T>& center, const T radius, const Frustum<T>& frustum)
+{
```
mosra: The declarations aren't too long to be wrapped like this, I think. Similarly in most cases below.

src/Magnum/Math/Geometry/Intersection.h
```diff
@@ -166,28 +394,206 @@ template<class T> bool pointFrustum(const Vector3<T>& point, const Frustum<T>& f
...
+    }
+
+    return false;
+}
+
+template<class T> bool sphereCone(
+    const Vector3<T>& sCenter, const T radius,
+    const Vector3<T>& origin, const Vector3<T>& normal, const Rad<T> angle)
+{
+
```
mosra: Superfluous empty line.

Similarly below.

src/Magnum/Math/Geometry/Test/IntersectionTest.cpp
```diff
@@ -124,20 +144,192 @@ void IntersectionTest::pointFrustum() {
...
+    /* Tangent to that surface vector */
+    auto tangent = Matrix4::rotation(Deg{90.0f}, Vector3::yAxis()).transformVector(surface);
+
+    /* Point on edge */
+    CORRADE_VERIFY(Intersection::pointCone(center, center, normal, angle));
+    /* Point inside */
+    CORRADE_VERIFY(Intersection::pointCone(center + normal, center, normal, angle));
+    /* Point outside */
+    CORRADE_VERIFY(!Intersection::pointCone(Vector3{}, center, normal, angle));
+    //TODO: Why does this fail? CORRADE_VERIFY(!Intersection::pointCone(center + 5.0f*surface + 0.1f*tangent, center, normal, angle));
```
mosra: Is this TODO still relevant?

src/Magnum/Math/Geometry/Intersection.h
```diff
@@ -140,20 +142,246 @@ points distance from the plane is negative) using @ref Distance::pointPlaneScale
...
+@param angle Cone opening angle
+
+Returns @cpp true @ce if the sphere intersects the cone.
+
+Transforms the sphere center into cone space (using the cone view matrix) and
+performs sphere-cone intersection with the zero-origin -Z axis-aligned cone.
+
+@see @ref sphereConeView(const Vector3<T>&, T, const Matrix4<T>&, T, T, T)
+*/
+template<class T> bool sphereConeView(const Vector3<T>& sphereCenter, T radius, const Matrix4<T>& coneView, Rad<T> angle);
```
mosra: Does this work as expected even with the pathological matrices correctly? Such as a skew matrix or non-uniform scaling?

If not, I'd maybe use a some more restricted representation ... or at least mention it in the docs.

mosra:
Oh, looked at the benchmarks -- the difference between `sphereFrustum()` and `sphereCone()` is quite significant, maybe it would be worth to mention `sphereCone()` in the docs of `sphereFrustum()` :)

src/Magnum/Math/Geometry/Intersection.h
```diff
@@ -140,20 +142,246 @@ points distance from the plane is negative) using @ref Distance::pointPlaneScale
...
+         d & = & \boldsymbol c \cdot \boldsymbol n \\
+         r & = & \boldsymbol c \cdot \text{abs}(\boldsymbol n) \\
+         d + r & < & -w
+     \end{array}
+@f]
+
+for plane normal @f$ \boldsymbol n @f$ and determinant @f$ w @f$.
+
+@see @ref aabbFrustum()
+*/
```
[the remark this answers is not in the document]
Squareys: Well, but it doesn't call aabbFrustum.

src/Magnum/Math/Geometry/Intersection.h
```diff
@@ -140,20 +142,246 @@ points distance from the plane is negative) using @ref Distance::pointPlaneScale
...
+@param p        The point
+@param origin   Origin of the cone
+@param normal   Normal of the cone
+@param angle    Apex angle of the cone
+
+Returns @cpp true @ce if the point is inside the cone.
+
+@see @ref pointCone(const Vector3&, const Vector3&, const Vector3&, T)
+*/
+template<class T> bool pointCone(const Vector3<T>& p, const Vector3<T>& origin, const Vector3<T>& normal, Rad<T> angle);
```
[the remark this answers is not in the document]
Squareys: `coneNormal` is the normal of the cone, from the apex towards the open portion:
~~~
  /
 /
<-----> (n)
 \
  \
~~~

src/Magnum/Math/Geometry/Intersection.h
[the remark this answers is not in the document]
mosra: But it *could* -- see [my other comment](https://github.com/mosra/magnum/pull/226/files/3055fd259ff9160861cdc541af527ef303f3d656#r177785355) .. or am I missing something?

src/Magnum/Math/Geometry/Intersection.h
Squareys: Because it's comapring with `-T(2)*plane.w()` instead of `-plane.w()`, to avoid the vector/scalar division when converting range to center & extent representation.

src/Magnum/Math/Geometry/Intersection.h
Squareys: From the Corrade [coding style](http://doc.magnum.graphics/corrade/corrade-coding-style.html) I read:

> vertical and horizontal compression, fitting more code on a screen while still making it possible to have more files open side by side by wrapping on the 80th column

I am pretty sure I only wrapped those which exceeded that quite a bit. You may need to specify "aren't too long" a bit more.

Squareys: This particular line is 105 colums long if not wrapped

Squareys: I did, however, miss quite a few (like the function above) it seems :D

src/Magnum/Math/Geometry/Test/IntersectionTest.cpp
Squareys: Yes, I should have a look at that, (sorry)

src/Magnum/Math/Geometry/Intersection.h
Squareys: It will not, it expect it to be a rigid transform with no or uniform scaling.
Restricted representation as in rotation matrix + translation? Or Dual quaternion? Or Assert on [isRigidTransformation()](http://doc.magnum.graphics/magnum/classMagnum_1_1Math_1_1Matrix3.html#a267cce4346f18620cd93df07bf858234)?

Squareys:
Regarding the benchmark results: I had a closer look at that and would not infer too much on that yet. The frustum and the cone are not very similar at the moment (I will fix that), so this may be the sphereFrustum test running into some worst- or worser-case maybe.

That reminds me, though: For camera frustums and view culling it is common to skip the test against the near plane. Just something to keep in mind and maybe support somehow in the future (e.g. by Frustum only having 5 planes.)

src/Magnum/Math/Geometry/Intersection.h
mosra: (As discussed offline: okay, leave it as is.)

src/Magnum/Math/Geometry/Intersection.h
mosra: Hmm. What would be the best in terms of usability? 

- Allowing Matrix4 with no check inside would lead to surprising results very often, considering scaling is usually part of a transformation. 3x3 rotation matrix suffers from the same problem.
- Assert on `isRigidTransformation()` might be overly expensive
- Requiring dual quaternion might feel too alien to some people

Hmm...

Squareys:
I cleaned up the benchmarks a bit. While the `sphereCone` _seems_ indeed quite a bit faster than the `sphereFrustum` on AppVeyor, which builds **Debug** config, when I build it in **Release** config locally, `sphereFrustum` is actually a little bit faster even.

I finished incorporating the review comments apart from the three that GitHub correctly didn't collapse.

src/Magnum/Math/Geometry/Intersection.h
mosra: Okay, yeah, the coding style is a bit flaky here. In particular, I don't *demand* strict 78 column wrapping like for example the Python PEP-8 tools, I require that only for documentation blocks and bigger blocks of wide code, where the editor wrapping them would make the block significantly harder to read.

Function signatures being a bit longer are okay, because that's just a single line among usually short lines, unless they are so long that an editor wraps them on three and more lines, making editing and reviewing the changes harder.

All function parameter lists that are here wrapped on one or one and half line are still in the okay range to have the full signature in a single line, I think.

src/Magnum/Math/Geometry/Test/IntersectionTest.cpp
```diff
@@ -124,20 +144,192 @@ void IntersectionTest::pointFrustum() {
...
+
+    /* Fully inside */
+    CORRADE_VERIFY(Intersection::rangeFrustum(Range3D{Vector3{1.0f}, Vector3{2.0f}}, frustum));
+    /* Intersects with exactly one plane each */
+    CORRADE_VERIFY(Intersection::rangeFrustum(Range3D{{2.4f, -0.1f, 4.9f}, {2.6f, 0.1f, 5.1f}}, frustum));
+    CORRADE_VERIFY(Intersection::rangeFrustum(Range3D{{2.4f, 0.9f, 4.9f}, {2.6f, 1.1f, 5.1f}}, frustum));
+    CORRADE_VERIFY(Intersection::rangeFrustum(Range3D{{-0.1f, 0.4f, 4.9f}, {0.1f, 0.6f, 5.1f}}, frustum));
+    CORRADE_VERIFY(Intersection::rangeFrustum(Range3D{{4.9f, 0.4f, 4.9f}, {5.1f, 0.6f, 5.1f}}, frustum));
+    CORRADE_VERIFY(Intersection::rangeFrustum(Range3D{{2.4f, 0.4f, -0.1f}, {2.6f, 0.6f, 0.1f}}, frustum));
+    CORRADE_VERIFY(Intersection::rangeFrustum(Range3D{{2.4f, 0.4f, 9.9f}, {2.6f, 0.6f, 10.1f}}, frustum));
```
mosra: What about using Range::fromSize() in the complex cases here? Would make it quite easier to maintain.

Squareys: Yeah, that would condense this a bit, I hope.

src/Magnum/Math/Geometry/Intersection.h
Squareys: As in

~~~
someFunc(
    This is, One line)
{
~~~

and

~~~
SomeOtherFunc(
    That is, Clearly way, Too long,
    Not to, Be broken) //two lines? One and a half?
{
~~~

src/Magnum/Math/Geometry/Test/IntersectionTest.cpp
Squareys: image
Well, not outside after all. Should be, though, wanted it to "just" not touch the surface.

Squareys:
@mosra Done, thanks for all the feedback! This is once again ready for review :)

src/Magnum/Math/Geometry/Intersection.h
Squareys: Considering this is used (if at all) with either camera transformation matrices or light transformation matrices, I would expect the isRigidTransformation to be the most sensible as all other would require a conversion.

mosra:
Attempted to merge this, but:

- Can you rebase on current master to let all the CIs run through? I'm not sure the `CORRADE_EXPECT_FAIL()` would cover so many test cases on Emscripten.
- You can add `Vector3d` and `Radd` typedefs to the test instead of using the verbose `Math::Vector3<Double>` there (and also can omit the `<Double>` for the function call)
- There is some trailing whitespace in the test
- The `sphereConeView()` function is called 640000 times from the benchmark, but not even once from the test. Also, can you add the assert for `isRigidTransformation()` there as discussed above (and a test case with some pathological matrix)? (TODO for me: don't allow benchmarks to contribute to code coverage)
- Some of my comments above are not resolved yet ([1](https://github.com/mosra/magnum/pull/226#discussion_r177779392), [2](https://github.com/mosra/magnum/pull/226#discussion_r177779606))

Thanks!

mosra:
~~One~~ ~~two~~ ~~three~~ ~~four~~ five more things, sorry that I'm not commenting in the code directly, the GitHub UI is awfully slow at this point :/

1. The `sphereConeView()` function now has the `cosAngle` parameter unused (code seems to changed compared to last time?):

  - Is that intentional? Shouldn't the function take one parameter less now?
  - In any case I don't think it's needed to force the user to specify all three of sin, cos and tan (one can be calculated from the other two)

2. The `pointCone()` test has `centerDouble`, `normalDouble` and `angleDouble` variables, none of which is used. Is there something missing? Those variables seem to be used in other tests, though.

3. some functions mention that 0 < cone opening angle < pi, but some not. Should I copy that everywhere for consistency? Is the same for "Apex angle of the cone" in the double cone case? Shouldn't that be also "Cone opening angle"? Or shouldn't be all "cone opening angles" rather "apex angles"? 
And an uppercase Greek letter Theta for this? (That's how Wikipedia has it, at least. You're using lowercase Theta for the half-angle.)

4. Looking at the code, basically all functions taking the apex angle (or whatever the name is) are cutting it in half. What about changing them to already take half angle and save one multiplication there? This will be easy to change now, but *hard* to change later.

5. Both `rangeCone()` functions calculate the same thing and then call `aabbCone()`. Shouldn't the first one (taking the angle) call the second one (taking tanAngleSqPlusOne) and then the second one call `aabbCone()`? Or did I miss something?

Since I'm quite far into the merge (had to rename function parameters for consistency with docs and fix a bunch of errors in the math formulas), I can do the above changes myself (otherwise I would get lots of nasty conflicts later), just tell me how ;)

(Note to myself, in case I forget again: I have an in-progress merge in a private `intersection-merge` branch)

Squareys:
1. Yeah `cosAngle` can be removed
2. A "point on cone" test would still be missing, I checked that locally and it's this test case:
~~~cpp
    // after auto sNormal = Matrix4::rotation(90.0_degf, axis).transformVector(surface);

    /* Same for Double precision */
    auto axisDouble = Math::cross(Vector3d::yAxis(), normalDouble).normalized();
    auto surfaceDouble = Matrix4d::rotation(0.5*angleDouble, axisDouble).transformVector(normalDouble);

    // ...

    /* Point touching cone */
    CORRADE_VERIFY(Intersection::pointCone(centerDouble, centerDouble + surfaceDouble, normalDouble, angleDouble));

~~~

*But* this test fails similar to the others, e.g. the sphere touching cone test, possibly because of precision. Also, the whether emscripten fails or not, I don't know.

3. Yes, it's the same opening angle everywhere and only "acute cones" (hence, 0 < angle < pi). Whether you call it opening angle or apex angle is not too important, I think. On wikipedia it's "apex angle", so maybe you want to use that consistently. For the angle notation just follow the convention you normally use for angles.

4. That's only for the non-precomputed slow function anyway. If you really want to save the multiplication, you can precompute it yourself. This one should be convenient, and I believe that specifying the entire angle instead of the half angle is more intuitive. I'm fine with documenting that it should be the half angle, too, though, if you prefer that.

5. Yes, indeed!

Thank you for all the effort! And sorry that this seems to be producing a lot of work :/

mosra:
1. done :)
2. this seems to be failing similarly to the other. Also it looks like it does not fail on the CI Emscripten but it does on my machine. Ugh.
3. changed to "apex angle" everywhere and using uppercase Theta to be consistent with Wikipedia.
4. good point :+1: does not make sense to optimize anything here

Once I resolve no.2, Is it okay if I push the changes to your branch so you can check I did not do anything terribly wrong to the code?

mosra:
Pushed :) If you can take a quick look, just to have more than one pair of eyes seeing the changes before putting them to master forever, that would be great.

Squareys:
Looks good to me, thanks! 👍

mosra:
Aaaand [you're shipped](http://doc.magnum.graphics/magnum/namespaceMagnum_1_1Math_1_1Intersection.html#a796a25abb39fdaecab9933aff6c70836)! :tada: 

Thanks a lot for this huge piece of work.

Squareys:
Hooray, awesome, thanks! 🎉

PR #228, opened 2018-02-12
[WIP] Python bindings

mosra:
Using [pybind11](https://github.com/pybind/pybind11). 

TODO: 

- [x] move the actual binding code to a new repository (solves compile times, makes it possible to have also corrade, extras, integration... in the same repo) -- mosra/magnum-bindings#1
- [x] enable `-fPIC` also for Application classes so we can add (shared library) bindings for these too
- [x] ~~make it possible to have assertions as exceptions so python doesn't irreversibly blow up (runtime configurable/pluggable function pointer?)~~ no, instead build with `CORRADE_NO_ASSERT` and replicate all assertions on the binding side as actual typed exceptions

### DIFF
--- a/Doxyfile-mcss
+++ b/Doxyfile-mcss
@@ -115,6 +115,7 @@ HTML_EXTRA_STYLESHEET = \
 ##! M_LINKS_NAVBAR2 = \
 ##!  "namespaces" \
 ##!  "annotated" \
-##!  "files"
+##!  "files" \
+##!  "<a href="../../../../magnum-bindings/build/doc/python/index.html">Python API</a>"
 
 # kate: hl Doxyfile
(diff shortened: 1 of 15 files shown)


Squareys:
This is pretty verbose :D (compared to swig or something.) Pretty cool though, as it also alows following python naming conventions.

I think this is super neat, but don't believe it should be a priority 🤔

mosra:
The actual binding code is moved to a new repo now, since it makes more sense (and also doesn't increase compile times of the core repo to twice as much). See mosra/magnum-bindings#1 for details. 

This PR will only contain changes needed for the bindings code to work.

mosra:
Merged the few things here since it doesn't seem I have to radically change how Magnum works in order to make it available in Python. All other adjustments needed for Python bindings (if any) will be added on an as-needed basis.

PR #229, opened 2018-02-19
MeshTools: adding generateSmoothNormals() and some more things

mosra:
Stashing my brain here so I don't forget the particulars.

Started out as a discussion on the DART integration PR (mosra/magnum-integration#29), with the following code snippet, extracted from DART itself, with a goal to make this testable:

```cpp
static Eigen::Vector3d normalFromVertex(const dart::dynamics::SoftBodyNode* bn,
                                        const Eigen::Vector3i& face,
                                        std::size_t v)
{
  const Eigen::Vector3d& v0 = bn->getPointMass(face[v])->getLocalPosition();
  const Eigen::Vector3d& v1 = bn->getPointMass(face[(v+1)%3])->getLocalPosition();
  const Eigen::Vector3d& v2 = bn->getPointMass(face[(v+2)%3])->getLocalPosition();

  const Eigen::Vector3d dv1 = v1-v0;
  const Eigen::Vector3d dv2 = v2-v0;
  const Eigen::Vector3d n = dv1.cross(dv2);

  double weight = n.norm()/(dv1.norm()*dv2.norm());
  weight = std::max( -1.0, std::min( 1.0, weight) );

  return n.normalized() * asin(weight);
}

static void computeNormals(std::vector<Eigen::Vector3d>& normals,
                           const dart::dynamics::SoftBodyNode* bn)
{
  for(std::size_t i=0; i<normals.size(); ++i)
    normals[i] = Eigen::Vector3d::Zero();

  for(std::size_t i=0; i<bn->getNumFaces(); ++i)
  {
    const Eigen::Vector3i& face = bn->getFace(i);
    for(std::size_t j=0; j<3; ++j)
      normals[face[j]] += normalFromVertex(bn, face, j);
  }

  for(std::size_t i=0; i<normals.size(); ++i)
    normals[i].normalize();
}
```

After seeing the snippet I thought calculating smooth normals does not look very complicated, but the angle calculation looked suspicious. So I looked up:

- [a very detailed explanation](http://www.bytehazard.com/articles/vertnorm.html), but with pseudocode which needs to be reworked to match our API, also quite different from above (as far as i can see, it's calculating the angle in three different cases, while the above lacks surface area weighting)
- [an implementation on SO](https://stackoverflow.com/a/45496726) that looks simple and correct but inefficient (array of normals can be reduced to simple `+=`, but I'm not sure why I need to calculate three angles instead of just one)

Besides that, this requires renaming a header in MeshTools.

TODO: 

- [x] mention the header renaming in the changelog
- [x] install the new header, install the old header only if building deprecated
- [x] don't remove duplicates in `generateFlatNormals()`, leave that up to the user (and mention that in docs) ~~(or not?) ... or remove them in some better way (reusing the original index array?)~~
- [x] integrate flat / smooth normal generation into `MeshTools::compile()`
- [x] mention using `MeshTools::generate*Normals()` in StanfordImporter -- mosra/magnum-plugins@382a72b12d26a7b44131d15f9cf738f0db420ec1
- [ ] make `magnum-player` generate normals if needed
- [x] implement `MeshTools::generateSmoothNormals()`, taking index array and vertex position array, from which to calculate neighboring faces (could reuse / publish the algo from `tipsify()`)
- [x] document that hard edges can be created simply by not sharing vertices along the edge
- [x] ~~assimp has some [angle limit for smoothing](http://assimp.sourceforge.net/lib_html/postprocess_8h.html#a64795260b95f5a4b3f3dc1be4f52e410a6afb4fee42eca4482674859196cb8685), implement that as well?~~ postponed
- [x] think about what data to use to properly test for all cases mentioned in the blog above (in particular, generating smooth normals for a cube should result in all of them being the same, just mirrored)


mosra:
This was a *fun* algorithmic exercise -- and I dare to say the implementation is *much* more efficient than what DART is attempting to do in the above snippet. Waiting on the CIs to get green, then it goes to master! :shipit: 

@costashatz I don't really remember the original discussion, is there any place in `DartIntegration` where this could get used?

costashatz:
> @costashatz I don't really remember the original discussion, is there any place in `DartIntegration` where this could get used?

Yes of course! It can be used for the Soft objects: right now, we are creating the faces twice, see [here](https://github.com/mosra/magnum-integration/blob/master/src/Magnum/DartIntegration/ConvertShapeNode.cpp#L307). I am putting it on my TO-DO list.. Thanks for reminding..

mosra:
As [you said there](https://github.com/mosra/magnum-integration/pull/29#discussion_r168971508), that would still need to implement some facet orientation fixer, right? Or did DART get fixed in the meantime? The thing igl uses ([paper](http://jcgt.org/published/0003/04/02/)) is 

> our method is very simple

but by simple they mean it depends on the whole of Embree :sweat_smile: ... so I guess we would need to implement something *actually* simple. Another option I can think of is, instead of duplicating all faces, making the Phong shader optionally double-sided, but that would be even less performant I think.

PR #233, opened 2018-03-01
[WIP] Splitting out GL and Trade functionality to separate libraries

mosra:
This is the first huge potentially breaking thing on the path to Vulkan support. Basically everything from the `Magnum` namespace that had anything to do with OpenGL is now moved into a new `Magnum::GL` library. That goes similarly for include paths and CMake imported targets. But don't worry too much (just a little), this is done with the utmost care to preserve as much backwards compatibility as possible.

I'm doing this one little step at a time to avoid getting crazy. Things to do, *in order*:

- [x] Separate Trade library out of Magnum (mostly just buildsystem changes, resulting in a new `Magnum::Trade` CMake imported target (and library)
- [x] Move everything GL-related to `Magnum/GL` subdir and a new library (at this step this means even the `Image` class and everything, which means also the whole Trade library and all plugins now depend on GL library), not adding a new namespace yet
- [x] Add a new `Magnum::GL` namespace
  - [x] Convert `Magnum::Extensions::GL::*` to `Magnum::GL::Extensions::*` ?
  - [x] Convert `Platform::Context` to `Platform::GLContext`
- [x] Provide deprecated compatibility headers (e.g. `<Magnum/Buffer.h>` that points to `<Magnum/GL/Buffer.h>`)
- [x] Provide deprecated aliases (e.g. `Magnum::Buffer` that is a typedef to `Magnum::GL::Buffer`)
- [x] Adapt rest of the project to not use deprecated headers and aliases (again, until this is done, the non-deprecated build will fail)
- [x] Adapt Developers Guide to use new file locations and namespaces
- [x] Adapt all docs to not use deprecated headers and aliases (Doxygen won't complain because of the compat versions)
  - [x] Adapt all `@section` identifiers to use the GL namespace
  - [x] Rename `@extension` to `@gl_extension`
- [x] Extract `PixelFormat` / `PixelType` enums back to root namespace and provide some mapping to GL types (need to check what Vulkan has here)
- [x] Separate GL-specific stuff out of `PixelStorage` and put the rest back to root namespace
  - [x] Temporary assertions for pixel storage things that are not supported in ES / WebGL (will be later replaced with proper implementation coming from the Vulkan part)
- [x] Put Image classes back to root namespace
  - [x] also tests
- [x] Separate `MeshPrimitive` and `Mesh::IndexType` enum out of GL lib
- [x] ~~Think about how to separate sampler state out of GL lib (also enums straight in Trade, how to map to GL?)~~ It's a bunch of Sampler enums in the root namespace now
- [x] Now that `Trade` does not depend on GL anymore, fix the linking and fix that also `Primitives` and all plugins
- [x] Verify that all dependent repos and plugins and examples build with the new lib w/o any source change
- [x] ~~Update all packages and CIs to explicitly use `WITH_GL` and `WITH_TRADE`~~ no need to, this is enabled by default
- [x] Introduce `TARGET_GL` option, allowing to enable/disable GL support in other libraries
- [x] Make the SDL2 and GLFW app usable with `TARGET_GL` being turned off (i.e., so it doesn't even link to GL)
- [x] Look in the OpenGL / platform overview docs and check that nothing is terribly wrong
  - [x] Avoid saying that GL is the "core part", it's not
  - [x] Adapt magnum-gl-info "screenshot"
- [x] Rename `magnum-info` to `magnum-gl-info`. Backwards compatibility for this is not considered, please complain if you need that.
- [x] Rename all `MAGNUM_ASSERT_*_SUPPORTED()` macros to `MAGNUM_ASSERT_GL_*_SUPPORTED()` or `MAGNUM_GL_*()` (how to handle backwards compat?)
  - [x] also `MAGNUM_VERIFY_NO_ERROR()`
- [x] Changelog entries for all of above

After all the above this PR could be merged back to master, I think. Separating the remaining libraries from GL will be done incrementally over time.

Things I'm not sure how to do:

- [x] ~~Backward compatibility so people linking to `Magnum` will be linked to `Magnum::GL` and `Magnum::Trade` as well~~ -- there would be a cyclic dependency with this, so nope
- [x] ~~Should GL::BufferImage have GL::PixelFormat or Magnum::PixelFormat? The former would require a mapping table back to Magnum::PixelFormat, the latter will feel alien (formatExtra etc.)~~ It's the GL-specific format now.

Cc: @Squareys


mosra:
Trade library split was merged to master in 920db7aa73b13178742b2d2b18b087edf7bfe0a3.

mosra:
Expecting angry people with pitchforks and torches doing a witch hunt on me now, haha :see_no_evil: :fire:

Documentation and proper release notes / upgrading guide coming soon with the 2018.04 release.

PR #246, opened 2018-05-07
Fix uninitialized pointer check in GlfwApplication.

skogler:
Unfortunately, this pointer is checked when initializing the window, which makes the current release crash on application start. (Tested on VS2017 x64).

This small fix makes it work again. Not sure if you can easily get this into vcpkg, I am using a patched version locally right now.

### DIFF
--- a/src/Magnum/Platform/GlfwApplication.h
+++ b/src/Magnum/Platform/GlfwApplication.h
@@ -480,7 +480,7 @@ class GlfwApplication {
 
         static GlfwApplication* _instance;
 
-        GLFWwindow* _window;
+        GLFWwindow* _window{ nullptr };
         Flags _flags;
         #ifdef MAGNUM_TARGET_GL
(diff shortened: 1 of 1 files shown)


mosra:
Damn, sorry about this. I should have tested more :see_no_evil:.

Merged, thanks a lot. I also discovered that, compared to Sdl2Application, it was not creating core contexts properly (fixed in f30ca8929e1b0a0729d7d62f47bc5178256949b6) and did a bunch of other minor cleanup. Please complain loudly if you encounter any other problems ;)

If you use vcpkg, you can do just this to pull latest master instead of the stable 2018.02:

```bat
vcpkg install magnum --head
```

I hope you can live with that, patching the 2018.02 version in vcpkg *is* possible, but the turnaround time for adding new things to vcpkg packages is roughly a week.

skogler:
Nice, thanks for the quick reaction! Did not know about the head option, that is fine!

PR #253, opened 2018-06-05
external: fix ES2 and ES3 include paths in iOS

steeve:
These are the official include paths from Apple.


mosra:
Interesting. When I did the iOS port, I took the other platforms as a base (with the common `ES2/gl.h`) and it magically "just worked" with CMake, so I never questioned it. But this builds green as well and doesn't seem to break anything else, so as long as it helps you, why not :)

PR #255, opened 2018-06-23
Fix builds without GL

Squareys:
Hi @mosra !

I applied your patch for the fix for building `WITH_GL=OFF`. If building with `BUILD_DEPRECATED=ON`, it would not find External/OpenGL, which I made sure that it gets installed anyway now (untested!).

Cheers, Jonathan

The issue you get when trying to build a project using magnum that was build accordingly:
~~~
-- The C compiler identification is GNU 4.8.5
-- The CXX compiler identification is GNU 4.8.5
-- Check for working C compiler using: Ninja
-- Check for working C compiler using: Ninja -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler using: Ninja
-- Check for working CXX compiler using: Ninja -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Found Corrade: /usr/local/include  found components:  Utility Containers 
-- Found Corrade: /usr/local/include  found components:  Utility Containers rc 
-- Found Magnum: /usr/local/include   
-- LIB_SUFFIX variable is not defined. It will be autodetected now.
-- You can set it manually with -DLIB_SUFFIX=<value> (64 for example)
-- LIB_SUFFIX autodetected as '64', libraries will be installed into /builds/vhiterabbit/webvr-pong/deploy/lib64
-- Found Magnum: /usr/local/include  found components:  Magnum 
-- Configuring done
CMake Error in src/server/CMakeLists.txt:
  Imported target "Magnum::Magnum" includes non-existent path

    "/usr/local/include/MagnumExternal/OpenGL"

  in its INTERFACE_INCLUDE_DIRECTORIES.  Possible reasons include:

  * The path was deleted, renamed, or moved to another location.

  * An install or uninstall procedure did not complete successfully.

  * The installation package was faulty and references files it does not
  provide.
~~~

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -213,13 +213,15 @@ if(BUILD_STATIC)
 endif()
 
 # Check dependencies
-if(NOT TARGET_GLES OR TARGET_DESKTOP_GLES)
-    set(OpenGL_GL_PREFERENCE GLVND) # since CMake 3.11
-    find_package(OpenGL REQUIRED)
-elseif(TARGET_GLES2)
-    find_package(OpenGLES2 REQUIRED)
-else()
-    find_package(OpenGLES3 REQUIRED)
+if(WITH_GL OR TARGET_GL)
+    if(NOT TARGET_GLES OR TARGET_DESKTOP_GLES)
+        set(OpenGL_GL_PREFERENCE GLVND) # since CMake 3.11
+        find_package(OpenGL REQUIRED)
+    elseif(TARGET_GLES2)
+        find_package(OpenGLES2 REQUIRED)
+    else()
+        find_package(OpenGLES3 REQUIRED)
+    endif()
 endif()
 
 # Configuration variables (saved later to configure.h)
--- a/src/Magnum/CMakeLists.txt
+++ b/src/Magnum/CMakeLists.txt
@@ -185,7 +185,9 @@ endif()
 target_include_directories(Magnum PUBLIC
     ${PROJECT_SOURCE_DIR}/src
     ${PROJECT_BINARY_DIR}/src)
-if(BUILD_DEPRECATED AND TARGET_GL) # TODO: remove once compat gets dropped
+if(BUILD_DEPRECATED) # TODO: remove once compat gets dropped
+    # Some deprecated APIs use headers (but not externally defined symbols)
+    # from the GL library, link those includes as well
     target_include_directories(Magnum PUBLIC
         ${PROJECT_SOURCE_DIR}/src/MagnumExternal/OpenGL)
 endif()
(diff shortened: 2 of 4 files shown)


CMakeLists.txt
```diff
@@ -213,13 +213,15 @@ if(BUILD_STATIC)
...
 
 # Check dependencies
-if(NOT TARGET_GLES OR TARGET_DESKTOP_GLES)
-    set(OpenGL_GL_PREFERENCE GLVND) # since CMake 3.11
-    find_package(OpenGL REQUIRED)
-elseif(TARGET_GLES2)
-    find_package(OpenGLES2 REQUIRED)
-else()
-    find_package(OpenGLES3 REQUIRED)
+if(WITH_GL OR TARGET_GL)
```
mosra: After thinking about this a bit more, this should be just `if(WITH_GL)`. `TARGET_GL` can be only enabled if `WITH_GL` is enabled.

src/Magnum/CMakeLists.txt
```diff
@@ -185,7 +185,9 @@ endif()
 target_include_directories(Magnum PUBLIC
     ${PROJECT_SOURCE_DIR}/src
     ${PROJECT_BINARY_DIR}/src)
-if(BUILD_DEPRECATED AND TARGET_GL) # TODO: remove once compat gets dropped
+if(BUILD_DEPRECATED) # TODO: remove once compat gets dropped
+    # Some deprecated APIs use headers (but not externally defined symbols)
+    # from the GL library, link those includes as well
```
mosra: In many places I used the following preprocessor statement:

```cpp
#if defined(MAGNUM_BUILD_DEPRECATED) && defined(MAGNUM_TARGET_GL)
```

but there are apparently cases that I missed, having just `#ifdef MAGNUM_BUILD_DEPRECATED`. Sorry.

... so, instead of installing the GL stuff even when the user *explicitly doesn't want to*, the deprecated stuff depending on GL should not be compiled at all when `TARGET_GL` is `OFF`. So all these CMake patches that add GL stuff should be removed and replaced by preprocessor patches that remove GL stuff.

I have no easy way of testing this myself without creating a barebones VM from scratch, so I'm kinda depending on you to find and fix all those cases :) Thank you :+1:

Squareys: Alright, that will require time that I currently don't have, though. I will have the time to work through this as of the 9th of July, I hope that is okay.

mosra:
I fixed most of the issues affecting core library in 89d988383aa357f799ab5ad7714fce947d16dea9 while integrating the Vulkan Triangle example. There might be some more left, but since none of the changes of this PR would get used anyway, I'm closing it.

PR #267, opened 2018-08-05
Cubic Hermite splines

mosra:
Important for animations, among other things, postponed from #191 to unblock progress, but still hopefully on track for 2018.08. Cubic Hermite splines were chosen because they are a superset of all other commonly used splines and thus having support for these will also mean support for Catmull-Rom, Bezier, TCB and others.

A "point" representation was chosen over a "segment" representation in order to better support the use in animations. [Math::Bezier](http://doc.magnum.graphics/magnum/classMagnum_1_1Math_1_1Bezier.html), OTOH, has a segment representation.

TODO:

- [x] Basic Hermite point class
  - [x] Docs explaining what it is, why it was chosen and why it is a point and not a segment
  - [x] Tests for the surface API
  - [x] Split into multiple commits
- Helper constructors to create the curve from
  - [x] Bezier
  - [x] ~~Catmull-Rom~~ postponed, see #101
  - [x] ~~[TCB](https://en.wikipedia.org/wiki/Kochanek–Bartels_spline)~~ postponed, see #101
  - [x] ~~[The representation that glTF uses](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#appendix-c-spline-interpolation)~~ no idea how the heck is that called, so it'll be just private to the glTF importer
- [x] `Bezier::fromCubicHermite()` (just because I have the class)
- [x] Cubic Hermite quaternions (needed by glTF), complex numbers, dual quaternions, ... -- or rather have a CubicHermite<T> where T is the underlying type?
- Hermite curve interpolation and tests
  - [x] `Math::select()`
  - [x] `Math::lerp()`
  - [x] `Math::splerp()`
  - [x] `Math::lerp()` for quaternions
  - [x] `Math::splerp()` for quaternions
- [x] `Animation::Interpolation::Spline` and support in related APIs
- [x] Support in `Trade::AnimationDataType` APIs
- [x] Implement these in the glTF importer to verify the API is sane to populate
- [x] Implement these in the magnum-player to verify the API is sane to use
- [x] Test translation to verify the implementation is sane 
- [x] Add "shortest path" variants for `lerp()`, `slerp()` and `sclerp()`
  - [x] Use them by `interpolatorFor()`
  - [x] Benchmark
- [x] Extend the transformation documentation and mention all the interpolation stuff we have
- [x] Create at least a dummy page for animations
- [x] Test rotation to verify the implementation is sane

### DIFF
--- a/doc/animation.dox
+++ b/doc/animation.dox
@@ -25,25 +25,24 @@
 
 namespace Magnum {
 
-/** @page features Feature guide
-@brief Fundamental principles and design goals.
+/** @page animation Keyframe-based animation
+@brief Layered framework for authoring, loading, optimizing and playing back animations
 
-Overview and tutorials of high-level feature groups in Magnum. It's not
-necessary to read through everything, pick only what you need.
+@tableofcontents
+@m_footernavigation
 
--   @subpage platform --- @copybrief platform
--   @subpage types --- @copybrief types
--   @subpage matrix-vector --- @copybrief matrix-vector
--   @subpage transformations --- @copybrief transformations
--   @subpage animation --- @copybrief animation
--   @subpage plugins --- @copybrief plugins
--   @subpage opengl-wrapping --- @copybrief opengl-wrapping
--   @subpage shaders --- @copybrief shaders
--   @subpage scenegraph --- @copybrief scenegraph
--   @subpage shapes @m_class{m-label m-danger} **deprecated** ---
-    @copybrief shapes
--   @subpage debug-tools --- @copybrief debug-tools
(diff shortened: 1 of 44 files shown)


Squareys:
[AbstractImporter Documentation](http://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1AbstractImporter.html) subtitle lists what it is able to import and missing "animations" there :)

_(as requested to put here via gitter)_

mosra:
Merged into master. Shortest path interpolation is part of this PR as well, the remaining issues are glTF-specific, so handling them in the glTF importer itself instead.

PR #268, opened 2018-08-05
Update building.dox

uzername:
This change warns users about possible issue while building examples.


mosra:
Hi, thanks for the contribution! :+1: I feel like the need for enabling Sdl2Application was documented on enough places already (and *especially* in the Getting Started Guide), so I only improved the wording a bit in 72696ee7d1f9087b6129a7aa9af344a5f7681a9b (listing you as a co-author). The `CMAKE_PREFIX_PATH` is mentioned in the [Manual Build](http://doc.magnum.graphics/magnum/building.html#building-manual) section and doing a full CMake intro is out of scope of the docs.

PR #275, opened 2018-08-30
Math: table-based half-float packing/unpacking.

mosra:
Was originally abandoned, because I found a much more straightforward and better documented solution from rygorous. 

The table-based implementation could be useful as a part of the "batch" math algo library as it is much faster for processing many numbers at once (keeping the tables warm in the cache). A variant of it (with a run-time-generated table) is currently used in the Half test and benchmark.


mosra:
Merged as a new pair of `packHalfInto()` / `unpackHalfInto()` functions, taking 2D strided array views (the original code took just a contiguous 1D array view, which is not enough for current use cases). The ALU-based variant turned out to be much worse than current `(un)packHalf()`, so I just dropped it; all tests are just reused from what was already on master.

This might get removed at some point in the future when most platforms have a corresponding intrinsic (FMA/FP16 instructions on x86 have it, ARM has it, but e.g. WebAssembly doesn't) or when a SIMD-accelerated variant proves to be faster everywhere.

PR #276, opened 2018-08-30
[WIP] C++14 constexpr additions

mosra:
Code from a very old branch, could be reused instead of starting from scratch. What needs to be done:

- [ ] create some `CORRADE_CONSTEXPR14` macro that compiles to `constexpr` only under C++14
- [ ] separate tests that are built only for C++14 and testing this C++14-only functionality
- [ ] test with *both* constexpr and non-constexpr in order to get correct code coverage (and this may also uncover some bugs)

### DIFF
--- a/src/Magnum/Math/BoolVector.h
+++ b/src/Magnum/Math/BoolVector.h
@@ -94,7 +94,7 @@ template<std::size_t size> class BoolVector {
         constexpr BoolVector(const BoolVector<size>&) = default;
 
         /** @brief Copy assignment */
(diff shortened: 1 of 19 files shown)


mosra:
Closing as the diff here is too old and fixing the merge conflicts + applying the `constexpr` to all APIs that were added since would be more work than starting from scratch.

The `CORRADE_CONSTEXPR14` macro exists now and the correct way to do this is like in mosra/corrade@0412daf8a73423fd04892d965b394d2416694174.

PR #277, opened 2018-08-30
[WIP] Audio: Source buffer queuing.

mosra:
Another WIP, back from 2013. This is still not done in `master`, hopefully some work from here can be reused.

> How to do the API? Can't pass pointers, as we don't know the pointer<->ID mapping (and explicitly managing it would be too memory intensive), passing buffers by value would be too inconvenient (much of the std::move() crap).

### DIFF
--- a/src/Audio/Source.cpp
+++ b/src/Audio/Source.cpp
@@ -35,6 +35,24 @@ Source& Source::setBuffer(Buffer* buffer) {
     return *this;
 }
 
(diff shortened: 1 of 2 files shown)


mosra:
Nice, even with a naked `new` and a corresponding memory leak. Done *much better* by @williamjcm in #338, so I'm going to let this piece rot in peace.

PR #278, opened 2018-08-30
[WIP] Trade: add texture coordinate bindings to MeshObjectData

mosra:
Another WIP back from 2013, adds an ability to specify which texture uses which texture coordinate array. Also not present in `master` yet.


mosra:
Superseded by much cleaner, much more flexible and much easier-to-implement #438.

PR #279, opened 2018-08-30
[WIP] Vector fields implementation

mosra:
From 2012. Inspired by http://bitsquid.blogspot.com/2012/09/a-data-oriented-data-driven-system-for.html.

The goal of this one was describe a parametrized math equation using a high-level API which compiles to nicely vectorizable code. Then using that to process large batches of data, with the idea that this is faster than writing scalar math code by hand.

Besides this compiling to a bunch of C++ templates, I had in mind export of GLSL code (or now, SPIR-V).


mosra:
The sole purpose of this PR was to archive the original code, closing to not make it look like there's something left to be done.

PR #280, opened 2018-09-04
Add LGTM.com build configuration for C/C++ analysis

sjvs:
Following discussion on `corrade`: https://github.com/mosra/corrade/pull/50

It turned out that just copying `travis-desktop.sh` with some minor tweaks did the trick. Feel free to make adjustments as you see fit, of course. As soon as this PR is merged I'll be able to enable C/C++ analysis and C/C++ results will appear on LGTM.com soon: https://lgtm.com/projects/g/mosra/magnum/

cc @mosra @xcorail

### DIFF
--- a/.lgtm.yml
+++ b/.lgtm.yml
@@ -1,0 +1,1 @@
+package/ci/lgtm.yml
--- a/package/ci/lgtm.sh
+++ b/package/ci/lgtm.sh
@@ -1,0 +1,62 @@
+#!/bin/bash
+set -ev
+
+# Corrade
+wget -O corrade-master.zip https://github.com/mosra/corrade/archive/master.zip?foobar=`time +%s`
+unzip corrade-master.zip
+
+# Build Corrade
+cd corrade-master
+mkdir _lgtm_build
+cd _lgtm_build
+cmake .. \
+    -DCMAKE_CXX_FLAGS="$CMAKE_CXX_FLAGS" \
+    -DCMAKE_INSTALL_PREFIX=$HOME/deps \
+    -DCMAKE_INSTALL_RPATH=$HOME/deps/lib \
+    -DCMAKE_BUILD_TYPE=Debug \
+    -DBUILD_DEPRECATED=$BUILD_DEPRECATED \
+    -DWITH_INTERCONNECT=OFF \
+    -G Ninja
+ninja install
+cd ../..
+
+mkdir _lgtm_build
+cd _lgtm_build
+# Not using CXXFLAGS in order to avoid affecting dependencies
+cmake .. \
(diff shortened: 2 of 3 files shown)


mosra:
@sjvs @xcorail I was finally able to look at this. Thanks for doing the initial work.

But... I can't seem to find a way to trigger a build (or even get some processing status) for a branch that's not `master`. At the moment I have a modification of this PR pushed into the `lgtm` branch but I was not able to find anything in the web UI that would tell me it recognizes the branch, has some processing scheduled for it or lets me switch to it... and I don't want to just put a random config on master because it will most definitely not work on a first try. I am logged in, but I don't see any status page or logs from builds being processed.

The whole processing is still a bit opaque to me -- doesn't it need the tests to be run? I think just building the sources isn't enough to discover problems (any my experience with all other analyzers/sanitizers is that you need to run the code to actually *see* something).

xcorail:
Hey @mosra 

> I can't seem to find a way to trigger a build (or even get some processing status) for a branch that's not master.

Indeed, LGTM at the moment analyzes only your default branch. However, if you activate automated code review (which I think you did) all your PRs will be analyzed, whatever the branch they're on.

> I don't want to just put a random config on master because it will most definitely not work on a first try.

Makes sense. So in this case you have put the configuration on the `lgtm` branch, so just create a PR on this branch, with any dummy code change, LGTM will analyze it, and this way you'll be able to test your new configuration before putting it on `master`.

> The whole processing is still a bit opaque to me -- doesn't it need the tests to be run?

No. We only build the sources, we don't run the program. It's true that most static analyzers will miss many problems, that can only be found by dynamic analyzers (that run the code). Our technology is somehow in the middle. Although we don't run the program, we are able to taint data flowing through your code, and identify complex control flow problems.
Does that make more sense?
 
Cheers
Xavier

mosra:
I decided to not go further with this. Maintaining the [current CI build matrix](https://magnum.graphics/build-status/) which has over 100 builds is enough of a maintenance burden and I'm *not* going to add another tool, unless I can either:

- see an independent review showing the benefits it had on another large project (like PVS-Studio does, for example), and
- run it locally and see non-trivial found issues, *before* having to trial-and-error against a 3rd party service

Seeing that it found exactly 0 alerts in Corrade (apart from two cases of exceptions being passed by value, which are very minor and gone since anyway), I don't expect anything serious being found here either.

Currently I have covered the code with generally about 90% coverage, on five different compiler implementations (and over a dozen of compiler versions), in some cases with Clang Sanitizers enabled. That has been more than enough in the past ~five years with no strange crashes or misbehavior happening for users *at all* (except GL drivers, of course).

PR #289, opened 2018-10-23
distance field: fix distance field cache on OpenGL ES

dolphineye:
## Motivations
Distance field text is not rendering on some Huawei devices since the OpenGL renderer will complain about an incomplete framebuffer attachment.

## Proposed solution
This pull request fixes the distance field cache generation on Huawei devices by changing the `TextureFormat` from `GL::TextureFormat::Red` to `GL::TextureFormat::R8`.

## Testing done
On-device testing on both iOS and Android.

### DIFF
--- a/src/Magnum/Text/DistanceFieldGlyphCache.cpp
+++ b/src/Magnum/Text/DistanceFieldGlyphCache.cpp
@@ -42,7 +42,7 @@ DistanceFieldGlyphCache::DistanceFieldGlyphCache(const Vector2i& originalSize, c
     #elif !defined(MAGNUM_TARGET_WEBGL)
     /* Luminance is not renderable in most cases */
     GlyphCache(GL::Context::current().isExtensionSupported<GL::Extensions::EXT::texture_rg>() ?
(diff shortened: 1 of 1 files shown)


steeve:
LGTM, the fix works for us so we're not stuck

dolphineye:
> This is the same issue that we were having on iOS. I was blocked by not having access to an ES device that exposes `EXT_texture_storage` on ES2 in order to test, but if you say Huawei has that, then I should be able to do the test.
> 
> Postponing to after I get the 2018.10 release out, if that's okay? :)

Sure thing. We do have a workaround in the meantime, so that's fine by me. Let us know when you got the fix out :)

mosra:
Thank you.

(For the record, a verified workaround for this is [disabling](https://doc.magnum.graphics/magnum/classMagnum_1_1GL_1_1Context.html#GL-Context-command-line) the `GL_EXT_texture_storage` extension on ES2.)

mosra:
Merged a while ago in 65b1931c62f7952f68397e6e91af9f8381268320. While testing I discovered the RGB format fallback was not correct either so that's fixed as well.

Thank you for the contribution!

PR #297, opened 2018-11-29
Fix compil error in GlyphCache.cpp + hunter

pthom:
Hi,

I am trying to add corrade and a magnum package to hunter. 
I will tell you more about it in the next days.

While compiling magnum inside the hunter CI, I encountered a compilation error inside GlyphCache.cpp : 

````cpp
src/Magnum/Text/GlyphCache.cpp:68:12: error: call to member function 'insert' is ambiguous
    glyphs.insert({0, {}});
    ~~~~~~~^~~~~~
/usr/lib/gcc/x86_64-linux-gnu/7.3.0/../../../../include/c++/7.3.0/bits/unordered_map.h:436:7: note: candidate function
      insert(node_type&& __nh)
      ^
/usr/lib/gcc/x86_64-linux-gnu/7.3.0/../../../../include/c++/7.3.0/bits/unordered_map.h:584:7: note: candidate function
      insert(value_type&& __x)
      ^
/usr/lib/gcc/x86_64-linux-gnu/7.3.0/../../../../include/c++/7.3.0/bits/unordered_map.h:578:7: note: candidate function
      insert(const value_type& __x)
      ^
etc
etc
etc
````

There is however something strange : I cannot reproduce this compilation error on my computer, it only happens inside hunter's CI.

It happens under MSVC (see https://travis-ci.org/pthom/hunter/jobs/461046587) and under clang (see https://travis-ci.org/pthom/hunter/jobs/461046587) .

I could solve this problem with this patch (see https://ci.appveyor.com/project/pthom/hunter/builds/20633842/job/bjsy2ke0kn1v46cw and https://travis-ci.org/pthom/hunter/jobs/461064174 ). You wil see that I am still struggling with hunter :-), but this file now compiles correctly.

I know that my correction could have been more terse, with something like : 

````cpp
glyphs.insert({0,  {{}, {}}  });
````

However, the CI testing process with hunter is quite tedious, so that I did not want to risk another failure due to initializer lists.

Did you encounter this error on another compiler ? I am surprised it was caught only by hunter's CI.

Thanks

### DIFF
--- a/src/Magnum/Text/GlyphCache.cpp
+++ b/src/Magnum/Text/GlyphCache.cpp
@@ -65,7 +65,7 @@ void GlyphCache::initialize(const GL::TextureFormat internalFormat, const Vector
         .setStorage(1, internalFormat, size);
 
     /* Default "Not Found" glyph */
-    glyphs.insert({0, {}});
+    glyphs.insert({0,  {{}, {}}  });
 }
 
 std::vector<Range2Di> GlyphCache::reserve(const std::vector<Vector2i>& sizes) {


pthom:
PS : if you are interested, some more info about my status concerning the hunter build.

I hope I am not interfering with some work that was ongoing, and that you will be interested in this proposition.

#### Corrade : 
I have a first working draft. See https://github.com/ruslo/hunter/issues/1633
* The CI successfully builds a package under clang, gcc, osx, msvc 2015, and 2017 (and sucessfuly links an example program with it). 
* The CI fails unders iOS and Android (missing `corrade-rc` exe), but it can work as long as `corrade-rc` is in the path.
* The CI fails unders windows with `msys` and `mingw`(this is not very important)

I did not have to make a fork of the corrade repo. 
However, the file magnum/src/CMakeLists.txt overrides CMAKE_CXX_FLAGS, and it might be required to correct this for hunter integration.
````cmake
# On MSVC remove /W3, as we are replacing it with /W4
if(CMAKE_CXX_COMPILER_ID STREQUAL "MSVC" OR CMAKE_CXX_SIMULATE_ID STREQUAL "MSVC")
    string(REPLACE "/W3" "" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
endif()
````

#### Magnum:
I opened an issue at hunter's main repo (https://github.com/ruslo/hunter/issues/1635)
For the moment, the hunter CI seems to successfully build the magnum package under windows and linux, but it fails when building an example program that uses `Sdl2Application`

I had to to fork and patch magnum, in order to add some simple (and optional) commands for hunter integration. 

You can see my fork here : https://github.com/pthom/magnum/tree/hunter-v2018.10

You can also see the corresponding work inside hunter here (this is wip !) : 
https://github.com/pthom/hunter/compare/pr.corrade...pthom:pr.magnum?expand=1

mosra:
Hi!

Happy to see this project being added to Hunter. There was a similar discussion a while ago in https://github.com/mosra/magnum-bootstrap/issues/13, but nothing ever came out of it.

Regarding building, CI, errors etc.: please see https://magnum.graphics/build-status/. Each project is tested in almost 30 configurations on various platforms and I'm working very hard on keeping the `master` green at all times. CI scripts are in the `package/ci/` directory of each project, it's good to look there for hints on how to build on each platform (in particular, you can see how Android and MinGW builds are handled there, they're known to work correctly).

That said, I have no idea why this fails on GCC 7.3.0 -- the CIs test on GCC 4.7, build on GCC 5 (Ubuntu packages) works as well and GCC 6 and 8 on my machine never had a problem with this particular piece of code. So far it looks to me like there was some temporary regression in 7.3.0 that caused this. I'll see if I can dig up GCC 7 from somewhere and try on it. Once I am sure I understand why this happens, I'll merge your patch. Besides that, the `insert()` is kept only to have compatibility with GCC 4.7 (it doesn't have `emplace()` yet) and I am in process of dropping 4.7 compat anyway, so this peculiarity could get fixed as a side-effect changing the code to use `emplace()` instead.

Regarding Hunter, a few comments on what I've seen in your work: 

- The default CMake options are the "barebones minimal build", which is not really useful and enabling a bunch more things is desired (of course depending on what the particular platform supports). Not sure how Hunter handles optional features (or if it has them at all), this for example worked very well with Vcpkg (there you can opt-in for each particular feature by saying e.g. `vcpkg install magnum[tgaimporter]`). Check the Homebrew, Debian and Gentoo packages in `package/ci` for a reasonable set of things to have enabled by default.
- In case you missed it, the [magnum-plugins](https://github.com/mosra/magnum-plugins) repository is full of essential plugins that people will want to use with magnum, so packaging those is a good idea as well. The difference from plugins in the core repository is that majority of these depend on external libraries and that's why they're maintained separately.
- The GLUT application is deprecated and scheduled for removal anyway, so you don't have to bother with it at all. Similarly with the Shapes library. Simply remove all traces of these from the package files :)
- Pulling Corrade using `"https://github.com/pthom/hunter/archive/test-corrade-v2018.10.tar.gz"` won't work, since that version works only with Magnum 2018.10. Magnum `master` also needs Corrade `master`. 
- I already expressed this opinion in the issue linked above -- I have to admit I'm not exactly happy about Hunter "taking over" the buildsystem, adding the monster `HunterGate.cmake` file and polluting the build scripts with various `if(HUNTER_ENABLED)`. If this all could be done from outside or contained in a single CMake script, without requiring those things to be scattered around to the project itself, I would be *much* happier. No other package we maintain (ArchLinux, Debian, Gentoo, Homebrew, Vcpkg) requires such intrusive buildsystem changes. Some ideas:
  - The `HunterGate.cmake` is huge and adding the same copy of it to each of the various magnum project repositories (corrade, magnum, magnum-plugins, magnum-extras, magnum-integration, magnum-examples) doesn't make any sense. Is there a possibility to have it pulled from some Hunter-maintained Git submodule, for example? Another option I could be okay with is putting it to the `toolchains/` submodule (https://github.com/mosra/toolchains), this one is referenced in all projects.
  - For the various `hunter_add_package(corrade)`, would it be possible to have this done from the outside as well? Again I'd like to have the buildsystem totally independent from the particular package manager, if there would be a Conan package support, I also wouldn't want it to creep into all CMake scripts. That also makes it easier to update the package indepdenently of the repository and reduces a risk of having the package broken by buildsystem updates. I could imagine a Hunter-specific CMake toolchain file residing in `package/hunter/Hunter.cmake` having all those Hunter-specific commands inside and then this whole thing being enabled and set up by passing `-DCMAKE_TOOLCHAIN_FILE=../package/hunter/Hunter.cmake` to CMake. Not a Hunter expert myself, so I don't know about potential problems with this approach. I could be also fine with having this single thing in the root CMakeLists, if it really wouldn't be possible to it any differently:

    ```cmake
    if(HUNTER_ENABLED)
        include(${PROJECT_SOURCE_DIR}/package/hunter/Hunter.cmake)
    endif()
    ```

  - For the `if(NOT HUNTER_ENABLED)` around modification of `CMAKE_CXX_FLAGS`: I guess Hunter is putting some `variable_watch()` on `CMAKE_CXX_FLAGS`? This patching *is* needed, because Magnum adds `/W4` to compile flags, CMake defaults have `/W3` and having both specified makes MSVC complain quite loudly. Would the following work? It checks the variable before attempting to modify it:

    ```cmake
    if((CMAKE_CXX_COMPILER_ID STREQUAL "MSVC" OR CMAKE_CXX_SIMULATE_ID STREQUAL "MSVC") AND CMAKE_CXX_FLAGS MATCHES "/W3") 
        string(REPLACE "/W3" "" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
    endif()
    ```

pthom:
Hello,
Thanks for your answers !
I think I will open a separate issue for this.

I do not have much time to answer all of them now (this is spare time project !), but I would like to know if  a structure like this would be better for you 

image

* The init of hunter must be done in two steps : one before `add_project`, and one after : 
  `HunterInit.cmake` includes HunterGate from a submodule (https://github.com/hunter-packages/gate)
* `HunterFetchDeps.cmake` groups all the hunter_add_package in a common location:

````cmake
hunter_add_package(corrade)
if(WITH_GLFWAPPLICATION)
    hunter_add_package(glfw)
endif()
if(WITH_SDL2APPLICATION)
    hunter_add_package(SDL2)
endif()
````

mosra:
Oh, wow, you're doing huge amounts of work here.

> As I said, this is a spare time project. I tried my best but I must admit it proved very time consuming, and I hope to be able to conclude soon :-)

I fully get that this is a spare time project and I'm thankful you're spending your spare time on this. That said, If it feels like I'm too demanding, please tell me :) ... I have to admit that that I'm getting a bit frigtened by the *extreme* requirements from Hunter in order to add a project to it. Magnum is already quite strict regarding testing, coding style and portability of all its code, but this ... this is a whole new level.

So, I finally got some more time to look into this:

- Thanks, now that I see that the GlyphCache compilation error is due to C++17, I could reproduce easily locally. Pushed your fix (with an explanatory comment) as 50bf219e53c4b3a40493b058f2cca9f7b8ef4c47.
- Also pushed your emscripten option typo fix as d7f015c3ea94ea36c32636c0efbd1a18f684e925 (this was silly, heh)
- Your (now disabled) Clang Analyzer build emitted a warning for Corrade (https://travis-ci.org/pthom/hunter/jobs/462492058#L1356), fixed in mosra/corrade@f21c5f6bd47f20cdd06f7e85c0ed43de2545f64c. I also went through other Analyzer warnings, but all of them were false positives coming from tests, which should not affect the main build at all.
- The MinGW error (about `using rebind = _Up*;`) is [a bug in GCC 7.2](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=83226), fixed in 7.3. Any chance the MinGW build could use newer GCC? Unfortunately the bug seems to have no easy workaround, I would need to change quite a lot of code to not use references as a map value type. Unless you really need to have MinGW supported, I would suggest waiting a bit until Hunter upgrades to 7.3 or newer.
- The MSYS error looks like it can't find any of the built libraries for some reason. This is usually because the libraries have some unexpected extension or CMake is trying to find files with some unexpected extension. Having a copy of `CMakeCache.txt` and seeing what `CORRADE_*` variables are set to `*-NOTFOUND` would help. I'm not really sure how's the MSYS build different from MinGW (since on Windows I usually have MinGW installed by MSYS2). Is it some Cygwin thing? Again, unless you really need MSYS I would suggest to disable it for the initial release ... otherwise we'll be stuck in buildsystem issues forever :)
- The iOS / Android builds need access to the `FindEGL` and `FindOpenGLES3` (or `FindOpenGLES2`) CMake modules that are in `modules/` directory of Magnum itself. Usually the projects simply copy these into their own `modules/` dir and go with that. So you probably need to do something similar for the CIs.
- Just a minor thing: in e.g. https://github.com/ruslo/hunter/commit/953a8a83aef6e55891cf7bf15afcbb23785f599b, I see Android being spelled as Androïd (umlaut(?) above i). Not sure if intentional, never saw it spelled like this :)

> I did not miss it, but I do not have enough time (this port to hunter required actually a lot of time). I hope another hunter contributor will work on it later.

Got it. You did a huge amount of work already. Once Magnum is in, I'm sure someone would want plugins (and the other repos) in there as well and they'll contribute accordingly.

> As a matter of fact, I read in the hunter doc that it is marked as critical not to mess with CMAKE_CXX_FLAGS

Yeah, this particular thing is more of a hack that fixes MSVC to not print out extreme amounts of warnings. I don't think it's too important to fix, Hunter is probably supplying its own flags anyway.

> https://github.com/pthom/hunter/blob/pr.magnum/cmake/projects/magnum/hunter.cmake

Here, take this instead: https://gist.github.com/mosra/2446f114216e2c0ba66f1309287c4fb2 (it's hard to know which components are relevant on which platform unless you are *very* familiar with the project, also there's like ten thousands of them)

> https://github.com/pthom/magnum_hunter

This is wonderful. This fits greatly into the style of [magnum-bootstrap](https://github.com/mosra/magnum-bootstrap) projects. I'll happily integrate it there.

- - -

To wrap this up and continue with merging the actual changes needed for Hunter support: **can you open a dedicated PR** for [your magnum_hunter branch](https://github.com/mosra/magnum/compare/master...pthom:magnum_hunter) so we can continue the discussion there? In short, I'm very happy with how you managed to make the buildsystem additions self-contained but I'm still unsure about one or two things and I feel like having the actual code at hand would be better for the discussion and my silly questions :)

Thank you for everything you did here!

mosra:
Since the change was integrated in https://github.com/mosra/magnum/commit/50bf219e53c4b3a40493b058f2cca9f7b8ef4c47, I think I can close this one :) The rest is done in #298.

PR #298, opened 2018-12-03
Magnum hunter

pthom:
Hi,
As discussed, here is the PR for the hunter related modifications.
Cheers


CMakeLists.txt
```diff
@@ -49,8 +49,20 @@ endif()
 if(NOT CMAKE_VERSION VERSION_LESS 3.4.0 AND NOT CMAKE_SYSTEM_NAME STREQUAL "Android")
     set(LANG CXX)
 endif()
+
+
+option(HUNTER_ENABLED "Get third parties (Corrade, glfw, SDL2) via hunter " OFF)
+if (HUNTER_ENABLED)
+    set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_CURRENT_LIST_DIR}/package/hunter")
+    include(HunterInit)
+endif()
```
mosra: As far as I can see, Corrade did not need any patching for Hunter support because 

- it doesn't have any dependencies
- thus no `hunter_add_package()` needs to be called
- thus there's no need to include HunterGate

However, here because we have dependencies and we need a bunch of `hunter_add_package()` for them, we need to include HunterGate. Is that correct?

I was thinking that, since Hunter will be *very seldom* used to develop Magnum itself and it will be mostly only used to develop projects *depending on* Magnum, we could skip including HunterGate and have just the `hunter_add_package()` calls wrapped in `if(HUNTER_ENABLED)`, assuming HunterGate would be already included from some super-project. But I could not find any concrete specification of this in the docs so I'm not sure if this is at all possible. Also, not sure about the behavior when Hunter CIs are producing binary versions of Magnum. So we *have to* include HunterGate in any case. Right?

package/hunter/HunterInit.cmake
```diff
@@ -0,0 +1,12 @@
...
+  https://raw.githubusercontent.com/hunter-packages/gate/master/cmake/HunterGate.cmake
+  ${CMAKE_CURRENT_LIST_DIR}//HunterGate.cmake
+)
+
+# FIXME : will need to replace this url with the main hunter url !
+include("${CMAKE_CURRENT_LIST_DIR}/HunterGate.cmake")
+HunterGate(
+  URL "https://github.com/pthom/hunter/archive/hunter_magnum_rc4.tar.gz"
+  SHA1 "31554df9e988f27f4daa9f4df9529d75007500bf"
+)
```
mosra: You just answered my question here -- I was not sure why we want to download Magnum sources inside HunterGate. Downloading Hunter package archive makes more sense, yup :)

package/hunter/HunterInit.cmake
```diff
@@ -0,0 +1,12 @@
+file(
+  DOWNLOAD
+  https://raw.githubusercontent.com/hunter-packages/gate/master/cmake/HunterGate.cmake
+  ${CMAKE_CURRENT_LIST_DIR}//HunterGate.cmake
+)
```
mosra: I'm not sure what the exact behavior of CMake's `file(DOWNLOAD)` (could not find any concrete info on it) and I have a bad feeling that it could attempt a download on every CMake run. Could this be wrapped in `if(NOT EXISTS  ${CMAKE_CURRENT_LIST_DIR}/HunterGate.cmake)`?

The other option I mentioned in #297 was putting the file into the toolchains repository. But that would require the user to explicitly init the toolchains submodule, which is not needed for anything else. So I think this is a better solution :+1: However, to prevent potential backwards-incompatible changes, could we pin it to some concrete revision? Current master is `https://raw.githubusercontent.com/hunter-packages/gate/c4e5b3e4d1c97fb46ae86af7f59baac95f7f1cf0/cmake/HunterGate.cmake`.

pthom:
Hi, some updates about hunter CI:

### Corrade Hunter CI
We will soon have green lights on corrade in the CI
* https://travis-ci.org/pthom/hunter/builds/462832221
* https://ci.appveyor.com/project/pthom/hunter 
(I gave up on msys and mingw, as this requires to upgrade the compiler)

### Magnum Hunter CI
I rebased all my code onto your last commits
* https://travis-ci.org/pthom/hunter/builds/462840758
* https://ci.appveyor.com/project/pthom/hunter/builds/20731321
(I gave up on msys and mingw, as they do not work in corrade)
I did not solve the issue with iOS (`[hunter ** FATAL ERROR **] 'find_package(EGL)' should be called with REQUIRED`). I have read to many logs, my eyes are bleeding :-)
Concerning this build, I had to upgrade the CI from ubuntu 14 to 16 (and I had to add libgl1-mesa-dev in order to have opengl)

CMakeLists.txt
pthom: > As far as I can see, Corrade did not need any patching for Hunter support because
> it doesn't have any dependencies
> thus no hunter_add_package() needs to be called
> thus there's no need to include HunterGate

That is right

> However, here because we have dependencies and we need a bunch of hunter_add_package() for them, we need to include HunterGate. Is that correct?
I do not think we need to include HunterGate into corrade : it does not need to do any call to `hunter_add_package`.

Concerning magnum, yes we do need it. 


> I was thinking that, since Hunter will be very seldom used to develop Magnum itself and it will be mostly only used to develop projects depending on Magnum, we could skip including HunterGate and have just the hunter_add_package() calls wrapped in if(HUNTER_ENABLED), assuming HunterGate would be already included from some super-project. But I could not find any concrete specification of this in the docs so I'm not sure if this is at all possible

I do not think that it is possible and I would not recommend it.

Let me explain: this inclusion is critical in order to be able to test the hunter build locally (and not wait for travis to fail).

Let me show you how the magnum_hunter repo can be used to test the magnum build in an environment that reproduces travis configuration. 

Step 1 : launch a docker image that reproduces travis build config:
````bash
> cd /path/to/magnum_hunter/Docker/ubuntu
➜  ubuntu git:(master) ✗> ./docker_create_image.sh  # Create a docker image close to travis CI config
>  ubuntu git:(master) ✗ ./docker_login.sh # log inside the container (beware, all modifs are removed after logout)
`````

Step 2 : you are now inside a docker container, where /sources_docker <=> /path/to/magnum_hunter

Example 1 : build magnum : this will build corrade via hunter, and magnum the standard way, using the toolchain gcc-7-cxx14

`````bash
➜  /sources_docker git:(master) ✗ ./TLDR_hunter.py test-build magnum gcc-7-cxx14
`````

Example 2 : build magnum_example_app_with_submodules. Since this project uses hunter from submodules , it will test the packaging of the current versions of magnum and corrade.
````bash
corrade and magnum versions (+ the packaging of sdl, glfw, etc)
➜  /sources_docker git:(master) ✗ ./TLDR_hunter.py test-build magnum_example_app_with_submodules gcc-7-cxx14
````

This could be extended to magnum plugins later.

File: package/hunter/HunterInit.cmake (hunk 2 above)
Comment (pthom) [reply package/hunter/HunterInit.cmake]: Yep, I was too fast !

File: CMakeLists.txt (hunk 1 above)
Comment (mosra) [reply CMakeLists.txt]: I see, thanks for the extensive explanation.

**Comment by mosra** [comment]
Re the `FATAL_ERROR` about `find_package(EGL)`: I'm aware of these issues and after reading Hunter docs I'm certain this needs a bigger update on my side. Putting that [on my TODO list](https://github.com/mosra/magnum/projects/6#card-15406032). I think not having an iOS build in the initial release is okay.

Re Ubuntu 14 vs 16: it's about this, right? https://travis-ci.org/pthom/hunter/jobs/462782227#L1290
I managed to [work around this error](https://github.com/mosra/corrade/blob/7714f2dd3c81e88b13aa1d5b3a16b44df99e67f9/src/Corrade/Utility/FileWatcher.h#L85-L90) for GCC builds, but not for Clang builds. This is because std::string move constructor is mistakenly not noexcept in libstdc++ < 5.5 but it's impossible to know what libstdc++ version is actually used. 

I'll be switching to Ubuntu 16 soon anyway, so unless this is a critical issue for Hunter, I'm not sure we should bother too much with it.

**Comment by pthom** [comment]
> Re the FATAL_ERROR about find_package(EGL):  I think not having an iOS build in the initial release is okay.

I think too !

> Re Ubuntu 14 vs 16: it's about this, right? https://travis-ci.org/pthom/hunter/jobs/462782227#L1290
I managed to work around this error for GCC builds, but not for Clang builds. This is because std::string move constructor is mistakenly not noexcept in libstdc++ < 5.5 but it's impossible to know what libstdc++ version is actually used.

Exactly, the error message is strange, given that the corrade code seems to do what is appropriate

> I'll be switching to Ubuntu 16 soon anyway, so unless this is a critical issue for Hunter, I'm not sure we should bother too much with it.

I hope the maintainer of hunter will bare with me for I changed from travis to xenial. Having a package like magnum is probably worth it !

**Comment by mosra** [comment]
I just pushed https://github.com/mosra/corrade/commit/ae1fc17d3d77b1bec248382fbdbefd8b9fbb44a9 which is hopefully a better workaround for the `noexcept` issue and could make build on 14.04 possible again. Not on `master` yet, [waiting for the build to succeed](https://travis-ci.org/mosra/corrade/builds/462876028) but looks good so far. Could you try on the CI with this commit and 14.04?

File: package/hunter/HunterInit.cmake (hunk 3 above)
Comment (pthom) [reply package/hunter/HunterInit.cmake]: > Could this be wrapped in if(NOT EXISTS ${CMAKE_CURRENT_LIST_DIR}/HunterGate.cmake)?

This is done

> The other option I mentioned in #297 was putting the file into the toolchains repository. But that would require the user to explicitly init the toolchains submodule, which is not needed for anything else. So I think this is a better solution 👍

I thought so too! And it was the only solution since hunter uses `git archive`during its build (and we of course lose the submodules this way...)

I pushed a commit with these modifications, and tested it inside a travis like container, like so; so that I am confident about it.

````bash
> cd /path/to/magnum_hunter/Docker/ubuntu
# Inside the docker container 
> /sources_docker >  rm magnum/package/hunter/HunterGate.cmake
> /sources_docker >  ./TLDR_hunter.py test-build magnum gcc-7-cxx14 --clean
> file magnum/package/hunter/HunterGate.cmake
magnum/package/hunter/HunterGate.cmake: ASCII text
````

**Comment by pthom** [comment]
Here is my checklist in order to finish this.

What do you think of it? I will need your input concerning the releases url and sha1 :-)

* [ ] Corrade hunter package
  * [ ] Hunter CI Back to travis ? (requires quite some boilerplate commits). What is your opinion?
  * [ ] Create a release on corrade main repo
  * [ ] Add the release link and it's sha1 to hunter / make it the default and only version
  * [ ] Push PR to hunter (and hope that the hack for corrade-rc will pass)
  * [ ] Proof read the readme inside hunter (https://github.com/pthom/hunter/blob/test.magnum/docs/packages/pkg/corrade.rst)
  * [ ] Proof-read pr.pkg.corrade and pr.corrade branchs
  * [ ] Push PR to hunter building repo (pr.pkg.corrade)
        (and hope that the hack for corrade-rc will pass)
  * [ ] Push PR to hunter main repo (pr.corrade)

* [ ] Magnum hunter package
  * [ ] Hunter CI Back to travis ? (requires quite some boilerplate commits)
  * [ ] Merge magnum_hunter branch
  * [ ] Create a release on magnum main repo
  * [ ] Add the release link and it's sha1 to hunter / make it the default and only version
  * [ ] Proof read the readme inside hunter (https://github.com/pthom/hunter/blob/test.magnum/docs/packages/pkg/magnum.rst)
  * [ ] Proof-read pr.pkg.magnum and pr.magnum branchs
  * [ ] Push PR to hunter building repo (pr.pkg.magnum)
        (and hope that the hack for corrade-rc will pass)
  * [ ] Push PR to hunter main repo (pr.magnum)

* Magnum (optiona steps)
  * [ ] Integrate example_apps into magnum-bootstrap ?
  * [ ] What to do about the magnum_hunter repo itself (which can be useful for future maintainers of the hunter package, or people wanting to add hunter plugins)

**Comment by mosra** [comment]
> Hunter CI Back to travis ?

Not really sure what this means, sorry :sweat_smile: 

> Create a release

I'm planning to have a 2018.12 release, but that will happen later this month (around Dec 30th I think) as I still have quite a few important features I want to finish before tagging another version. Would it be possible to use a temporary tag on your fork in the meantime? Once 2018.12 is out, I can switch the packages to the "upstream" tag.

> Integrate example_apps into magnum-bootstrap ?

Yes, definitely. I can do that once Magnum is available in Hunter, I also wanted to play with it a bit to get a feeling how it's used. For the rest of the `magnum_hunter` I'm not quite sure yet, maybe it could go also to `package/hunter/` to aid with packaging and testing in the future. Would it be okay to just keep it on your account until I have a clearer idea?

**Comment by pthom** [comment]
> Hunter CI Back to travis ?

I need to discuss this directly with Ruslan Baratov (maintainer of hunter)

> Create a release: I'm planning to have a 2018.12 release, but that will happen later this month (around Dec 30th I think) as I still have quite a few important features I want to finish before tagging another version

There is no hurry.

> Would it be possible to use a temporary tag on your fork in the meantime?

Not on my fork. If we want to publish a release before the 2018.12 release, I would need to place my fork inside https://github.com/hunter-packages/ (which I did not).
But if I copy the fork there, it should be ok, and I can create a tag there.

  > For the rest of the magnum_hunter I'm not quite sure yet, maybe it could go also to package/hunter/ to aid with packaging and testing in the future. Would it be okay to just keep it on your account until I have a clearer idea?

Of course, this is the good way to go.

**Comment by pthom** [comment]
One question that is abolutely not related. 
I am quite new to magnum actually. When I build the primitive example on my mac (https://github.com/mosra/magnum-examples/tree/master/src/primitives) , I get a black screen with the following output:
````
Platform::Sdl2Application: warning: the executable is not a HiDPI-enabled app bundle
Renderer: AMD Radeon Pro 560 OpenGL Engine by ATI Technologies Inc.
OpenGL version: 4.1 ATI-2.2.8
Using optional features:
    GL_ARB_ES2_compatibility
    GL_ARB_separate_shader_objects
    GL_ARB_texture_filter_anisotropic
    GL_ARB_texture_storage
    GL_ARB_vertex_array_object
    GL_EXT_debug_label
    GL_EXT_debug_marker
Using driver workarounds:
    no-layout-qualifiers-on-old-glsl
````

Did I miss a setting?

It does work on windows, though.

**Comment by mosra** [comment]
No, you didn't. it's a known bug in macOS 10.14. Shake the window, minimize/maximize it or click into it to make the contents appear. Related info for SDL is here: https://discourse.libsdl.org/t/macos-10-14-mojave-issues/25060 (so far I'm not aware of a new version fixing this :/ ...), more info (and some workarounds) here in a GLFW issue: https://github.com/glfw/glfw/issues/1334

**Comment by pthom** [comment]
> Shake the window, minimize/maximize it or click into it to make the contents appear

This is the funniest workaround I have heard off :-)

It does not work on my side because I cannot resize the example app window. It is not important at all, though: I can work on linux / windows.

**Comment by mosra** [comment]
Yes, it's quite annoying. Not related to Magnum at all, but rather to how the underlying toolkits (SDL/GLFW) manage the GL context. I hope the fixed versions of SDL / GLFW get released soon.

For the release tag: so is it okay to have the release delayed until I get 2018.12 out? I really don't want to be blocking you from getting your work done :)

Oh, but Corrade (2018.10) could get packaged as-is already so we can tick at least some boxes, right? Then, once that's done, the hunter URL here could get updated to an upstream version of Hunter that contains Corrade and then this PR could get merged without any TODOs left. Is that a reasonable path forward?

**Comment by pthom** [comment]
Ok, let's try to tick some boxes for corrade :

Here is the first PR out of two :  https://github.com/ingenue/hunter/pull/300
(this one contains the CI scripts)

File: CMakeLists.txt [hunk 4]
```diff
@@ -49,8 +49,20 @@ endif()
...
+
 project(Magnum ${LANG})
 
+if(HUNTER_ENABLED)
+    include(HunterFetchDeps)
```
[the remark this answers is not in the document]
pthom: It was a request of the author to have the calls to hunter_add_package in  a separate location, so that the packaging info is grouped (magnum handles packaging for conan, homebrew, debian, gentoo, and archlinux and they are nicely grouped inside magnum/package).

My tests showed that it has worked on all platforms : some of them have `WITH_AUDIO`, some do not; same for `WITH_GLFWAPPLICATION` and it works on all combinations.

package/hunter/HunterInit.cmake
```diff
@@ -0,0 +1,14 @@
+if(NOT EXISTS ${CMAKE_CURRENT_LIST_DIR}/HunterGate.cmake)
+  file(
+    DOWNLOAD
+    https://raw.githubusercontent.com/hunter-packages/gate/c4e5b3e4d1c97fb46ae86af7f59baac95f7f1cf0/cmake/HunterGate.cmake
+    ${CMAKE_CURRENT_LIST_DIR}//HunterGate.cmake
```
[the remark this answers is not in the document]
pthom: See https://github.com/mosra/magnum/pull/298/commits/1af4ef026521f763578bbd9bf1970244b10a8d85 : it checks the status, the sha256 and prints nice messages on failure.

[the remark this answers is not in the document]
ruslo: > See 1af4ef0 : it checks the status, the sha256 and prints nice messages on failure

My real point here was that adding HunterGate.cmake to sources is the simpiest and robust solution. For example your current code do not support download retry logic and if somebody will interrupt download procedure with Ctrl+C you will have broken `HunterGate.cmake` and `if(NOT EXISTS ${hunter_gate_file})` logic will not allow to calculate hash of it.

Also note that your approach require to always have connection to network, it's not a requirement for Hunter. Both Hunter archives and packages sources are ok to be on local machine only.

CMakeLists.txt
mosra: @ruslo Yes, I'm aware of that ;) This `if()` was added in order to not require `HunterGate.cmake` to be present in the source directory at all, which I requested.

CMakeLists.txt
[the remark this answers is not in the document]
mosra: @ruslo I never experienced any issues related to the `WITH_*` options being used before the actual `option()`. The options are either present on the command-line `-DWITH_*`, in which case they're already available, or they are present in CMake cache from previous runs. Again, as @pthom says, this was my request to have it in a separate location (previously it was done like you suggest).

package/hunter/HunterInit.cmake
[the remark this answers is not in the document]
mosra: @ruslo yes, I know. Don't want to sound mean here, my request here was simply to avoid having the file in the project. With Magnum I go to extreme lengths to reduce non-essential features and boilerplate and that goes also for the buildsystem.

For me, cases like abort in the middle of a download or broken checksums fall into the broad category of rare "CMake build dir is totally hosed" problems and the usual solution is to wipe the build directory and start over. When Ithe network connection requirement or broken HunterGate files become a real problem for users, I'll revisit this; until then I'm not sure this is worth spending too much time on.

CMakeLists.txt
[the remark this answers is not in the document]
ruslo: > The options are either present on the command-line -DWITH_*, in which case they're already available, or they are present in CMake cache from previous runs.

Okay, I see, that's because all of them set to `OFF` by default. It will not work if you change one of them to `ON`.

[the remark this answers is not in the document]
mosra: Yes, true. All features that have external dependencies (and thus need to be explicitly handled by Hunter) are `OFF` by default as a rule.

pthom:
Hi,
I just pushed some modifs to https://github.com/ingenue/hunter/pull/300#pullrequestreview-180925465

This version should pass, as it disables the hack for corrade-rc.

For the future release of corrade on magnum, we have two options :
* Leave the p.pkg.corrade branch as it is, i.e do not build iOS and Android packages in hunter's CI (I added a note inside the readme that explains it can work if corrade-rc is in the path)
* Add some scripts into the corrade repo : something that will cause corrade-rc to try to build automatically if `find_program(corrade-rc)` fails (this might be based on a cmake flag) . I do not think it is worth adding a dependency to hunter for this, as corrade does not require hunter (it has no dependency), so that a shell script called by cmake could be sufficient. 

Do you have a preference?

mosra:
@pthom thank you!

I would leave it as-is, for now. With other cross-compilation packages I'm usually adding a dependency to native Corrade installation (so e.g. ArchLinux `emscripten-corrade` depends on native `corrade`). Not sure if anything like that can be expressed in Hunter, but again, let's leave it for a later time :)

ruslo:
>  I do not think it is worth adding a dependency to hunter for this, as corrade does not require hunter (**it has no dependency**), so that a shell script called by cmake could be sufficient

**Target** version of package depends on **host** build of same package. So there is a dependency.

> Not sure if anything like that can be expressed in Hunter

Yes, it can be, see comments:
* https://github.com/ingenue/hunter/pull/300#discussion_r238459547
* https://github.com/ruslo/hunter/issues/1635#issuecomment-443826466

mosra:
@ruslo right, thanks, sorry that you had to repeat everything again :sweat_smile: 

Yes, now I think I know how this could be done (as a next step after the native packages are done). Since I still had some problems with e.g. `find_package()` not having `REQUIRED` clauses for Android / iOS and so I need to fix that first.

mosra:
Great!

I'll reference it in the [building documentation](https://doc.magnum.graphics/corrade/building-corrade.html). Is there anything I should mention besides "Corrade is available in Hunter" and linking to https://docs.hunter.sh/en/latest/packages/pkg/corrade.html ?

Can you update the Hunter release URL here? I'll merge this PR right after.

mosra:
Yay, badge! Of course :)

Done in mosra/corrade@a4c4fe6b28295ad8ab8070070d1ef69c0308c7d2, updated docs [here](https://doc.magnum.graphics/corrade/building-corrade.html#building-corrade-packages-hunter).

mosra:
Merged a squashed version in 91b41a729717af299291be4b43fa86b20654a616, I did a bit of simplification and coding style updates on top. 

@pthom once I tag 2018.12, I'll let you know, and in case you are not around, I'll attempt to finish the Magnum package submit myself. Thanks a lot for the contribution! :+1:

PR #299, opened 2018-12-09
Strict weak ordering for math types

iboB:
* Added a new template `StrictWeakOrdering` which defines a strict weak ordering between two values. 
* Added specializations of this struct for math types 
* Added a shorthand usage `StrictWeakOrderingFunc` which allows the use of `StrictWeakOrdering` without explicit template arguments, by providing a template `operator()` instead

### DIFF
--- a/src/Magnum/Math/BoolVector.h
+++ b/src/Magnum/Math/BoolVector.h
@@ -33,6 +33,7 @@
 #include <Corrade/Utility/Debug.h>
 
 #include "Magnum/Types.h"
+#include "Magnum/Math/Math.h"
 #include "Magnum/Math/Tags.h"
 
 namespace Magnum { namespace Math {
@@ -301,6 +302,27 @@ template<std::size_t size> inline BoolVector<size> BoolVector<size>::operator~()
         out._data[i] = ~_data[i];
 
     return out;
+}
+
+/* Specialization of helper types*/
+namespace Implementation {
+template<std::size_t size> struct StrictWeakOrdering<BoolVector<size>> {
+    bool operator()(const BoolVector<size>& a, const BoolVector<size>& b) const {
+        auto ad = a.data();
+        auto bd = b.data();
+        for(std::size_t i = 0; i < BoolVector<size>::DataSize - 1; ++i) {
+            if(ad[i] < bd[i])
+                return true;
+            if(ad[i] > bd[i])
+                return false;
+        }
+
+        // mask last element with to hide unused bits
+        constexpr UnsignedByte mask = UnsignedByte(0xFF) >> (BoolVector<size>::DataSize * 8 - size);
+        constexpr std::size_t i = BoolVector<size>::DataSize - 1;
+        return (ad[i] & mask) < (bd[i] & mask);
+    }
+};
 }
 
 }}
--- a/src/Magnum/Math/CMakeLists.txt
+++ b/src/Magnum/Math/CMakeLists.txt
@@ -48,6 +48,7 @@ set(MagnumMath_HEADERS
     Packing.h
     Range.h
     RectangularMatrix.h
+    StrictWeakOrdering.h
     Swizzle.h
     Tags.h
     Unit.h
--- a/src/Magnum/Math/Color.h
+++ b/src/Magnum/Math/Color.h
@@ -1272,6 +1272,10 @@ namespace Implementation {
     template<class T> struct TypeForSize<3, Color4<T>> { typedef Color3<T> Type; };
     template<class T> struct TypeForSize<4, Color3<T>> { typedef Color4<T> Type; };
     template<class T> struct TypeForSize<4, Color4<T>> { typedef Color4<T> Type; };
+
+    template<class T> struct StrictWeakOrdering<Color3<T>> : public StrictWeakOrdering<Vector<3, T>> {};
+    template<class T> struct StrictWeakOrdering<Color4<T>> : public StrictWeakOrdering<Vector<4, T>> {};
+
 }
 
 }}
--- a/src/Magnum/Math/CubicHermite.h
+++ b/src/Magnum/Math/CubicHermite.h
@@ -538,6 +538,23 @@ template<class T> inline bool CubicHermite<T>::operator==(const CubicHermite<T>&
         TypeTraits<T>::equals(_outTangent, other._outTangent);
 }
 
+namespace Implementation {
+template<class T> struct StrictWeakOrdering<CubicHermite<T>> {
+    bool operator()(const CubicHermite<T>& a, const CubicHermite<T>& b) const {
+        StrictWeakOrdering<T> o;
+        if(o(a.inTangent(), b.inTangent()))
+            return true;
+        if(o(b.inTangent(), a.inTangent()))
+            return false;
+        if(o(a.point(), b.point()))
+            return true;
+        if(o(b.point(), a.point()))
+            return false;
+        return o(a.outTangent(), b.outTangent());
+    }
+};
+}
+
 }}
 
 #endif
--- a/src/Magnum/Math/Dual.h
+++ b/src/Magnum/Math/Dual.h
@@ -395,6 +395,21 @@ template<class T> std::pair<Dual<T>, Dual<T>> sincos(const Dual<Unit<Rad, T>>& a
 template<class T> std::pair<Dual<T>, Dual<T>> sincos(const Dual<Unit<Deg, T>>& angle) { return sincos(Dual<Rad<T>>(angle)); }
 #endif
 
+/* Specialization of helper types*/
+namespace Implementation {
+template<class T> struct StrictWeakOrdering<Dual<T>> {
+    bool operator()(const Dual<T>& a, const Dual<T>& b) const {
+        StrictWeakOrdering<T> o;
+        if(o(a.real(), b.real()))
+            return true;
+        if(o(b.real(), a.real()))
+            return false;
+
+        return o(a.dual(), b.dual());
+    }
+};
+}
+
 }}
 
 #endif
--- a/src/Magnum/Math/Quaternion.h
+++ b/src/Magnum/Math/Quaternion.h
@@ -739,6 +739,21 @@ template<class T> inline Vector3<T> Quaternion<T>::transformVectorNormalized(con
     return vector + _scalar*t + Math::cross(_vector, t);
 }
 
+/* Specialization of helper types*/
+namespace Implementation {
+template<class T> struct StrictWeakOrdering<Quaternion<T>> {
+    bool operator()(const Quaternion<T>& a, const Quaternion<T>& b) const {
+        StrictWeakOrdering<Vector3<T>> o;
+        if(o(a.vector(), b.vector()))
+            return true;
+        if(o(b.vector(), a.vector()))
+            return false;
+
+        return a.scalar() < b.scalar();
+    }
+};
+}
+
 }}
 
 namespace Corrade { namespace Utility {
--- a/src/Magnum/Math/StrictWeakOrdering.h
+++ b/src/Magnum/Math/StrictWeakOrdering.h
@@ -1,0 +1,78 @@
+﻿#ifndef Magnum_Math_StrictWeakOrdering_h
+#define Magnum_Math_StrictWeakOrdering_h
+/*
+    This file is part of Magnum.
+
+    Copyright © 2010, 2011, 2012, 2013, 2014, 2015, 2016, 2017, 2018
+              Vladimír Vondruš <mosra@centrum.cz>
+
+    Permission is hereby granted, free of charge, to any person obtaining a
+    copy of this software and associated documentation files (the "Software"),
+    to deal in the Software without restriction, including without limitation
+    the rights to use, copy, modify, merge, publish, distribute, sublicense,
+    and/or sell copies of the Software, and to permit persons to whom the
+    Software is furnished to do so, subject to the following conditions:
+
+    The above copyright notice and this permission notice shall be included
+    in all copies or substantial portions of the Software.
+
+    THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+    IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+    FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+    THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+    LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+    FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+    DEALINGS IN THE SOFTWARE.
+*/
+
+/** @file
+ * @brief Class @ref Magnum::Math::StrictWeakOrdering
+ */
+
+namespace Magnum { namespace Math {
+
+namespace Implementation {
+template <class T> struct StrictWeakOrdering { // Specialized for supported types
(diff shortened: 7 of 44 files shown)


src/Magnum/Math/Color.h
```diff
@@ -1274,6 +1274,15 @@ namespace Implementation {
     template<class T> struct TypeForSize<4, Color4<T>> { typedef Color4<T> Type; };
 }
 
+/* Specialization of helper types*/
+#if !defined(DOXYGEN_GENERATING_OUTPUT)
+template<class T>
+struct StrictWeakOrdering<Color3<T>> : public StrictWeakOrdering<Vector<3, T>> {};
```
mosra: Coding style: no space in front of `:`.

src/Magnum/Math/StrictWeakOrdering.h
```diff
@@ -0,0 +1,93 @@
...
+
+/** @file
+ * @brief Class @ref Magnum::Math::StrictWeakOrdering
+ */
+
+namespace Magnum { namespace Math {
+/**
+@brief A functor which implements strict weak ordering for types in the math
+library.
+@param T the type for which to have strict weak ordering.
```
mosra: Needs to be `@tparam T`.

src/Magnum/Math/StrictWeakOrdering.h
```diff
@@ -0,0 +1,93 @@
...
+*/
+
+/** @file
+ * @brief Class @ref Magnum::Math::StrictWeakOrdering
+ */
+
+namespace Magnum { namespace Math {
+/**
+@brief A functor which implements strict weak ordering for types in the math
+library.
```
mosra: I encountered various problems with `@brief` spanning over multiple lines in the past. It has to be all on one line (even if the line would be 500 chars long).

src/Magnum/Math/StrictWeakOrdering.h
```diff
@@ -0,0 +1,93 @@
...
+@param T the type for which to have strict weak ordering.
+
+A strict weak ordering enables a comparison operator between to elements of a
+set. With it an ordering can be achieved between elements.
+
+This is useful mainly for interoperability with ordering containers from the
+C++ standard library, like std::map and std::set. Instead of using this
```
mosra: We have a tag file for STL, so this can be `@ref std::map` and `@ref std::set`.

src/Magnum/Math/StrictWeakOrdering.h
```diff
@@ -0,0 +1,93 @@
...
+@brief A shorthand interface for @ref StrictWeakOrdering
+
+Allows the use of @ref StrictWeakOrdering in a shorter and more succinct way.
+To use for std::map or std::set, just add it as their compare template
+argument.
+
+@code{.cpp}
+std::set<Magnum::Vector2, Magnum::Math::StrictWeakOrderingFunc> mySet;
+std::map<Magnum::Vector4, int, Magnum::Math::StrictWeakOrderingFunc> myMap;
+@endcode
```
mosra: I'm compiling all code snippets to ensure they don't get out of sync with future updates of the API. I'm not sure the Developers Guide mentions this, so:

- there's a file in `doc/snippets/MagnumMath.cpp` with a *huge* `main()` function inside
- add a scoped block somewhere to it and denoted by `/* [StrictWeakOrderingFunc] */` (the snippets are roughly alphabetically ordered), include `<map>` / `<set>` if needed and put the contents of the above `@code` inside
- reference the snippet using `@snippet MagnumMath.cpp StrictWeakOrderingFunc`
- the snippets are compile as part of `BUILD_TESTS`, so the CI might fail in case there's an error (or missing include etc.)

I can do the above myself post-merge if it's too much of a hassle for you.

Also, please omit the `Magnum::` prefix in all code snippets to avoid redundancy.

src/Magnum/Math/Test/ColorTest.cpp
```diff
@@ -850,6 +854,30 @@ void ColorTest::xyY() {
...
+    auto r = Color4::red();
+    auto g = Color4::green();
+    auto b = Color4::blue();
+
+    CORRADE_VERIFY(o(b, r));
+    CORRADE_VERIFY(o(g, r));
+    CORRADE_VERIFY(o(b, g));
+
+    auto ba = b;
+    ba.a() = 0.5;
```
mosra: `0.5f`, since it's a `float` literal. I *think* some compilers are emitting a warning here, but not sure which.

src/Magnum/Math/Test/DualComplexTest.cpp
```diff
@@ -445,6 +449,24 @@ void DualComplexTest::transformPoint() {
     CORRADE_COMPARE(transformedB, Vector2(-2.918512f, 2.780698f));
 }
 
+void DualComplexTest::strictWeakOrdering()
+{
+    StrictWeakOrderingFunc o;
+    const DualComplex a{{1, 0}, {1, 3}};
+    const DualComplex b{{1, 2}, {3, 4}};
+    const DualComplex c{{1, 0}, {1, 4}};
+    const DualComplex d{{1, 0}, {1, 3}};
```
mosra: For type clarity and consistency I'm adding explicit `.0f` to all literals in constructors of floating-point types, even though it's not strictly needed. Similarly elsewhere.

Also, probably just slipped through no newline before a `{` ;)

src/Magnum/Math/Test/StrictWeakOrderingTest.cpp
```diff
@@ -0,0 +1,178 @@
...
+    std::map<Vector2, int> m;
+
+    m[Vector2{ 1, 2 }] = 23;
+    m[Vector2{ 4, 5 }] = 55;
+
+    CORRADE_VERIFY(m.size() == 2);
+    CORRADE_VERIFY(m.begin()->second == 23);
+    CORRADE_VERIFY(m.rbegin()->second == 55);
+
+    m[Vector2{ 1, 2 }] = 99;
```
mosra: No spaces inside `{ }`, similarly in `mapShort` above.

src/Magnum/Math/Test/Vector4Test.cpp
```diff
@@ -293,6 +297,24 @@ void Vector4Test::planeEquationNormalPoint() {
...
+
+    CORRADE_VERIFY( ov4(v4a, v4b));
+    CORRADE_VERIFY(!ov4(v4b, v4a));
+    CORRADE_VERIFY( ov4(v4a, v4c));
+    CORRADE_VERIFY(!ov4(v4c, v4a));
+    CORRADE_VERIFY( ov4(v4c, v4b));
+    CORRADE_VERIFY(!ov4(v4b, v4c));
+
+    CORRADE_VERIFY(!ov4(v4a, v4d));
+    CORRADE_VERIFY(!ov4(v4d, v4a));
```
mosra: Thinking out loud: I was looking if you are testing also comparison with the item itself and took me a while until I realized `v4a` and `v4d` are the same. What about removing `v4d` and using `CORRADE_VERIFY(!ov4(v4a, v4a));` instead? (and similarly elsewhere?)

Yes, I know, this change would weaken the test in case the `StrictWeakOrdering` implementation would have a special case for `&a == &b`. But it's not like that, so I think it's safe.

src/Magnum/Math/StrictWeakOrdering.h
```diff
@@ -0,0 +1,93 @@
...
+ */
+struct StrictWeakOrderingFunc {
```
mosra: An idea regarding the naming: since `StrictWeakOrderingFunc` will be used 99% of the time and people wouldn't even need to know that `StrictWeakOrdering` exists, what about reducing the API surface area a bit:

- move `StrictWeakOrdering` struct into the `Math::Implementation` namespace (it's hidden from Doxygen, so then you also don't need to have the `#ifndef`s around each)
- rename `StrictWeakOrderingFunc` to `StrictWeakOrdering` and merge documentation of these two, so instead of:

    ```
    If you want to enable strict weak ordering through this or the
    @ref StrictWeakOrderingFunc shorthand, you must specialize this struct for your
    own type.
    ```

    just mention that the this is implemented for all `Math` types and handwave the implementation by mentioning the struct specialization in `Math::Implementation` and advising the users to check source for details (... but I just realized it's not done for `Bezier` and `CubicHermite`, can you add a specialization for these as well?)
- the tests could keep using the `StrictWeakOrdering` (of course without the explicit template param after) to verify the shorthand also does the job correctly

iboB: Sure. I'll do it tonight. 

I hadn't actually realized that Bezier and CubicHermite were types. I thought they were just files with functions for splines. 😄

src/Magnum/Math/BoolVector.h
```diff
@@ -34,6 +34,7 @@
 
 #include "Magnum/Types.h"
 #include "Magnum/Math/Tags.h"
+#include "Magnum/Math/StrictWeakOrdering.h"
```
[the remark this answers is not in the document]
iboB: check

src/Magnum/Math/CMakeLists.txt
```diff
@@ -48,6 +48,7 @@ set(MagnumMath_HEADERS
     Packing.h
     Range.h
     RectangularMatrix.h
+	StrictWeakOrdering.h
```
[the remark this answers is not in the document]
iboB: Oops. My editor's settings were messed up. Thanks for noticing :)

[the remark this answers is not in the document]
mosra: Does your editor support [.editorconfig](https://editorconfig.org/)? Most of them do. Not sure why it was not here yet, so I just added it in b7e6ed4e84d3e9d1cf76a1fb066d205962283753.

src/Magnum/Math/StrictWeakOrdering.h
mosra: Oh and `Range` as well. Then there's `Half`, which is a completely opaque type for 16-bit floats (and thus no `operator<`). Not sure about that one.

The `Deg`, `Rad` types (derived from `Unit`) have `operator<` implemented so nothing needed I think.

iboB: Ah, so `Half` doesn't have an `operator<`. I'll create an specialization for it as well then. It makes sense to have a way to order them.

src/Magnum/Math/Dual.h
```diff
@@ -395,6 +396,22 @@ template<class T> std::pair<Dual<T>, Dual<T>> sincos(const Dual<Unit<Rad, T>>& a
 template<class T> std::pair<Dual<T>, Dual<T>> sincos(const Dual<Unit<Deg, T>>& angle) { return sincos(Dual<Rad<T>>(angle)); }
 #endif
 
+/* Specialization of helper types*/
+#if !defined(DOXYGEN_GENERATING_OUTPUT)
+template<typename T>
+struct StrictWeakOrdering<Dual<T>> {
+    bool operator()(const Dual<T>& a, const Dual<T>& b) const {
+        StrictWeakOrdering<T> o;
```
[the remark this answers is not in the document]
iboB: I had this one too, but `operator()` can't be `constexpr` in C++11 because it has a for loop inside.

src/Magnum/Math/BoolVector.h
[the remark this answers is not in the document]
iboB: Should I still do it, when the template `StrictWeakOrdering` is in the `namespace Implementation`?

src/Magnum/Math/Dual.h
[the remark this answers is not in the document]
mosra: Ah, true, forgot that `T` can be `Quaternion` as well.

src/Magnum/Math/BoolVector.h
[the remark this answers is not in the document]
mosra: Even in the namespace it still avoids including `StrictWeakOrdering.h`, so yes.

src/Magnum/Math/BoolVector.h
```diff
@@ -303,6 +304,25 @@ template<std::size_t size> inline BoolVector<size> BoolVector<size>::operator~()
...
+        auto bd = b.data();
+        for (std::size_t i = 0; i < BoolVector<size>::DataSize; ++i) {
+            if (ad[i] < bd[i])
+                return true;
+            else if (ad[i] > bd[i])
+                return false;
+        }
+
+        return false;
+    }
```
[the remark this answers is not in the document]
mosra: Ugh, the mask I suggested is not correct, sorry. Correct version should be, I *think* (please correct me if I'm wrong again):

```cpp
const UnsignedByte mask = (1 << (size % 8)) - 1;
```

This error also didn't cause any test to fail, which points to the importance of extending the test to verify random padding values (as I mentioned above). Can you extend it? Thank you!

To make this slightly shorter, I think it could get simplified to a single loop, it loses `constexpr` but this code usually isn't in the critical path anyway:

```cpp
        for(std::size_t i = 0; i < BoolVector<size>::DataSize; ++i) {
            const UnsignedByte mask = (1 << (i + 1 < BoolVector<size>::DataSize ? 8 : size % 8)) - 1;
            if((ad[i] & mask) < (bd[i] & mask))
                return true;
            if((ad[i] & mask) > (bd[i] & mask))
                return false;
        }
```

src/Magnum/Math/Quaternion.h
```diff
@@ -739,6 +739,21 @@ template<class T> inline Vector3<T> Quaternion<T>::transformVectorNormalized(con
...
 
+/* Specialization of helper types*/
+namespace Implementation {
+template<class T> struct StrictWeakOrdering<Quaternion<T>> {
+    bool operator()(const Quaternion<T>& a, const Quaternion<T>& b) const {
+        StrictWeakOrdering<Vector3<T>> o;
+        if(o(a.vector(), b.vector()))
+            return true;
+        if(o(b.vector(), a.vector()))
+            return false;
```
mosra: Codecov reports these two `return` statements being never reached. Extending the test a bit, maybe?

src/Magnum/Math/CubicHermite.h
```diff
@@ -538,6 +538,23 @@ template<class T> inline bool CubicHermite<T>::operator==(const CubicHermite<T>&
...
+    bool operator()(const CubicHermite<T>& a, const CubicHermite<T>& b) const {
+        StrictWeakOrdering<T> o;
+        if(o(a.inTangent(), b.inTangent()))
+            return true;
+        if(o(b.inTangent(), a.inTangent()))
+            return false;
+        if(o(a.point(), b.point()))
+            return true;
+        if(o(b.point(), a.point()))
+            return false;
```
mosra: Codecov reports the two `point()`-related `return` statements being never reached. Can you extend the test? Thanks!

src/Magnum/Math/BoolVector.h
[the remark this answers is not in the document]
iboB: oops :scream:  You're right
I guess I was sleeping last night

src/Magnum/Math/Quaternion.h
iboB: good catch. The problem was I wasn't even using quaternions in the test :laughing:

src/Magnum/Math/BoolVector.h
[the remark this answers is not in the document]
iboB: Fixed with my own calculation: `0xFF >> (BoolVector<size>::DataSize * 8 - size)`

`BoolVector<size>::DataSize * 8 - size` is the padding, so with this we just pad an all-ones-in-binary accordingly

iboB:
Squashed all commits

mosra:
Merged in f7f2ab8cb5b5590c455aafbb53a4458d1d76aeb6, I cherry-picked the commit on current master instead of doing a merge commit. Hope that's okay :)

You probably already saw it all, so just for the record I did some minor formatting updates and deduplicated the test code a bit in the following commits. Updated docs are uploaded now [as well](https://doc.magnum.graphics/magnum/structMagnum_1_1Math_1_1StrictWeakOrdering.html).

Thanks a lot for the contribution, looking forward for more! :blush:

PR #300, opened 2018-12-15
Emscripten application

Squareys:
Hi @mosra !

This is the pullrequest for the light-weight emscripten HTML5 API based application hackathon project.

Cheers, Jonathan

**TODO**
 - [x] Adapt config to Sdl2Application (e.g. have sampleCount)
 - [x] mouse input
 - [x] keyboard input
 - [x] viewport events
 - [x] text input
 - ~[ ] fullscreen handling~ => `emscripten_(request|exit)_fullscreen()` on user gestures as Application does not usually handle this.
 - [x] bootstrap project
 - [x] Documentation
 - [x] Port some Applications
     - [x] [magnum-examples/master](https://github.com/mosra/magnum-examples/) => https://github.com/mosra/magnum-examples/pull/60
         - [x] webvr
     - [x] [magnum-examples/ports](https://github.com/mosra/magnum-examples/tree/ports/src) => https://github.com/mosra/magnum-examples/pull/59
         - [x] triangle
         - [x] primitives
         - [x] textured-triangle
         - [x] arealights
         - [x] audio
         - [x] box2d
         - [x] bullet
         - [x] mouseinteraction
         - [x] picking
         - [x] text
         - [x] imgui
     - [x] Bootstrap application => https://github.com/mosra/magnum-bootstrap/pull/17
     - [x] Magnum Extras (gallery, player) => https://github.com/mosra/magnum-extras/pull/10


src/Magnum/Platform/EmscriptenApplication.cpp
```diff
@@ -89,10 +105,18 @@ bool EmscriptenApplication::tryCreate(const Configuration& configuration, const
         (glConfiguration.flags() & GLConfiguration::Flag::FailIfMajorPerformanceCaveat) != GLConfiguration::Flags{};
     attrs.enableExtensionsByDefault =
         (glConfiguration.flags() & GLConfiguration::Flag::EnableExtensionsByDefault) != GLConfiguration::Flags{};
+#if FALSE /* "EMSCRIPTEN_VERSION > 1.38.12 */ && OFFSCREEN_FRAMEBUFFER
+    // TODO! Good thing my Emscripten version is from the bronze era
+    // We probably never want to use this anyway, except when we have a
+    // WindowlessEmscriptenApplication
```
mosra: We have `WindowlessEglApplication`, all in-browser GL tests use it. No need to have headless rendering here :)

src/Magnum/Platform/EmscriptenApplication.cpp
```diff
@@ -0,0 +1,454 @@
...
+        };
+    emscripten_set_mousemove_callback("#canvas", this, false, mouseMoveCallback);
+
+    // TODO: Trying to map touch to mouse opened quite a can of worms:
+    //  1. Tap is already mapped via click by emscripten => mouse press, mouse move, mouse release
+    //  2. Touch down, move, up is not a tap
+    //  3. If single touch, everything is clear, second touch-down could be handled with canceling, but
+    //     would require a manual mouse release event. Or we could continue just following the first touch?
+    //
+    // Map first touch as left mouse, yes or no?
```
Squareys: @mosra This is up for discussion :)

src/Magnum/Platform/EmscriptenApplication.cpp
```diff
@@ -0,0 +1,454 @@
...
+EmscriptenApplication::KeyEvent::Key EmscriptenApplication::KeyEvent::toKey(const EM_UTF8* key, const EM_UTF8* code) {
+    // FIXME (Discuss with @mosra)
+    //
+    // This code does not seem very pretty
+    //  1. It's pretty similar to linear search through an array
+    //  2. It's branchy and hard to read
+    //  3. Given the event.key() == ? pattern, this would make sense to do _once_
+    //     in the constructor of KeyEvent rather than every key() call
+    //
+    //  Maybe putting all in a std::unordered_map is better?
```
Squareys: @mosra This is also up for discussion :)

Squareys:
@mosra Is it sensible to really port all of the examples? Should I keep a macro platform switch to demonstrate that Sdl2Application is compatibly replaceable with EmscriptenApplication?

mosra:
> Should I keep a macro platform switch to demonstrate that Sdl2Application is compatibly replaceable with EmscriptenApplication?

The EmscriptenApplication should be the sensible default for Emscripten, but Sdl2Application is still the sensible default basically everywhere else. So, yes there should be a platform switch between EmscriptenApplication, Sdl2Application (and AndroidApplication) because the ports branch still needs to compile on all platforms, but there's no need for having an option to use SDL2 on Emscripten.

Squareys:
@mosra This is ready for review! If there is any bigger tasks I completely overlooked, send those over first so I can work on them in parallel to the rest of the review 🙂

src/Magnum/Platform/EmscriptenApplication.cpp
```diff
@@ -0,0 +1,505 @@
...
+        [](int, const EmscriptenKeyboardEvent* event, void* userData) -> Int {
+            KeyEvent e{event};
+            reinterpret_cast<EmscriptenApplication*>(userData)->keyReleaseEvent(e);
+            return e.isAccepted();
+        };
+    emscripten_set_keyup_callback("#canvas", this, false, keyReleaseCallback);
+}
+
+EmscriptenApplication::KeyEvent::Key EmscriptenApplication::KeyEvent::toKey(const EM_UTF8* key, const EM_UTF8* code) {
+    const size_t keyLength = strlen(key);
```
mosra: `std::strlen()`, `std::strcmp()`, ... similarly elsewhere

src/Magnum/Platform/EmscriptenApplication.cpp
```diff
@@ -0,0 +1,505 @@
...
+       representation of the key, i.e. we get `z` for z depending on layout
+       where code may give us `y` independent of the layout. */
+    if(keyLength == 1) {
+        if(key[0] >= 'a' && key[0] <= 'z') return Key(key[0]);
+        else if(key[0] >= 'A' && key[0] <= 'Z') return Key(key[0] + 32);
+    }
+
+    /* We use code for 0-9 as it allows us to differentiate towards Numpad digits.
+       For digits independent of numpad or not, key is e.g. '0' for Zero */
+    if(Utility::String::beginsWith(code, "Digit")) {
```
mosra: Unfortunately the `beginsWith()` etc. functions take a `std::string` as a first parameter, which means it gets allocated again on every call. I need to introduce a StringView class first to fix that, until then can you use `viewBeginsWith()` everywhere instead?

src/Magnum/Platform/EmscriptenApplication.cpp
```diff
@@ -0,0 +1,505 @@
...
+
+void EmscriptenApplication::startTextInput() {
+    _flags |= Flag::TextInputActive;
+}
+
+void EmscriptenApplication::stopTextInput() {
+    _flags &= ~Flag::TextInputActive;
+}
+
+void EmscriptenApplication::setTextInputRect(const Range2Di& CORRADE_UNUSED(rect)) {
```
mosra: why not just `setTextInputRect(const Range2Di&)`? :)

modules/FindMagnum.cmake
```diff
@@ -410,6 +412,10 @@ endif()
 
 set(_MAGNUM_Trade_DEPENDENCIES )
 set(_MAGNUM_AndroidApplication_DEPENDENCIES GL)
+set(_MAGNUM_EmscriptenApplication_DEPENDENCIES GL)
```
mosra: Ummmmmm I guess you wanted to say

```suggestion
set(_MAGNUM_EmscriptenApplication_DEPENDENCIES )
```

instead, right?

src/Magnum/Platform/EmscriptenApplication.cpp
mosra: I need to introduce proper touch events anyway, trying to shovel them into mouse events is not a good idea. From those, Sdl2App has the pinch event, but (just checked) that one doesn't even work on Emscripten anyway. And as far as I see, none of the examples need the touch events now, only the mouse emulation for tap, right?

src/Magnum/Platform/EmscriptenApplication.cpp
```diff
@@ -0,0 +1,505 @@
...
+            if(app->isTextInputActive() && strlen(event->key) == 1) {
+                TextInputEvent e{{event->key, 1}};
+                app->textInputEvent(e);
+                return e.isAccepted();
+            }
+            KeyEvent e{event};
+            app->keyPressEvent(e);
+            return e.isAccepted();
+        };
+    emscripten_set_keydown_callback("#canvas", this, false, keyDownCallback);
```
mosra: what about doing directly

```suggestion
    emscripten_set_keydown_callback("#canvas", this, false, [](int, const EmscriptenKeyboardEvent* ...) {
```

instead? Similarly elsewhere.

src/Magnum/Platform/EmscriptenApplication.cpp
```diff
@@ -0,0 +1,505 @@
...
+EmscriptenApplication::KeyEvent::Key EmscriptenApplication::KeyEvent::toKey(const EM_UTF8* key, const EM_UTF8* code) {
+    const size_t keyLength = strlen(key);
+    if(keyLength == 0) return Key::Unknown;
+
+    /* We use key for a-z as it gives us a keyboard layout respecting
+       representation of the key, i.e. we get `z` for z depending on layout
+       where code may give us `y` independent of the layout. */
+    if(keyLength == 1) {
+        if(key[0] >= 'a' && key[0] <= 'z') return Key(key[0]);
+        else if(key[0] >= 'A' && key[0] <= 'Z') return Key(key[0] + 32);
```
mosra: `- 'A' + 'a'` instead of 32? :)

src/Magnum/Platform/EmscriptenApplication.h
```diff
@@ -0,0 +1,1509 @@
...
+    DEALINGS IN THE SOFTWARE.
+*/
+
+#if defined(CORRADE_TARGET_EMSCRIPTEN) || defined(DOXYGEN_GENERATING_OUTPUT)
+/** @file
+ * @brief Class @ref Magnum::Platform::EmscriptenApplication, macro @ref MAGNUM_EMSCRIPTENAPPLICATION_MAIN()
+ */
+#endif
+
+#include <memory>
```
mosra: Uh oh ... a leftover from the December hackaton? :sweat_smile:

src/Magnum/Platform/EmscriptenApplication.h
```diff
@@ -0,0 +1,1509 @@
...
+# ...
+if(CORRADE_TARGET_EMSCRIPTEN)
+    target_link_libraries(your-app Magnum::EmscriptenApplication)
+endif()
+@endcode
+
+If no other application is requested, you can also use the generic
+`Magnum::Application` alias to simplify porting. Again, see @ref building and
+@ref cmake for more information. Note that unlike on other platforms you need
+to create *shared library* instead of executable.
```
mosra: Ummm ... a copypaste leftover from `AndroidApplication`? :sweat_smile:  Similarly for the JNI blurb below.

src/Magnum/Platform/EmscriptenApplication.h
```diff
@@ -0,0 +1,1509 @@
...
+        /**
+         * @brief Construct with default configuration
+         *
+         * Equivalent to calling @ref EmscriptenApplication(const Arguments&, const Configuration&)
+         * with default-constructed @ref Configuration.
+         */
+        explicit EmscriptenApplication(const Arguments& arguments);
+
+        /**
+         * @brief Construct without creating a window
```
mosra: Can you replace all occurences of "window" with "setting up a canvas"? It's misleading. And while at it, WebGL instead of OpenGL.

src/Magnum/Platform/EmscriptenApplication.h
```diff
@@ -0,0 +1,1509 @@
...
+         * @see @ref startTextInput(), @ref isTextInputActive(), @ref textInputEvent()
+         *      @ref textEditingEvent()
+         */
+        void stopTextInput();
+
+        /**
+         * @brief Set text input rectangle
+         *
+         * The @p rect defines an area where the text is being displayed, for
+         * example to hint the system where to place on-screen keyboard.
```
mosra: Can you put a "Currently not implemented, included only for compatibility with other Application implementations" here?

src/Magnum/Platform/EmscriptenApplication.cpp
```diff
@@ -0,0 +1,505 @@
...
+    emscripten_set_main_loop_arg([](void* arg) {
+        static_cast<EmscriptenApplication*>(arg)->mainLoopIteration();
+    }, this, 0, true);
+
+    return _exitCode;
+}
+
+void EmscriptenApplication::exit(int exitCode) {
+    emscripten_cancel_main_loop();
+    _exitCode = exitCode;
```
mosra: Does an exit code make any sense in the context of an in-browser app? I don't think so. The `exitCode` parameter would then be included only for compatibility with other app implementations (and mentioned as such in the docs).

src/Magnum/Platform/EmscriptenApplication.h
```diff
@@ -0,0 +1,1509 @@
...
+            ProxyContextToMainThread,
+        };
+
+        /**
+         * @brief Context flags
+         *
+         * @see @ref setFlags(), @ref Context::Flags
+         * @requires_gles Context flags are not available in WebGL.
+         */
+        //typedef Containers::EnumSet<Flag, Flag::PremultipliedAlpha> Flags;
```
mosra: A forgotten leftover line?

src/Magnum/Platform/EmscriptenApplication.h
```diff
@@ -0,0 +1,1509 @@
...
+
+        /**
+         * @brief Set context flags
+         * @return Reference to self (for method chaining)
+         *
+         * Default is @ref Flag::ForwardCompatible on desktop GL and no flags
+         * on OpenGL ES.
+         * @see @ref addFlags(), @ref clearFlags(), @ref GL::Context::flags()
+         * @requires_gles Context flags are not available in WebGL.
+         */
```
mosra: This doc block is ... not making any sense ;) Similarly the `@requires_gles` part in all related functions below.

src/Magnum/Platform/EmscriptenApplication.h
```diff
@@ -0,0 +1,1509 @@
...
+
+        /** @brief Moving is not allowed */
+        ViewportEvent& operator=(ViewportEvent&&) = delete;
+
+        /**
+         * @brief Window size
+         *
+         * On some platforms with HiDPI displays, window size can be different
+         * from @ref framebufferSize(). See @ref Platform-Sdl2Application-dpi
+         * for more information.
```
mosra: Not the case on the web. Should be just `Equivalent to @ref framebufferSize().`, similarly below.

src/Magnum/Platform/EmscriptenApplication.h
```diff
@@ -0,0 +1,1509 @@
...
+        const Containers::ArrayView<const char> _text;
+        bool _accepted;
+};
+
+/**
+@brief Text editing event
+
+@see @ref textEditingEvent()
+*/
+class EmscriptenApplication::TextEditingEvent {
```
mosra: Since the EmscriptenApp doesn't implement any of this, I think this shouldn't be here at all. The same is with GLFW, it has no `textEditingEvent()` at all.

src/Magnum/Platform/Test/EmscriptenApplicationTest.cpp
```diff
@@ -0,0 +1,215 @@
...
+
+#include "Magnum/Platform/EmscriptenApplication.h"
+#include "Magnum/GL/Renderer.h"
+#include "Magnum/GL/DefaultFramebuffer.h"
+#include "Magnum/Math/Color.h"
+
+namespace Magnum { namespace Platform { namespace Test {
+
+
+Debug& operator<<(Debug& debug, const Platform::Application::KeyEvent::Key value) {
```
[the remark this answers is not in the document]
Squareys: Was not, I wanted to check if I was mapping to the Enum correctly, and actually found a few bugs with that, but no longer needed and just a hassle to keep up to date.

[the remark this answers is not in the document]
Squareys: Hmm, I guess it probably would have been after all 🤔

modules/FindMagnum.cmake
Squareys: Yeah 😅

src/Magnum/Platform/EmscriptenApplication.h
Squareys: Okay, I tried to clarify the docs in this way and added a note to `windowSize` to explain that it is named `windowSize` because of API compat with other Platforms, but retrieves the canvas size.

src/Magnum/Platform/EmscriptenApplication.h
Squareys: Is the mention of it being different on other platforms still be worth keeping?

src/Magnum/Platform/EmscriptenApplication.cpp
Squareys: The mouse interaction example currently does not work on my phone. Clicking does, but up/down and drag do not.

Squareys: Aparently with SDL, the touch drag is supported, but down/up are also not. For both the mouse-interaction and the arealights this leads to suboptimal behaviour in the showcase 😅

mosra: If that's just phones, then it's okay I would say. If I remember correctly, even in the original Sdl2App the mouse event has some bug in the coordinate calculation so it's not possible to click on buttons etc.

src/Magnum/Platform/EmscriptenApplication.h
mosra: No. This could be misleading people to think that EmscriptenApplication works differently on non-Emscripten platforms, which makes no sense.

package/ci/travis-emscripten.sh
```diff
@@ -65,8 +65,7 @@ cmake .. \
     -DWITH_GL_INFO=ON \
     -DWITH_AL_INFO=ON \
     -DBUILD_TESTS=ON \
-    -DBUILD_AL_TESTS=ON \
```
Squareys: Uh... yeah, uhm... no.

(TODO with next rebase)

Squareys:
@mosra This is also ready for the next review pass, then I'll squash the remaining commits down into one.

src/Magnum/Platform/EmscriptenApplication.h
```diff
@@ -0,0 +1,1441 @@
...
+        Vector2 _dpiScaling;
+        Vector2i _lastKnownCanvasSize;
+
+        Flags _flags;
+
+        #ifdef MAGNUM_TARGET_GL
+        Containers::Pointer<Platform::GLContext> _context;
+        #endif
+
+        int _exitCode;
```
mosra: This is used nowhere, right? Remove :)

src/Magnum/Platform/EmscriptenApplication.cpp
```diff
@@ -0,0 +1,474 @@
...
+void EmscriptenApplication::textInputEvent(TextInputEvent&) {}
+
+EmscriptenApplication::GLConfiguration::GLConfiguration():
+    _colorBufferSize{8, 8, 8, 0}, _depthBufferSize{24}, _stencilBufferSize{0} {}
+
+void EmscriptenApplication::mainLoopIteration() {
+    /* The resize event is not fired on window resize, so poll for the canvas
+       size here. But only if the window was requested to be resizable, to
+       avoid resizing the canvas when the user doesn't want that. Related
+       issue: https://github.com/kripken/emscripten/issues/1731 */
```
mosra: This AFAIK holds for `emscripten_set_resize_callback()` as well, right? Maybe update the comment to make it clear it's not just a copypasted assumption from Sdl2App.

src/Magnum/Platform/EmscriptenApplication.cpp
```diff
@@ -0,0 +1,474 @@
...
+    };
+
+    #pragma GCC diagnostic push
+    #pragma GCC diagnostic ignored "-Wmissing-prototypes"
+    /* Returns the id of the element that is used for listening to key events
+       or '#document' if none is configured in the Module */
+    EM_JS(const char*, getKeyboardListeningElement, (), {
+        return Module['keyboardListeningElement'] || '#document';
+    });
+    #pragma GCC diagnostic pop
```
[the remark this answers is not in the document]
Squareys: > wasm symbol export and 

There is no symbol exported from wasm, though...

As far as I know, this generates basically the same JavaScript/WASM code, but with a temporary unnamed function. I therefore would very much prefer the function as you can read the code in a "default-C++-reading mode" until you actually look up the implementation of `getKeyboardListeningElement`. Your version is pretty hard to read, but I fear that mine returns memory ownership and `const char*` is therefore not correct. OTOH... `intArrayFromString` sounds like it will have 3 bytes inbetween the chars...?

"inlining JS" in this way only inlines in the C++ code, not in the binary.

I will compile both and report back when I verified that this is the case and know what's going on with those allocations :)

src/Magnum/Platform/EmscriptenApplication.cpp
Squareys: Good question, I think it may actually fire even.

src/Magnum/Platform/EmscriptenApplication.cpp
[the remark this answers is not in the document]
Squareys: Summarizing follow-up discussion on gitter for later reference:

`EM_ASM` produces:

~~~js
var ASM_CONSTS = [function() { var element = Module['keyboardListeningElement'] || '#document'; return allocate(intArrayFromString(element), 'i8', ALLOC_NORMAL); },
 function($0) { var name = UTF8ToString($0); return typeof process !== 'undefined' && name in process.env ? allocate(intArrayFromString(process.env[name]), 'i8', ALLOC_NORMAL) : 0; }];

function _emscripten_asm_const_i(code) {
  return ASM_CONSTS[code]();
}

function _emscripten_asm_const_ii(code, a0) {
  return ASM_CONSTS[code](a0);
}
~~~

`EM_JS` produces:

~~~js
var ASM_CONSTS = [function($0) { var name = UTF8ToString($0); return typeof process !== 'undefined' && name in process.env ? allocate(intArrayFromString(process.env[name]), 'i8', ALLOC_NORMAL) : 0; }];

function _emscripten_asm_const_ii(code, a0) {
  return ASM_CONSTS[code](a0);
}
function _getKeyboardListeningElement(){ return Module['keyboardListeningElement'] || '#document'; }
~~~

... on emscripten version `1.38.6`.

While the former seems to produce more boilerplate code around the function call, there are other functions in corrade that would generate those also. 

The `EM_ASM` version makes the necessity of the `std::free` way clearer. In the `EM_JS` version this is not apparent and I actually forgot the `std::free` here. We decided to use `EM_ASM` for the time being.

src/Magnum/Platform/EmscriptenApplication.cpp
Squareys: Yes, it holds, see the [note in the emscripten documentation](https://emscripten.org/docs/api_reference/html5.h.html#c.emscripten_set_scroll_callback)

CMakeLists.txt
```diff
@@ -103,7 +103,7 @@ option(WITH_SHADERS "Build Shaders library" ON)
...
 cmake_dependent_option(WITH_TEXTURETOOLS "Build TextureTools library" ON "NOT WITH_TEXT;NOT WITH_DISTANCEFIELDCONVERTER" ON)
 cmake_dependent_option(WITH_TRADE "Build Trade library" ON "NOT WITH_MESHTOOLS;NOT WITH_PRIMITIVES;NOT WITH_IMAGECONVERTER;NOT WITH_ANYIMAGEIMPORTER;NOT WITH_ANYIMAGECONVERTER;NOT WITH_ANYSCENEIMPORTER;NOT WITH_OBJIMPORTER;NOT WITH_TGAIMAGECONVERTER;NOT WITH_TGAIMPORTER" ON)
-cmake_dependent_option(WITH_GL "Build GL library" ON "NOT WITH_SHADERS;NOT WITH_GL_INFO;NOT WITH_ANDROIDAPPLICATION;NOT WITH_WINDOWLESSIOSAPPLICATION;NOT WITH_CGLCONTEXT;NOT WITH_GLXAPPLICATION;NOT WITH_GLXCONTEXT;NOT WITH_XEGLAPPLICATION;NOT WITH_WINDOWLESSWGLAPPLICATION;NOT WITH_WGLCONTEXT;NOT WITH_WINDOWLESSWINDOWSEGLAPPLICATION;NOT WITH_DISTANCEFIELDCONVERTER" ON)
+cmake_dependent_option(WITH_GL "Build GL library" ON "NOT WITH_SHADERS;NOT WITH_GL_INFO;NOT WITH_ANDROIDAPPLICATION;NOT WITH_WINDOWLESSIOSAPPLICATION;NOT WITH_CGLCONTEXT;NOT WITH_GLXAPPLICATION;NOT WITH_GLXCONTEXT;NOT WITH_XEGLAPPLICATION;NOT WITH_WINDOWLESSWGLAPPLICATION;NOT WITH_WGLCONTEXT;NOT WITH_WINDOWLESSWINDOWSEGLAPPLICATION;NOT WITH_EMSCRIPTENAPPLICATION;NOT WITH_DISTANCEFIELDCONVERTER" ON)
```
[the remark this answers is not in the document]
Squareys: Was a very good idea, found quite a couple of issues with the GL-less build.
I think the SDL has compile issues without GL, though 🤔  That appears to be the combination of emscripten and non-GL. That would not make sense for Sdl2Application to support, though, right?

src/Magnum/Platform/EmscriptenApplication.cpp
mosra: Hmm. That led me to think we could attach the callback to the window instead of this ugly polling. Will try this locally.

CMakeLists.txt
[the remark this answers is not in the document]
mosra: Thanks!

SDL should compile w/o GL mainly on desktop and mobile (Vulkan), I don't care much if it breaks on Emscripten tho, since it was superseded by your implementation now :) For EmscriptenApplication we should be able to turn off GL either for WebGPU or 3rd party renderers.

Squareys:
Emscripten uses `String.fromCharCode()` to convert `event.charCode` into a string for the `SDL_TextInputEvent`. That attribute is deprecated and it is recommended to use `event.key` instead, which is already a UTF8 string. Merely my use of that string was incorrect and it works now.

src/Magnum/Platform/EmscriptenApplication.cpp
```diff
@@ -506,8 +506,9 @@ void EmscriptenApplication::setupCallbacks(bool resizable) {
     emscripten_set_keydown_callback(keyboardListeningElement, this, false,
         ([](int, const EmscriptenKeyboardEvent* event, void* userData) -> Int {
             EmscriptenApplication& app = *static_cast<EmscriptenApplication*>(userData);
-            if(app.isTextInputActive() && std::strlen(event->key) == 1) {
-                TextInputEvent e{*event, {event->key, 1}};
+            const std::size_t keyLen = std::strlen(event->key);
+            if(app.isTextInputActive() && keyLen != 0) {
+                TextInputEvent e{*event, {event->key, keyLen}};
```
mosra: Right, in my patch I forgot to change the `1` and so the UTF-8 char got sliced, producing crap :sweat_smile: 

Wait ... but won't this add stuff like Shift or F1 to text input as well?

Squareys: No, that's only in `event.code`, not `event.key`... I think, let me check

Squareys: Oh no! You're right :(

mosra: Use my patch then :)

src/Magnum/Platform/EmscriptenApplication.cpp
```diff
@@ -562,14 +562,62 @@ EmscriptenApplication::GLConfiguration::GLConfiguration():
...
+    return 0;
+}
+
+void EmscriptenApplication::redraw() {
+    _flags |= Flag::Redraw;
+
+    /* Loop already running, no need to start */
+    if(!!(_flags & Flag::LoopActive)) return;
+
+    // TODO: @mosra There is a race condition potential here,
```
Squareys: @mosra This is up for discussion.

src/Magnum/Platform/EmscriptenApplication.h
```diff
@@ -1486,7 +1486,8 @@ code. See
 @code{.cpp}
 int main(int argc, char** argv) {
     className app({argc, argv});
-    return app.exec();
+    app.exec();
+    return 0;
```
mosra: Nope, app.exec() returns `0` now (see https://github.com/mosra/magnum/pull/300/commits/2383159ce4269901af3e86924270f9402a06b4a8). Actually, the macro should be updated to `return app.exec()` as well.

Squareys: Ah, I see

src/Magnum/Platform/EmscriptenApplication.h
```diff
@@ -1485,19 +1507,24 @@ code. See
...
-    app.exec();
+    className* app = new className({argc, argv});
+    app->exec();
     return 0;
 }
 @endcode
 
+As asynchronous javascript code may still require the application instance
+after @ref EmscriptenApplication::exec() finishes, the instance is allocated
+on heap and not deallocated. See @ref Platform-EmscriptenApplication-browser-main-loop
```
mosra: maybe "allocated on heap and intentionally leaked"?

src/Magnum/Platform/EmscriptenApplication.cpp
```diff
@@ -0,0 +1,707 @@
...
+        auto& app = *static_cast<EmscriptenApplication*>(userData);
+
+        if((app._flags & Flag::Redraw) && !(app._flags & (Flag::ExitRequested))) {
+            app._flags &= ~Flag::Redraw;
+            app.drawEvent();
+        }
+
+        /* If redraw is requested, we will not cancel the already requested
+           animation frame */
+        if(!!(app._flags & (Flag::Redraw)) && !(app._flags & (Flag::ExitRequested))) {
```
mosra:
```suggestion
        if(!!(app._flags & Flag::Redraw) && !(app._flags & Flag::ExitRequested)) {
```

(too) many braces, ((same above)) :trollface:

src/Magnum/Platform/EmscriptenApplication.cpp
mosra: What about making `redraw()` really just the flag setter like it was before and move all the rest into a dedicated function that gets explicitly called at the end of every event callback? That way it shouldn't happen that it gets called from inside itself.

src/Magnum/Platform/EmscriptenApplication.cpp
Squareys: Woups, that came from experimentation with trying to merge the two conditions into one &

mosra:
:shipit: :tada:

Squareys:
@mosra Thanks for merging! And especially thanks for all the great cleanup work! 🎉

PR #306, opened 2019-01-19
[WIP] SIMD playground

mosra:
Right now this is rather an experiment investigating possibilities of integrating SIMD-enabled algorithms into the math library. Currently there's the simplest algorithms ever, packing and unpacking index arrays to larger/smaller representations (and I was not able to beat the compiler, my AVX2 version was as what GCC managed to optimize out of the scalar implementation). It also shows how GCC function multi-versioning could be used to have multiple implementations in the same binary instead of compiling always only for one target.

Things to do:

- [ ] implement some `Utility::System::alignedAllocation()` that uses [posix_memalign](https://stackoverflow.com/a/6974037) (or the Windows equivalent), since aligned allocations are not available until C++17
- [ ] Think more about the design of the SIMD dispatch tags (should they have its own namespace? should I name them `Math::SseSimdT` instead to avoid colon cancer? should I put them into the Tags.h header to avoid having an excessive number of extremely small headers? how to make the templated benchmarks nicer without using the ugly `T{typename T::Init{}}`?)
- [ ] Should the particular tagged implementations be exposed as well? (if not, then the SIMD tags are useless to have publicly) Combine their docs into a single documented function explaining what tags can be used instead of having it repeated 4+ times?
- [ ] make the algorithms behave properly with "leftovers" (not a full block of 16 elements at the end)
- [ ] investigate [perf effects of aligned/unaligned loads and stores](https://stackoverflow.com/questions/20259694/is-the-sse-unaligned-load-intrinsic-any-slower-than-the-aligned-load-intrinsic-o) (this is where it could be possible to beat the compiler, if I force the input data to be aligned)
- [ ] investigate possible slowdowns due to [accidental leftovers in AVX registers](https://github.com/Kobzol/hardware-effects/tree/master/floating-point#mixing-sse-and-avx-instructions)
- [ ] GCC is warning about unused functions in the multi-versioned case, not sure why? is that a bug?
- [ ] investigate what subset of FMV is available on GCC 4.8 (is AVX2 there?)
- [ ] investigate how to do intrinsics on MSVC, if there's any possibility of FMV-like functionality or we need to select at compile time (using some builtin defines? which defines?)
- [ ] look into ARM NEON for these simple cases
- [ ] ryg has SIMD-enabled half->float variants on https://gist.github.com/rygorous/2144712, implement them? is there also float->half?

### DIFF
--- a/doc/snippets/MagnumMath.cpp
+++ b/doc/snippets/MagnumMath.cpp
@@ -878,6 +878,40 @@ Debug{} << Math::Vector3<UnsignedShort>{a};  // prints {16968, 48552, 15993}
 }
 
 {
+/* [unpack-template-explicit] */
+// Literal type is (signed) char, but we assumed unsigned char, a != 1.0f
+Float a = Math::unpack<Float>('\xFF');
+
+// b = 1.0f
+Float b = Math::unpack<Float, UnsignedByte>('\xFF');
+/* [unpack-template-explicit] */
+static_cast<void>(a);
+static_cast<void>(b);
+}
+
+{
+/* [unpack] */
+Float a = Math::unpack<Float, UnsignedShort>(8191);     // 0.124987f
+Float b = Math::unpack<Float, UnsignedShort, 14>(8191); // 0.499969f
+Float c = Math::unpack<Float, 14>(8191u);               // 0.499969f
+Float d = Math::unpack<Float, 14>(8191);                // 1.0f
+/* [unpack] */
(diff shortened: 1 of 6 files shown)


mosra:
Compile-time and runtime CPU feature detection APIs were moved to Corrade, as it makes more sense for them to live there. There's still a few unresolved design / implementation issues so it's in a draft PR for now: mosra/corrade#115

PR #307, opened 2019-01-26
package/msys: Add PKGBUILDs.

williamjcm:
Also add a .gitattributes file to force LF line endings.

Once again, both PKGBUILDs were tested and built without issues on both i686 and x86_64 MinGW-w64.

Comments and feedback welcome!


package/msys/PKGBUILD
```diff
@@ -0,0 +1,67 @@
...
+    MSYS2_ARG_CONV_EXCL="-DCMAKE_INSTALL_PREFIX=" \
+        ${MINGW_PREFIX}/bin/cmake .. \
+            -G'Ninja' \
+            -DCMAKE_BUILD_TYPE=Debug \
+            -DCMAKE_INSTALL_PREFIX=${MINGW_PREFIX} \
+            -DWITH_AUDIO=ON \
+            -DWITH_SHAPES=ON \
+            -DWITH_VK=ON \
+            -DWITH_GLFWAPPLICATION=ON \
+            -DWITH_GLUTAPPLICATION=ON \
```
mosra: `GlutApplication` will be soon removed (it's now deprecated), so this doesn't need to be here. Thus also the `freeglut` dependency. I had it in the development PKGBUILD only because I still need to ensure it compiles correctly.

And the `Shapes` library is scheduled for removal as well, so no need for `WITH_SHAPES` either.

package/msys/PKGBUILD
```diff
@@ -0,0 +1,67 @@
...
+pkgbase=mingw-w64-${_realname}-git
+pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}-git"
+pkgver=dev
+pkgrel=1
+pkgdesc="C++11/C++14 graphics middleware for games and data visualization"
+arch=('any')
+url="https://magnum.graphics/"
+license=('MIT')
+# The corrade-git dependency corresponds to the package generated by Corrade's packages/msys/PKGBUILD file.
+depends=("${MINGW_PACKAGE_PREFIX}-corrade-git"
```
mosra: Isn't the corrade package called just `${MINGW_PACKAGE_PREFIX}-corrade`? Same for the `-git` suffixes in `pkgbase` and `pkgname`.

Since this builds current working directory (instead of pulling a version from git), I wouldn't put a `-git` suffix on it.

package/msys/PKGBUILD
```diff
@@ -0,0 +1,67 @@
...
+options=('!strip')
+
+_rootdir=$startdir/../../
+
+build() {
+    mkdir -p "$_rootdir/build-${CARCH}"
+    cd "$_rootdir/build-${CARCH}"
+    
+    newcxxflags=$(echo $CXXFLAGS | sed s/-O.//g | sed s/-D_FORTIFY_SOURCE=.//g)
+    export CXXFLAGS="$newcxxflags"
```
mosra: This was rather a productivity hack for me (my default CXXFLAGS are rather crazy), no need to have it here.

package/msys/magnum/PKGBUILD
```diff
@@ -0,0 +1,74 @@
...
+            -DWITH_VK=ON \
+            -DWITH_WAVAUDIOIMPORTER=ON \
+            -DWITH_WINDOWLESSWGLAPPLICATION=ON \
+            "${extra_config[@]}" \
+            ../${_realname}-${pkgver}
+    ninja
+}
+
+package() {
+    #DESTDIR="$pkgdir" ninja -C build-${CARCH} install
```
mosra: I removed this commented-out line for `corrade`, so I guess you can do it here as well :)

package/msys/magnum/PKGBUILD
```diff
@@ -0,0 +1,74 @@
...
+    if check_option "debug" "n"; then
+        extra_config+=("-DCMAKE_BUILD_TYPE=Release")
+    else
+        extra_config+=("-DCMAKE_BUILD_TYPE=Debug")
+    fi
+    
+    MSYS2_ARG_CONV_EXCL="-DCMAKE_INSTALL_PREFIX=" \
+        ${MINGW_PREFIX}/bin/cmake \
+            -G'Ninja' \
+            -DCMAKE_CXX_FLAGS="$CXXFLAGS -Wfatal-errors" \
```
mosra: Huh, not sure what `@xyproto` had this for. Looks like a leftover from some debugging session. Remove this line?

package/msys/PKGBUILD
williamjcm: Will do.

package/msys/PKGBUILD
williamjcm: I think the `-git` suffix is also in the equivalent Corrade PKGBUILD, so it should also be removed there.

package/msys/magnum/PKGBUILD
williamjcm: Yeah. I don't know why I kept it in the first place.

package/msys/PKGBUILD
mosra: Oh, huh, right. Not sure where I was looking then.

Will push that right away together with the `git` dependency addition.

mosra:
Looks great now. Please give me a while to merge this, I'm still in the middle of the unique_ptr removal :sweat_smile:

mosra:
Merged in 107253f5fb17940f98c06109005da03c209b7afd, thanks a lot! :)

PR #317, opened 2019-02-18
package/msys/magnum: update PKGBUILD optdepends

williamjcm:
This change depends on mosra/magnum-integration#39's merge.


mosra:
Thank you!

PR #325, opened 2019-03-08
Sdl2Application: add glContext() accessor.

williamjcm:
Some SDL2 functions, like [`SDL_GL_MakeCurrent()`](https://wiki.libsdl.org/SDL_GL_MakeCurrent), take a `SDL_GL_Context` as part of their arguments.


mosra:
Merged as daf1deab3e6f65d15dd453d641a5bde311c503be, hiding this function from Emscripten builds. Thank you! :+1:

Issue #149, opened 2016-06-04
Audio library and OpenAL on other platforms


mosra:
Currently the Audio library is known to work on Windows, Linux and OSX. It compiles for iOS so I assume it works also there. The remaining platforms that aren't tested are:

- [ ] Emscripten -- OpenAL is provided there somehow, but I had no chance to test it.
- [ ] Windows RT -- https://github.com/kakashidinho/openal-soft-winphone
- [ ] Android -- OpenAL Soft should have a OpenSL ES backend. Some outdated info here: http://stackoverflow.com/a/7506114/6108877

It would be great to investigate the possibilities, document them and enable the Audio library in the CI builds for those platforms.

Squareys:
@mosra MinGW + Windows:

from https://chocolatey.org/packages/mingw
> The default package now uses SJLJ (32-bit) and SEH (64-bit) exception handling and posix threads. Other configurations can be selected using the params:
exception: sjlj, seh (64-bit only), dwarf (32-bit only)
threads: posix, win32
For example: -params "/exception:sjlj /threads:win32"

So you are building 64 bit binaries, but copying `Win32/soft_al.dll` instead of `Win64/soft_al.dll`.
Either use 32bit MinGW or copy `Win64/soft_al.dll` to `OpenAL32.dll` (yes, 32 is intentional).

Alicemargatroid:
@mosra Emscripten + OpenAL is working, but the current FindOpenAL does not work.
Emscripten's has its own OpenAL that works, which I just copy into a directory and point Magnum towards on Emscripten builds

stevetranby:
Is the Android OpenSL backend working yet?

Squareys:
@stevetranby `Magnum::Audio` uses OpenAL. https://github.com/kcat/openal-soft is an implementation of OpenAL which also supports OpenSL backend. `Magnum::Audio` should work perfectly fine with that.

mosra:
@Alicemargatroid Amazing, thank you! I forked the upstream FindOpenAL.cmake and made modifications in order to make it working on both desktop/mobile systems and Emscripten and now I'm also trying to enable building of Audio library on the CI. It's the `openal-emscripten` branch, once it goes green, I'll merge the changes to master.

mosra:
The Audio library is building on Windows CIs for some time already and with 85979ad996c3d632c4f99da5335695d9ab636e7d it is also fully working on the Emscripten and passing tests on the CI.

[...]

PR #326, opened 2019-03-09
Move audio import into Trade

mosra:
Halfway through adding file callback support to `Audio::AbstractImporter` for consistency with scene and font importers I realized it just doesn't make sense to have audio import separate in the Audio library. Reasons:

- `Audio` and the `BufferFormat` enum depend on OpenAL, while the actual data import doesn't, *at all*, basically preventing users from using the importer plugins together with, let's say, fmod, SoLoud or other implementations
- the importers are built for importing just one audio track -- what about playlists etc.? once we add multi-track support, we're even closer to `Trade::AbstractImporter`
- complex scenes that have animations usually *also* have audio data for the animations, audio attached to various objects, having various orientation etc. -- but at the moment, there's no builtin way to connect these two apart from digging manually into `importerState()` -- and then, if the audio data are embedded in the file, no way of passing those to `AudioImporter`s
- there's the proposed [MSFT_audio_emitter](https://github.com/KhronosGroup/glTF/pull/1400) extension for glTF (or [this](https://github.com/Squareys/Blender-OpenGEX/wiki/AudioSource-Extension) for OGEX), would be nice to have that supported in the [Magnum Player](https://magnum.graphics/showcase/player/), for example

A possible follow-up question could be if `Text::AbstractFont` should get merged into `Trade` as well. In my opinion nope, since text rendering is a *very* specific thing and it's far from "having a buffer with data that you pass to the GPU or the audio card". Besides that, text objects are usually not embedded in scenes (unlike audio) and if such need arises, it can be always done via a format-specific extension.

Things to do:

- [ ] drop the WIP commit I have here, as nothing there is useful anyway
- [ ] introduce a new `Magnum::AudioFormat` zero-based enum
    - [ ] provide its mapping to `Audio::BufferFormat`
    - [ ] what about all the weird mu-Law things? Supported by the Wav importer, but ...
- [ ] new `Trade::AudioData` class, wrapping format, frequency and buffer data
- [ ] add `audioCount()`, `audioName()`, `audioForName()`, `audio()` accessors to ``Trade::AbstractImporter`
- [ ] rewrite `AnyAudioImporter` and `WavAudioImporter` to be a part of `Trade`
- [ ] start testing audio import on Android (was blocked by #149 until now)
- [ ] rewrite all other audio importer plugins 
- [ ] stop depending on Audio and OpenAL in the plugins repo
    - [ ] simplify the CIs
    - [ ] test audio plugins on Android there
- [ ] think about how can we provide backwards compatibility? `Audio::AbstractImporter` hooking into `Trade`? ugh

### DIFF
--- a/src/Magnum/Audio/AbstractImporter.cpp
+++ b/src/Magnum/Audio/AbstractImporter.cpp
@@ -26,8 +26,11 @@
 #include "AbstractImporter.h"
 
 #include <Corrade/Containers/Array.h>
+#include <Corrade/Containers/Optional.h>
 #include <Corrade/Utility/Assert.h>
 #include <Corrade/Utility/Directory.h>
+
+#include "Magnum/FileCallback.h"
 
 #ifndef CORRADE_PLUGINMANAGER_NO_DYNAMIC_PLUGIN_SUPPORT
 #include "Magnum/Audio/configure.h"
@@ -36,7 +39,7 @@
 namespace Magnum { namespace Audio {
 
 std::string AbstractImporter::pluginInterface() {
-    return "cz.mosra.magnum.Audio.AbstractImporter/0.1";
+    return "cz.mosra.magnum.Audio.AbstractImporter/0.2";
 }
 
 #ifndef CORRADE_PLUGINMANAGER_NO_DYNAMIC_PLUGIN_SUPPORT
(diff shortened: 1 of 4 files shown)


hsdk123:
Looking forward to this!

PR #331, opened 2019-04-04
GL: fix segfault in Context destructor if constructed with NoCreate

xqms:
If we use the `GL::Context(NoCreate, ...)` constructor and never actually create it, the destructor will segfault, since it unconditionally deletes `_state`.

In my code, it happens like this:
```
Platform::GLContext context{NoCreate, argc, argv.data()};
...
if(somethingFailed)
   return; // segfault here

context.tryCreate();
```

### DIFF
--- a/src/Magnum/GL/Context.cpp
+++ b/src/Magnum/GL/Context.cpp
@@ -495,20 +495,17 @@ Context::Context(Context&& other): _version{other._version},
     _extensionRequiredVersion{Containers::NoInit},
     _extensionStatus{other._extensionStatus},
     _supportedExtensions{std::move(other._supportedExtensions)},
-    _state{other._state},
+    _state{std::move(other._state)},
     _detectedDrivers{std::move(other._detectedDrivers)}
 {
     /* StaticArray is deliberately non-copyable */
     for(std::size_t i = 0; i != Implementation::ExtensionCount; ++i)
         _extensionRequiredVersion[i] = other._extensionRequiredVersion[i];
 
-    other._state = nullptr;
     if(currentContext == &other) currentContext = this;
 }
 
 Context::~Context() {
-    delete _state;
-
     if(currentContext == this) currentContext = nullptr;
 }
 
@@ -731,7 +728,7 @@ bool Context::tryCreate() {
         }
     }
 
-    _state = new Implementation::State{*this, output};
+    _state.emplace(*this, output);
 
     /* Print a list of used workarounds */
     if(!_driverWorkarounds.empty()) {
--- a/src/Magnum/GL/Context.h
+++ b/src/Magnum/GL/Context.h
@@ -33,6 +33,7 @@
 #include <vector>
 #include <Corrade/Containers/EnumSet.h>
 #include <Corrade/Containers/Optional.h>
+#include <Corrade/Containers/Pointer.h>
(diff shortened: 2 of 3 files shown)


xqms:
If you would like a unit test for this, let me know. I have no idea where to put it though - GL or Platform?

xqms:
... and maybe a cleaner solution would be `Corrade::Pointer` instead of a raw pointer?

mosra:
Thanks! :+1: :100: 

Yes, `Corrade::Pointer` would be very nice to have here -- originally I hesitated using `std::unique_ptr` because of the compile-time overhead, but since there's a lightweight type now, it makes sense to use it.

The unit test for this could go into `GL/Test/ContextTest.cpp` (*not* in `ContextGLTest`) -- simply construct a `NoCreate` instance of `Platform::GLContext` and then let it go out of scope, similarly how's that done [here for Buffer](https://github.com/mosra/magnum/blob/ff947257aa66aa3c3ea8753690761fa726fa6ba5/src/Magnum/GL/Test/BufferTest.cpp#L57-L64). Actually I'm not sure why this test wasn't there already, hopefully there's no unsolvable design issue preventing this from being tested :)

xqms:
Thanks for the suggestions, will update the PR :+1:

By the way, your response time is awesome ;-)

xqms:
I'm seeing a linking problem now in the unit test, it seems MagnumGL does not link to flextGL and I would need to link to some Platform library?

```
contrib/magnum/src/Magnum/GL/Test/CMakeFiles/GLContextTest.dir/ContextTest.cpp.o: In function `Magnum::Platform::GLContext::GLContext(Magnum::NoCreateT, int, char const**)':
contrib/magnum/src/Magnum/Platform/GLContext.h:100: undefined reference to `flextGLInit'
```

The reason that I want to test with `Platform::GLContext` rather than `GL::Context` is that the `NoCreate` constructor is private - maybe it would make sense to have a local `friend`ed class in the test file and avoid the `Platform::GLContext` that way?

mosra:
Yeah I know, it's private for a reason :) Can you try linking it to the `MagnumOpenGLTester` library (like the `*GLTest` tests)? So it doesn't attempt to create a context but still picks up all needed libraries. That should do the trick.

xqms:
> Can you try linking it to the MagnumOpenGLTester library (like the *GLTest tests)?

That works, but needs `WITH_OPENGLTESTER`. Should we simply add the test function to `ContextGLTest.cpp`?

mosra:
Uh oh. That might be the "unsolvable design issue preventing this from being tested" I was talking about :D I'll try to figure out a solution. How fast do you need this to be merged?

xqms:
No rush, this only happens in my error path and leads to a crash instead of an exit with error code. It doesn't impact my work at all.

In any case, I pushed the last version with the test in `ContextGLTest.cpp`, that at least compiles and runs. I understand that it would be better to have the test in the non-GL testsuite, though.

mosra:
Merged in c38319ef0001d986c8210b9f5996ab1bf7029534 and f58c26c66d679aa3d4c5890b6683d0855c36d282, thank you! In the end I made the relevant methods of `GL::Context` protected instead of private, which allowed me to both have it testable in `ContextTest.cpp` and removed the need for friending `Platform::GLContext`.

PR #332, opened 2019-04-09
Platform: improve {Glfw,Sdl2}Application::exit().

williamjcm:
Now, it's possible to exit an application properly with a non-zero exit code.

On my end, I could only test the `Sdl2Application` version.

Review and feedback are welcome!


mosra:
Merged, thanks!

PR #338, opened 2019-05-10
Audio: add buffer (un)queuing to Source.

williamjcm:
And also add an implementation for Source::type().

Reviews and feedback are welcome!

### DIFF
--- a/src/Magnum/Audio/Source.cpp
+++ b/src/Magnum/Audio/Source.cpp
@@ -57,7 +57,7 @@ std::size_t Source::unqueueBuffers(Containers::ArrayView<Containers::Reference<B
 
     Containers::Array<ALuint> unqueuedIds(processedBuffers);
     alSourceUnqueueBuffers(_id, unqueuedIds.size(), unqueuedIds.data());
-    auto isNotUnqueued = [&unqueuedIds](Containers::Reference<Buffer> buffer){
+    auto isNotUnqueued = [&unqueuedIds](Buffer& buffer) {
         for(ALuint id : unqueuedIds) {
             if(buffer->id() == id)
                 return false;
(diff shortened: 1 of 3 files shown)


src/Magnum/Audio/Source.cpp
```diff
@@ -38,6 +41,32 @@ Source& Source::setBuffer(Buffer* buffer) {
...
+
+std::size_t Source::unqueueBuffers(Containers::ArrayView<Containers::Reference<Buffer>> buffers) {
+    ALint processedBuffers;
+    alGetSourcei(_id, AL_BUFFERS_PROCESSED, &processedBuffers);
+
+    if(!processedBuffers) return 0;
+
+    Containers::Array<ALuint> unqueuedIds(processedBuffers);
+    alSourceUnqueueBuffers(_id, unqueuedIds.size(), unqueuedIds.data());
+    auto isNotUnqueued = [&unqueuedIds](Containers::Reference<Buffer> buffer){
```
mosra: would the following work too?

```suggestion
    auto isNotUnqueued = [&unqueuedIds](Buffer& buffer) {
```

src/Magnum/Audio/Source.h
```diff
@@ -734,6 +758,12 @@ inline Source& Source::operator=(Source&& other) {
     return *this;
 }
 
+inline auto Source::type() const -> Type {
```
mosra: As we discussed on Gitter, can you add a simple test for this? Thanks!

williamjcm: I forgot to push it. 🤦‍♂

src/Magnum/Audio/Source.cpp
williamjcm: I forgot that `Reference<Type>` was convertible to `Type&`. 🤦‍♂

mosra:
Merged in b9fb03ab8e30e52e66bd3da6ab5deefff42aabfe and 1104d8bb15cfb42e89aa89d1248d2083239e89eb, thank you! While at it, in c442c07190c010f854a3f5e1216f4e7e26b5773b I added also debug output for `Source::Type`.

PR #342, opened 2019-05-27
Audio: add support for buffer loop points.

williamjcm:
This feature requires the `AL_SOFT_loop_points` extension.

Also, a test for `Buffer::setData()` was added to `BufferALTest`, along with the tests for the new functions.

Review/feedback welcome!

### DIFF
--- a/src/Magnum/Audio/Buffer.cpp
+++ b/src/Magnum/Audio/Buffer.cpp
@@ -3,7 +3,6 @@
 
     Copyright © 2010, 2011, 2012, 2013, 2014, 2015, 2016, 2017, 2018, 2019
               Vladimír Vondruš <mosra@centrum.cz>
-    Copyright © 2015 Jonathan Hale <squareys@googlemail.com>
     Copyright © 2019 Guillaume Jacquemin <williamjcm@users.noreply.github.com>
 
     Permission is hereby granted, free of charge, to any person obtaining a
@@ -25,41 +24,31 @@
     DEALINGS IN THE SOFTWARE.
 */
 
-namespace Magnum {
+#include "Buffer.h"
 
-/** @page openal-support OpenAL support state
-@brief List of (un)supported OpenAL features and extensions.
+#include "Magnum/Math/Functions.h"
 
-@tableofcontents
-@m_footernavigation
+namespace Magnum { namespace Audio {
 
-@section openal-support-state OpenAL implementation state
+Buffer& Buffer::setLoopPoints(Int loopStart, Int loopEnd) {
+    ALint sizeInBytes;
+    ALint channels;
+    ALint bits;
 
-The extension implementation is considered complete if all its defined types,
-functions and enum values are exposed through the API.
+    alGetBufferi(_id, AL_SIZE, &sizeInBytes);
+    alGetBufferi(_id, AL_CHANNELS, &channels);
+    alGetBufferi(_id, AL_BITS, &bits);
 
-@subsection openal-extension-support Extensions
+    ALint lengthInSamples = sizeInBytes * 8 / (channels * bits);
 
-@m_class{m-fullwidth}
+    ALint loopPoints[2] = {
+        Math::max<ALint>(0, loopStart),
+        Math::min<ALint>(loopEnd, lengthInSamples)
+    };
 
-Extension                                   | Status
-------------------------------------------- | ------
-@alc_extension{ENUMERATION,EXT}             | done
-@al_extension{EXT,double}                   | done
-@al_extension{EXT,float32}                  | done
-@al_extension{EXT,ALAW}                     | done
-@al_extension{EXT,MULAW}                    | done
-@al_extension{EXT,MCFORMATS}                | done
+    alBufferiv(_id, AL_LOOP_POINTS_SOFT, &loopPoints[0]);
 
-@subsection openal-extension-support-soft OpenAL Soft Extensions
-
-@m_class{m-fullwidth}
-
-Extension                                   | Status
-------------------------------------------- | ------
-@alc_extension{SOFTX,HRTF}                  | done
-@alc_extension{SOFT,HRTF}                   | done
-@al_extension{SOFT,loop_points}             | done
-
-*/
+    return *this;
 }
+
+}}
--- a/src/Magnum/Audio/Buffer.cpp
+++ b/src/Magnum/Audio/Buffer.cpp
@@ -31,19 +31,9 @@
 namespace Magnum { namespace Audio {
 
 Buffer& Buffer::setLoopPoints(Int loopStart, Int loopEnd) {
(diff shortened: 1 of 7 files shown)


src/Magnum/Audio/Test/BufferALTest.cpp
```diff
@@ -35,18 +39,109 @@ struct BufferALTest: TestSuite::Tester {
...
+        CORRADE_SKIP("Extension AL_SOFT_loop_points isn't supported on this system.");
+    }
+
+    Buffer buf;
+    constexpr char data[] { 25, 17, 24, 122, 67, 24, 48, 96 };
+
+    buf.setData(BufferFormat::Mono8, data, 22050).setLoopPoints(1, 6);
+
+    Containers::Array<ALint> points{Containers::InPlaceInit, { 0, 0 }};
+    alGetBufferiv(buf.id(), AL_LOOP_POINTS_SOFT, points.data());
```
mosra: I'm wondering, wouldn't it make sense to add a getter for these? Returning `std::pair<ALint, ALint>`, for example. I think it could make testing and debugging easier. (Then you also don't need to do the complex container testing, just `CORRADE_COMPARE(buf.loopPoints(), std::make_pair(0, 8))` for example.)

src/Magnum/Audio/Test/BufferALTest.cpp
```diff
@@ -35,18 +39,109 @@ struct BufferALTest: TestSuite::Tester {
...
 
+void BufferALTest::setData() {
+    Buffer buf;
+    constexpr char data[] { 25, 17, 24, 122, 67, 24, 48, 96 };
+
+    buf.setData(BufferFormat::Mono8, data, 22050);
+
+    ALint bufSize;
+    alGetBufferi(buf.id(), AL_SIZE, &bufSize);
+    CORRADE_VERIFY(bufSize == 8);
```
mosra: `CORRADE_COMPARE(bufSize, 8)` -- prints a much more useful message when the test fails.

src/Magnum/Audio/Buffer.h
```diff
@@ -102,6 +104,57 @@ class Buffer {
...
+         */
+        Buffer& setLoopPoints(Int loopStart, Int loopEnd);
+
+        /**
+         * @brief Set buffer to loop from the beginning until a certain point
+         *
+         * Equivalent to calling @ref setLoopPoints() with @p loopStart equal to
+         * 0.
+         * @param loopEnd   The loop's end point in samples
+         * @return Reference to self (for method chaining)
```
mosra: `@param` and `@return` goes under `@brief`, detailed docs go after that.

src/Magnum/Audio/Buffer.cpp
```diff
@@ -0,0 +1,54 @@
...
+namespace Magnum { namespace Audio {
+
+Buffer& Buffer::setLoopPoints(Int loopStart, Int loopEnd) {
+    ALint sizeInBytes;
+    ALint channels;
+    ALint bits;
+
+    alGetBufferi(_id, AL_SIZE, &sizeInBytes);
+    alGetBufferi(_id, AL_CHANNELS, &channels);
+    alGetBufferi(_id, AL_BITS, &bits);
```
mosra: Hmm, would these be useful as public getters as well?

src/Magnum/Audio/Buffer.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+    ALint lengthInSamples = sizeInBytes * 8 / (channels * bits);
+
+    ALint loopPoints[2] = {
+        Math::max<ALint>(0, loopStart),
+        Math::min<ALint>(loopEnd, lengthInSamples)
+    };
```
mosra: What happens when `loopStart` > `loopEnd`? Does the extension say anything about that?

src/Magnum/Audio/Buffer.h
williamjcm: In some of the docs, the detailed docs were above. I think `Buffer.h` itself has an example.

src/Magnum/Audio/Buffer.cpp
williamjcm: > An AL_INVALID_VALUE error is generated if alBufferiv is called with offset
> values for AL_LOOP_POINTS_SOFT where the second value is not greater than
> the first value.

Right.

src/Magnum/Audio/Buffer.cpp
williamjcm: I added them to get the length of the buffer in sample, so yeah I guess I could also add a getter for them and the length.

src/Magnum/Audio/Test/BufferALTest.cpp
williamjcm: It'd make sense. I'll add it.

src/Magnum/Audio/Buffer.cpp
mosra: Ok, so (similarly to GL) this is user's responsibility to check, I guess.

src/Magnum/Audio/Test/BufferALTest.cpp
williamjcm: A quick test on my end leads to a success. Marking this as resolved.

src/Magnum/Audio/Test/BufferALTest.cpp
williamjcm: Changed locally. Marking this as resolved.

src/Magnum/Audio/Buffer.h
williamjcm: Changed locally.

src/Magnum/Audio/Buffer.cpp
williamjcm: Yeah.

src/Magnum/Audio/Test/BufferALTest.cpp
mosra: Yes, they should be, `std::pair` is a plain struct. I just realized that maybe `Range1Di` could fit the criteria here as well. Either is okay from my side.

src/Magnum/Audio/Buffer.cpp
williamjcm: Done locally. Will push ASAP. Marking this as resolved.

src/Magnum/Audio/Test/BufferALTest.cpp
williamjcm: `std::pair` it is, then. I saw the comment a bit late. 😅

mosra:
Merged as f5302edcc137c92bdbe13800be3cc16269568bc6 and 8029abca07d36171c57a8db041e6cc344f55ad35, with some very minor changes. Thank you!

Issue #340, opened 2019-05-21
2019.10 release


mosra:
It's becoming very unlikely that I'll be able to release 2019.0b until the end of May, which means no 2019.05. Because my OCD forces me to release on even-numbered months :laughing:, the next release will be ~~2019.07, optimistically at the beginning of July~~ ~~well... 2019.09.~~ Actually, what's wrong with 2019.10?

Things to do (list expected to grow *slightly*):

- [x] Finalize initial Python bindings and put them into master (mosra/magnum-bindings#1)
  - [x] Bind shader APIs
  - [x] Bind also textures and images, importers
  - [x] Look into faster buffer protocol
  - [x] Finish search and symbol crosslinking for Python docs
  - [x] Windows support (mosra/magnum-bindings#2)
  - [x] Port the textured triangle and viewer examples
  - [x] And publish a blog post
  - [x] Expose remaining primitives -- mosra/magnum-bindings@e334cc17b247697ec6f0d5294f96952451b4d689
  - [x] The Windows static build on CI gets stuck when running Magnum tests -- mosra/corrade@01d8cdd30cf9c51e3abc9280f7a6da6ad8c926b4, 909887ad7c05c92ba7904b5d443d9716ce8ddf06
- [x] Fix issues with multiple Context instances when having multiple `*.so`s with static Magnum (weak symbols, hopefully?) -- mosra/corrade@1bef56947f30fba91e448fe6e5561341a83c18d2, e52c9e8d9ee21a81f02cfabe631323d89dcfef7a
  - [x] Explicit static build testing for this in Corrade (plugin managers, Debug) and Magnum (GL/Audio Context) -- mosra/corrade@21ec22720f66bfe41d46a87b40ee3e0efc35c571, 59aade762a2996d336da522a9243697b1f87571c
  - [x] Make this work on Windows as well (`thread_local`s can't be dllexported and this damn platform doesn't have weak symbols, so this needs to be a crazy mess of DllMain and calling GetProcAddress on itself) -- mosra/corrade@7b98f1983cb1778a9cbb38184c42705320e0a266, 24f7d45364510a4521a2cb70ed5600a7a83bb142
- [x] Update for MSVC 2019 support (#343)
  - [x] Test on the CIs
  - [x] Adapt bootstrap projects (and add MSVC2019 CI for them) #354
- [x] Integrate the DART example (mosra/magnum-examples#63)
  - [x] And the blog post :)
[...]

PR #343, opened 2019-05-27
Support for VS2019

isc30:
Hi, this PR fixes compilation on VS19.
There is also a problem with "Cannot delete pointer to incomplete type" that I'm not sure how to fix.

### DIFF
--- a/src/Magnum/ResourceManager.h
+++ b/src/Magnum/ResourceManager.h
@@ -488,7 +488,7 @@ template<class ...Types> ResourceManager<Types...>*& ResourceManagerInlineInstan
 }
 
 template<class T> void safeDelete(T* data) {
-    static_assert(sizeof(T) > 0, "Cannot delete pointer to incomplete type");
+    //static_assert(sizeof(T) > 0, "Cannot delete pointer to incomplete type");
     delete data;
 }
 
--- a/src/Magnum/DebugTools/CompareImage.cpp
+++ b/src/Magnum/DebugTools/CompareImage.cpp
@@ -53,12 +53,14 @@ template<std::size_t size, class T> Float calculateImageDelta(const ImageView2D&
     /* Precalculate parameters for pixel access */
     Math::Vector2<std::size_t> dataOffset, dataSize;
 
+    int64_t dataOffsetSum = dataOffset.sum();
+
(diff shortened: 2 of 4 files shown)


mosra:
Hi! I was planning to do this by reusing a subset of workarounds that are enabled by the `MSVC2017_COMPATIBILITY` CMake option in Corrade (so introducing a new `MSVC2019_COMPATIBILITY` option and changing corresponding `#ifdefs`) -- I guess majority of them is still relevant, but some might not be needed anymore. Plus the AppVeyor CI setup has to be updated to regularly test with VS19, otherwise things would get broken by accident again.

If you enable the above option on your side, does everything compile cleanly or is there extra work needed to be done?

src/Magnum/ResourceManager.h
```diff
@@ -488,7 +488,7 @@ template<class ...Types> ResourceManager<Types...>*& ResourceManagerInlineInstan
 }
 
 template<class T> void safeDelete(T* data) {
-    static_assert(sizeof(T) > 0, "Cannot delete pointer to incomplete type");
+    //static_assert(sizeof(T) > 0, "Cannot delete pointer to incomplete type");
```
isc30: TODO: don't forget to fix this properly!

isc30:
Yup, if I set `MSVC2017_COMPATIBILITY` in CMake it builds without any problem.
It's your decision on how to fix it: adding a VS19 compatibility option that internally applies VS17 compatibility or fixing the code directly as I did. I could do both.
Thanks

mosra:
It's on my list for #340, so I'm going to look at it very soon one way or another. Can you live with the MSVC2017 option enabled for now?

Most of those are real workarounds for nonstandard compiler behavior and I'd rather keep them explicit to avoid them being reverted by accident. I still hope that at some point in the future there would be no need for such flags, but it's still not there yet unfortunately.

isc30:
Sure thing, I can live with the VS17 workarounds. Thanks

mosra:
For the record, MSVC 2019 support is now done with mosra/corrade@161c9e0, mosra/magnum@ddb16f6, mosra/magnum@9e2a111 and mosra/magnum@d542e2b :tada:

PR #344, opened 2019-05-30
Fixing gentoo build instructions

mtao:
Fixing gentoo build instructions ( -99999 -> 9999 in the ebuild path…); added media-libs/openal dependency to the ebuild


mosra:
Hi, thank you! Can you explain why 9999 is okay while 99999 is not? I Googled around, but didn't find much. Is it some convention for marking packages built from Git?

mosra:
Merged as bce9a9d5f0405098c9cd4b9baaa6d374834b5066, thank you! I also updated instructions for other repos accordingly.

PR #346, opened 2019-06-04
Added equal and not equal operators in Magnum::Resource

roig:
I added the equal and not equal operators to check if a resource is equal or not to another Resource.

My use case was this:
```
void Sprite::setTexture(const Magnum::Resource<Magnum::GL::Texture2D>& texture) {
	if (texture != _texture) {
		_texture = texture;
		updateVertices();
		updateUVs();
	}
}
```

### DIFF
--- a/src/Magnum/Resource.h
+++ b/src/Magnum/Resource.h
@@ -150,6 +150,12 @@ class Resource {
         /** @brief Move assignment */
         Resource<T, U>& operator=(Resource<T, U>&& other);
 
+        /** @brief Equal operator */
+        bool operator==(const Resource<T, U>& other) const;
+
+        /** @brief Not equal operator */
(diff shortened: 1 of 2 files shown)


mosra:
Thank you! Code looks good, can you add a test to `src/Magnum/Test/ResourceManagerTest.cpp` so code coverage is happy as well? For example fetching two different resources and `CORRADE_VERIFY()` that both `==` and `!=` do what expected.

roig:
Added the test!

mosra:
Merged as 909b05b8fa8573082667498c03616617afde305c, thank you! 

While going over the code I discovered that the comparison could lead to unwanted false positives when comparing different types, fixed that in f89e935f7d311b0e7dd4100639b77bf5e04dd03e.

PR #349, opened 2019-06-11
Platform: fix DpiScaling to handle calling conventions properly on Windows.

cegbertOculus:
The function imported from the dll was being implicitly cast to a __cdecl calling convention, but the function itself can be in a __stdcall calling convention, depending on the Windows build environment.  Adding  WINAPI to the cast puts the function in the correct calling convention regardless of the Windows build environment.


mosra:
Merged, thank you! (And sorry for the delay.)

PR #362, opened 2019-07-26
Audio: add support for quiet logging in Context.

williamjcm:
This required the addition of *many* constructor variants. 😅 

Review and feedback welcome!

### DIFF
--- a/src/Magnum/Audio/Context.cpp
+++ b/src/Magnum/Audio/Context.cpp
@@ -4,6 +4,7 @@
     Copyright © 2010, 2011, 2012, 2013, 2014, 2015, 2016, 2017, 2018, 2019
               Vladimír Vondruš <mosra@centrum.cz>
     Copyright © 2015 Jonathan Hale <squareys@googlemail.com>
+    Copyright © 2019 Guillaume Jacquemin <williamjcm@users.noreply.github.com>
 
     Permission is hereby granted, free of charge, to any person obtaining a
     copy of this software and associated documentation files (the "Software"),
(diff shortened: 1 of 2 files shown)


mosra:
Looks great! :+1: My only comment is that it could use some docs similar to [what GL::Context has](https://doc.magnum.graphics/magnum/classMagnum_1_1GL_1_1Context.html#GL-Context-command-line).

And as I mentioned on Gitter, before I can merge this I need to update `Arguments` to avoid this blowing up when e.g. `--magnum-gpu-validation` is passed on the command line (where `GL::Context` should use this option, but `Audio::Context` should ignore it -- basically I need to implement an option to ignore unknown arguments, that's it).

williamjcm:
> My only comment is that it could use some docs similar to [what GL::Context has](https://doc.magnum.graphics/magnum/classMagnum_1_1GL_1_1Context.html#GL-Context-command-line).

On it.

mosra:
Merged in 7d7ebb379cf21afc56ffe49d885ae44c71aae893! :tada: I modified the delegations to really parse the arguments in all cases (the `NoCreate` constructor w/o argc/argv would completely ignore `MAGNUM_LOG`, maybe that was the heisenbug you had? :slightly_smiling_face:). Plus added missing `noexcept` to the new `NoCreate` variants and added back the lost docblock.

While doing the work on ignoring unknown arguments (done in 2c280f6b970c3fb34346e5bc79a5d22b91ce242e), I realized I could extend `--magnum-disable-extensions` to work for AL as well (very important for testing, actually), so that's done too, in 8af19f21f06ec54a73aced357d08dd547cb82474. Feature parity, at last -- many thanks for realizing this in the first place!

Issue #363, opened 2019-08-19
Add support for iterating all the resource<T> from a ResourceManager.


roig:
It would nice to have this functionality in the ResourceManager. For example to check all the resources that are added, or for example to iterate them and show them in ImGui :+1: 

I'm trying to use AbstractResourceLoader to save the key as a workaround when calls:
`        virtual void doLoad(ResourceKey key) = 0;`

mosra:
As of 46377b70b93fa997216ac8267c9e60cbb7b2cabb the `ResourceManager` and all related APIs are deprecated, closing as obsolete.

PR #364, opened 2019-08-19
Added function to retrieve all the resources<T> from the ResourceManager

roig:
Hello mosra, I need that feature.  #363 

Let's discuss how to implement it :smile:  I added a getAll function. Should I go with tests or do you want that feature implemented in another way?

### DIFF
--- a/src/Magnum/ResourceManager.h
+++ b/src/Magnum/ResourceManager.h
@@ -118,6 +118,9 @@ template<class T> class ResourceManagerData {
 
         template<class U> Resource<T, U> get(ResourceKey key);
 
+        template<class U> std::vector<Resource<T, U>> getAll();
+
+
         void set(ResourceKey key, T* data, ResourceDataState state, ResourcePolicy policy);
 
         T* fallback() { return _fallback; }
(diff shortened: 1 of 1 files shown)


mosra:
Hi, sorry, being busy with other things. How fast you need that?

I think the best way to go about this is exposing a `resources<T>()` that returns something that can be iterated through (so, having a `begin()`/`end()`), as that's better than forcing the users into a `std::vector` and is also allocation-free. I have a sketch in mind, but won't be able to get to it until much later today (but more likely tomorrow) -- in the meantime, can you live with the code you have right now?

roig:
Yes of course. In the meantime I will try to implement it myself.

roig:
woah.. I don't have the knowledge to do this.. sorry. I will wait your implementation. :cry: 

I started a class Resources<T> with begin and end methods that will return a ResourceIterator. But don't know exactly how to retrieve iterators from Implementation::ResourceManagerData<T>.

mosra:
As of 46377b70b93fa997216ac8267c9e60cbb7b2cabb the `ResourceManager` and all related APIs are deprecated, closing as obsolete.

PR #365, opened 2019-08-19
GL: imageSize is not available for GLES < 3.1

msbaines:
Fixes the following compile error:

magnum/src/Magnum/GL/Texture.h:747:44: error: no member named 'imageSize' in 'Magnum::GL::AbstractTexture::DataHelper<2>'
            return DataHelper<dimensions>::imageSize(*this, level);
                                           ^


mosra:
This got overlooked, thank you!

PR #368, opened 2019-08-23
dox: Fix vcpkg install everything command

Squareys:
Hi @mosra !

Just came by this and remembered a discussion on gitter. To ensure it's still bareable for vcpkg users to install everything on windows I went trough all available packages and listed them.

Otoh: This is temporary and unacceptable in my opinion in terms of usability, since the average user will try `vcpkg install magnum[*]` first and having to check the docs to copy the entire list is not great either.

I propose `vcpkg install magnum[windowlesscontext]` to autodetect which one to build or (if some certain platforms have multiple contexts available) automatically detect features that cannot be built and silently (or with warning "skipping") remove them from the list.

I cannot promise that I will get to implementing such a thing too soon, though. Maybe with the next release update of the ports.

Cheers, Jonathan

### DIFF
--- a/doc/building.dox
+++ b/doc/building.dox
@@ -3,7 +3,7 @@
 
     Copyright © 2010, 2011, 2012, 2013, 2014, 2015, 2016, 2017, 2018, 2019
               Vladimír Vondruš <mosra@centrum.cz>
-    Copyright © 2018 Jonathan Hale <squareys@googlemail.com>
+    Copyright © 2018, 2019 Jonathan Hale <squareys@googlemail.com>
 
     Permission is hereby granted, free of charge, to any person obtaining a
     copy of this software and associated documentation files (the "Software"),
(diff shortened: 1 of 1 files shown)


mosra:
The *real* solution for this would be Vcpkg having per-platform `Default-Features`. I asked for this [when Linux support for the Magnum vcpkg package was added](https://github.com/microsoft/vcpkg/pull/3611#issuecomment-392820303) but didn't get any answer for that. Maybe opening a feature request for this? Or, optimistically, maybe such a feature was already added since then?

> or (if some certain platforms have multiple contexts available) automatically detect features that cannot be built and silently (or with warning "skipping") remove them from the list.

That would be an alternative solution -- can presence of `*` be detected? If so, then it could just prune the unavailable features ... but then vcpkg would think those are installed, while they in fact aren't?

mosra:
Um, before I merge this -- didn't `vcpkg install package[bla]` also implicitly include all defaults (and saying `vcpkg install package[core,bla]` didn't)? I remember some such thing. If that's the case, the line could be much shorter.

mosra:
@ras0219-msft oh, wonderful -- didn't know the `(windows)` / `(!uwp)` qualifiers were a thing until now. That solves everything (and it's no problem that it's a slight workaround), thank you!

alanjfs:
Ooo, just spotted this. This would have helped me the other day. Related to #373.

Squareys:
So, summing up the current state of this PR:

 - This changes the docs to recommend avoiding `magnum[*]` as it will install features that are platform-dependent. There is no native way to declare a feature to be available only on a single platform.
 - Vcpkg seem to be installing `magnum[*]` for tests, which causes problems out of aforementioned reason.
 - Platform dependent defaults are possible with the mentioned workaround.
 - Since the qualifiers can be used to have different dependencies for different platforms, this could be used to replace `windowlesswglapplication` and `windowlessglxapplication` with `windowlessapplication`, which then could be used to set the correct platform-dependent cmake option.

Anything I forgot? I think the most userfriendly would be to support `vcpkg install magnum[*]` in a way, and it would fix the tests that they do at vcpkg.

mosra:
This is unfortunately problematic because there are features that are inevitably available just on a single platform, even more so with the introduction of filesystems (https://github.com/mosra/corrade/pull/39).  Collapsing the current problematic ones to just `glcontext`, `windowlessapplication` could "fix" things right now but there will be still issues where for example on Linux you can have both GLX or EGL installed, or just one, or neither, and the package has no way to check for that.

One idea that could fix `[*]` is that the portfile just ignores (or ignores with a warning) features that are definitely not available on a particular platform (so e.g. it won't add `-DWITH_GLXCONTEXT=ON` to CMake on Windows). Could that work?

mosra:
This is now fixed with https://github.com/microsoft/vcpkg/pull/12211 by ignoring features that are incompatible with given platform.

alanjfs:
Oh joy, just in time for my return to Magnum. :D Thanks @mosra, see you in the chat shortly.

PR #369, opened 2019-08-27
Trade: support mip levels in image import

mosra:
There's a new level option in each `imageND()` API, plus an `imageNDLevelCount()` query to get image level count. The level defaults to 0, in which case it behaves exactly like before. For levels above 0, the abstract importer checks against the reported level count and asserts if it's out of bounds.

Things left to do:

- [x] changelog entry
- [x] adapt `DdsImporter`, use the levels instead of reporting N images
- [x] adapt `BasisImporter` (currently it opens just the base level)
- [x] adapt other image importer plugins
- [x] adapt `AssimpImporter`, `OpenGexImporter` and `TinyGltfImporter` plugins
  - [x] they need to properly propagate this to AnyImageImporter and should cache the importer instance to avoid a single image file being read multiple times for each level (and with assertions enabled, once more for the check in each `imageND()` call)
  - [x] doing this in a clean way may need to have `AnyImageImporter` movable
  - [x] reuse mipmap test files from `DdsImporter` and test level propagation on those

@Squareys since you're working on mosra/magnum-plugins#62, could you have a quick look at this to see if it fits your expectations or needs design changes? Thanks! :pray:

### DIFF
--- a/doc/changelog.dox
+++ b/doc/changelog.dox
@@ -150,6 +150,13 @@ See also:
     and @ref SceneGraph::AbstractBasicTranslationRotation3D::rotateLocal(const Math::Quaternion<T>&) "rotateLocal()"
     overloads taking a @ref Math::Quaternion
 
+@subsubsection changelog-latest-new-trade Trade library
+
+-   Ability to import image mip levels via an additional parameter in
+    @ref Trade::AbstractImporter::image2D(),
+    @ref Trade::AbstractImporter::image2DLevelCount() and similar APIs for 1D
+    and 3D images
+
 @subsubsection changelog-latest-new-vk Vk library
 
 -   Updated Vulkan headers for version 1.2
--- a/src/Magnum/Trade/AbstractImporter.cpp
+++ b/src/Magnum/Trade/AbstractImporter.cpp
@@ -52,7 +52,7 @@
 namespace Magnum { namespace Trade {
 
 std::string AbstractImporter::pluginInterface() {
-    return "cz.mosra.magnum.Trade.AbstractImporter/0.3";
+    return "cz.mosra.magnum.Trade.AbstractImporter/0.3.1";
 }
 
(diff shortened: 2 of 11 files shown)


Squareys:
At first I was concerned whether `image2d(index, level)` would enforce anti-patterns in the implementations, but I think quite the opposite is true, this is pretty great, actually :)

Looks good to me! 👍

mosra:
> would enforce anti-patterns in the implementations

I'm interested, what exactly? Just so I can be sure that I'm doing it correctly.

Squareys:
> what exactly? 

I was thinking that this might require doing per-image work per mip level. So if you'd call:
```cpp
for(int level = 0; level < maxLevel; ++level) {
    importer->image2D(0, level);
    /* this might do work maxLevel-1 times that would only need
       to be done once. E.g. figure out image metadata etc. */
}
```

But it turns out that in Basis and in DDS that part is already done in `doOpenData()` usually. Except for some light image info stuff in basis maybe.
Imagine a format that compressess over multiple mips and uncompresses all of them for every image2D call... per-image state would need to be either cached in the importer's state or re-computed for every mip.

mosra:
> Imagine a format that compressess over multiple mips and uncompresses all of them for every image2D call

Yeah, then the importer would need to be more clever and cache the uncompressed data for subsequent runs (or move some work from `doImage()` into the `doOpen*()` function). That's the same AssimpImporter & co. need to do as well -- avoid instantiating, opening and parsing the images for each subsequent loaded level.

mosra:
For completeness, the corresponding change in plugins is https://github.com/mosra/magnum-plugins/compare/15e890b8b0ac5a5b788cf1ed460fa89eb0c2156f...99b9e2f7ef7e57fb9331ac5e23023303fece6381. It's a bit more involved in order to be efficient when loading consecutive mip levels -- scene importers have to cache an existing image importer instance to avoid decoding the image again from scratch for each level.

PR #370, opened 2019-08-27
 AnyImageImporter: Delegate .basis files

Squareys:
[BasisImporter](https://github.com/mosra/magnum-plugins/pull/62) | [BasisImageConverter](https://github.com/mosra/magnum-plugins/pull/65) | >AnyImageImporter/Converter<
------------ | ------------- | -------------

Hello @mosra !

This part of https://github.com/mosra/magnum-plugins/pull/62 is hereby up for discussion.

**TODOs**
 - [x] Rebase
 - [x] Test ?


mosra:
Why `BasisImporterEtc2` and not `BasisImporter`? Concrete format selection could be done either with

```cpp
manager.setPreferredPlugins("BasisImporter", {"BasisImporterEtc2"});
```

or by modifying global configuration for the `BasisImporter` plugin:

```cpp
manager.metadata("BasisImporter")->configuration().setValue("format", "Etc2");
```

Squareys:
The issue is that `BasisImporter` has `Etc2` as default, but a hypothetical `TinyBasisImporter` could use `Etc1`. So the user would get a different `->compressedFormat()` depending on which plugins he has installed :thinking:

mosra:
> Hypothetical

Yes. But even in that case, both defaults are equally arbitrary and probably wrong, which leads me to think...

> default

What about ... not having a default at all? Like, the plugin failing to load a file when you forget to give it a default? That seems like the best way to avoid errors. Things that deal with AnyImageImporter need to handle loading errors anyway so this wouldn't introduce anything unexpected, and if the code wants to deal with basis files, it needs to set those defaults properly *in advance*, otherwise `image2D()` always gives `NullOpt`.

Squareys:
That is actually an awesome idea, as it will nudge people to read the docs for how to configure and make a probably way more suited decision than "use default Etc2". Will make that change to BasisImporter

src/MagnumPlugins/AnyImageConverter/AnyImageConverter.h
```diff
@@ -68,18 +68,18 @@ Supported formats for uncompressed data:
...
 -   Radiance HDR (`*.hdr`), converted with any plugin that provides
     `HdrImageConverter`
--   JPEG (`*.jpg`, `*.jpe`, `*.jpeg`), converted with @ref JpegImageConverter
-    or any other plugin that provides it
 -   Portable Network Graphics (`*.png`), converted with @ref PngImageConverter
     or any other plugin that provides it
 -   Truevision TGA (`*.tga`, `*.vda`, `*.icb`, `*.vst`), converted with
     @ref TgaImageConverter or any other plugin that provides it
+-   Windows Bitmap (`*.bmp`), converted with any plugin that provides
+    `BmpImageConverter`
```
mosra: Um, the ordering should be by the extension, not by the "marketing name", as that's what people look for. It was mostly okay except for bmp :)

src/MagnumPlugins/AnyImageImporter/AnyImageImporter.cpp
```diff
@@ -128,8 +130,11 @@ void AnyImageImporter::doOpenData(Containers::ArrayView<const char> data) {
     CORRADE_INTERNAL_ASSERT(manager());
 
     std::string plugin;
+    /* https://github.com/BinomialLLC/basis_universal/blob/7d784c728844c007d8c95d63231f7adcc0f65364/transcoder/basisu_file_headers.h#L78 */
+    if(Utility::String::viewBeginsWith(data, "sB\0\x16")
+        plugin = "BasisImporter";
     /* https://docs.microsoft.com/cs-cz/windows/desktop/direct3ddds/dx-graphics-dds-pguide */
-    if(Utility::String::viewBeginsWith(data, "DDS "))
+    else if(Utility::String::viewBeginsWith(data, "DDS "))
```
mosra: Care to add a test for this to the `detect()` test case? It's really simple, just pick the smallest basis file you have (don't forget to reference it in CMakeLists) and then add one line to the test instance data. It'll auto-check both the extension and magic comparison.

Squareys:
> Maybe I'll merge the converter part later, don't want people to complain that "it wants a BasisImageConverter, but I can't find it anywhere! where it is?" 😆

I doubt that anyone will stumble over this in the next couple of days... I'm planning to get https://github.com/mosra/magnum-plugins/pull/65 merge-ready by Sunday at the latest 😅 Depends on your schedule, though :P

src/MagnumPlugins/AnyImageImporter/AnyImageImporter.h
```diff
@@ -66,6 +66,7 @@ of the `Magnum` package and link to the `Magnum::AnyImageImporter` target. See
 
 Supported formats:
 
+-   Basis Universal (`*.basis`), loaded with any plugin that provides `BasisImporter`
```
mosra: Should say `with @ref BasisImporter or any other plugin that provides it` instead ... for `BmpImporter` etc there's no dedicated plugin so nothing to link to, but here (and in case of `PngImporter`, `DdsImporter`, ...) it is.

Same for the `BasisImageConverter` (and then I can't merge that part until the converter PR is merged otherwise the docs won't build cleanly).

src/MagnumPlugins/AnyImageImporter/Test/AnyImageImporterTest.cpp
```diff
@@ -87,7 +87,8 @@ constexpr struct {
     {"BMP", "image.bmp", nullptr, "BmpImporter"},
     {"GIF", "image.gif", nullptr, "GifImporter"},
     {"PSD", "image.psd", nullptr, "PsdImporter"},
-    {"TIFF", "image.tiff", nullptr, "TiffImporter"}
+    {"TIFF", "image.tiff", nullptr, "TiffImporter"},
+    {"BASIS", "rgb.basis", nullptr, "BasisImporter"}
```
mosra: `Basis`? It's not an acronym I think :)

src/MagnumPlugins/AnyImageImporter/AnyImageImporter.h
Squareys: I'm fine with getting this ready to merge and then you can merge it later with the converter 👍

mosra:
Merged as a1c2c9c6b3ab2eeaaa90c665bb486acf0c2cc4ea and b4ca71ef9bda2134bb08deb8a8a3227b2f70d1bd, thank you!

mosra:
Ugh why is it that I always discover bugs *after* things get pushed to master -- the file signature is just `sB`, the bytes after are just a version that changes all the time (in particular, files generated with current `basisu` have some other version than the bundled `kodim20.basis` etc. in the Basis repo, so those won't load).

Fix pushed as 64e365eeee3dc14741e0e54bd1b8da74917e16b6.

PR #371, opened 2019-09-13
Mesh Data rework

mosra:
Current state:

image

WIP docs: https://doc.magnum.graphics/magnum-meshdata/namespaceMagnum_1_1Trade.html (of interest is the `MeshData` class and `DataChunkHeader` describing the new `*.blob` file format)

What will this do / things to do:

- [x] Combine `MeshData2D` and `MeshData3D` into a single type, blurring the difference between these two (similarly to how `AnimationData` got done before), reducing code duplication
    - [x] Need to deprecate the two original types and the old `compile()`
    - [x] Adapt all code that consumes those to new APIs
    - [x] Add implicit conversion for seamless backward compatibility
- [x] Expose new unified Importer API for those
    - [x] Need to deprecate related Importer APIs (and make use of the implicit conversion for seamless backward compatibility)
- [x] Create a ~~`MeshAttributeType`~~ `VertexFormat` enum, expanding the current state with packed / normalized integer formats
    - [x] Provide mapping to `GL::DynamicAttribute`
    - [x] Provide an ability to store implementation-specific values in it, like with `PixelFormat`, restrict those in convenience `MeshData` APIs
    - [x] Provide mapping to `VkType`
    - [x] Document mapping to D3D and Metal formats, like with `PixelFormat`
- [ ] Create a new ~~`AbstractMeshConverter` plugin interface for operating on
    the `MeshData` ... or~~ `AbstractSceneConverter` for merging / splitting meshes etc <kbd>in progress</kbd>
    - for meshoptimizer etc
    - most algorithms need to operate on the whole scene tree, mesh conversion alone has little use
- [ ] Create a new ~~`magnum-meshconverter`~~ `magnum-sceneconverter` utility that would operate in-place on imported data (needs #240 as well, tho) <kbd>in progress</kbd>
- [x] Better support for generic vertex attributes (reserved IDs in the enum for those)
    - [x] `AbstractImporter::meshAttributeFor(const std::string&)`
    - [x] One-byte `MeshAttributeName` is not nearly enough considering all possible new attribs for instancing, per-face/edge/instance/meshlet attribs etc, expand to two bytes
- [ ] Meshlet / mesh shader support
    - [ ] ~~Two new views on the indexData, one a "running offset" saying how many elements is per a meshlet, second picking vertices from the global vertex array for given meshlet; the original index array is then for each meshlet telling which *local* vertices to pick (so it's probably just 8-bit always); how to name them? `meshletOffsets()` & `meshletIndices()`?~~ instead a completely new MeshData instance containing per-instance and two new `MeshPrimitive::TrianglesMeshlet` and `MeshPrimitive::Meshlet` primitives distinguishing the two from each other
    - [x] ~~Make it possible to have just `meshletOffsets()` w/o `meshletIndices()`, then `indices()` is global again; this can also be abused to group triangles into N-gons (but can't have both because a N-gon could be split across two meshlets and that would complicate things), preserve PLY/OBJ per-face attribs etc.~~ simply another MeshData instance with a `MeshPrimitive::Face`, `MeshPrimitive::Edge` etc primitive type to distinguish them from each other
    - [x] Ability to import these "additional mesh data layers" via a second parameter to `AbstractImporter::mesh()`, similarly to #369 
    - [ ] Some `MeshTools` algo to combine the global and local index arrays back to a classic index array
    - [x] Ability to have per-meshlet / per-Ngon / per-instance attributes (for example a normal / base texel offset for each, allowing the per-vertex ones to be less precise; [a per-meshlet bounding box etc.](https://github.com/zeux/meshoptimizer/blob/8cae35792da51e29c1cf18469d6912a956e40d99/src/meshoptimizer.h#L317-L331)) -- this would mean there's e.g. two `Normal` attributes, one per-vertex, another per-meshlet. How to distinguish these? ~~A per-attribute flag?~~ -- by putting them into one of `MeshPrimitive::Meshlet` or `MeshPrimitive::Instance` MeshData instances
    - [x] Since some per-meshlet attributes coming from meshoptimizer are actually arrays, provide a way to store and retrieve those (e.g. via `attribute<UnsignedByte[]>()` returning a 2D strided view, querying the array size with `attributeArraySize()`)
    - [x] Since meshlet meshes will be *very* different from normal ones, these need to be *designed* to prevent a meshlet one being used as a normal one (and then using a wrong local index array, wrong local attributes etc.). ~~E.g. instead of a binary `isIndexed()` have a `type()` returning `NonIndexed`, `Indexed`, `MeshletIndexed`, `MeshletIndexedMeshletAttributes`, `IndexedMeshletAttributes`, ...? UGH?~~ `MeshPrimitive::TrianglesMeshlet` / `MeshPrimitive::LinesMeshlet` / `MeshPrimitive::PointsMeshlet` instead of `MeshPrimitive::Triangles` etc.
    - [x] ~~Do meshlets actually support anything else than triangles? The NV introduction didn't seem to imply that. If so, we need `*Meshlet` variants of other primitive types.~~ yes they do
    - [x] ~~How does meshoptimizer fit into this design? Anything to get inspiration from?~~ Fits nicely.
- [x] Might need to implement growable `Containers::Array` to avoid this being too painful for hte importers
    - [x] ~~and some "pack this together, but preserve alignment" utility~~ now there's `MeshTools::interleavedLayout()` which is :heart: 
- [x] `interleave()` that eats and produces MeshData, additionally taking either bytes to reserve for each vertex or a list of `Trade::MeshAttributeData` to mix in
- [x] `duplicate()` that takes MeshData
- [x] `compile()` working with the new APIs 
    - [x] Also the normal generation
    - [x] Ability to pass it a "global" index/vertex buffer for the case where multiple meshes share one data buffer
- [x] `MeshTools::combineAttributes()` that replaces the STL-based insanity in `combineIndexedArrays()` 
  - [x] Needs an allocation-less variant of `removeDuplicates()`
  - [x] Deprecate `combineIndex*Arrays()`
- [x] `MeshTools::concatenate()` that takes a list of MeshData, assumes the same primitive, and combines them together (picking attributes of the first mesh, leaving zeros when the others don't have them, ignoring additional ones)
- [x] Make STL-less alternatives of all MeshTools 
  - [x] deprecate the originals
  - [x] Change `flipNormals()`, `tipsify()` to operate on array views instead
- [x] an option for the MeshAttributeData / MeshIndexData to store relative offsets to help with serialization and cases where layout is known but pointer not
    - [x] Ability to omit vertex count in there also? ~~Would need 6 new constructors :(~~
    - [x] Make all attribute layouts in Primitives done at compile-time
- [x] Store the StridedArrayView unpacked to save 10 bytes (by using less bits for stride and vertex count)
- [x] Do the essential updates needed in MeshTools so we can avoid vectors
- [x] Port Primitives to it
    - [x] Adapt tests
    - [x] Adapt the primitive generator utility in doc/generated
    - [x] MSVC 2015 has some weirdness where it duplicates constexpr data leading to runtime asserts about views being contained in the data
     - [x] MSVC asserts in MeshDataTest where it should be a graceful assert
- [x] Port `compressIndices()` away from `std::vector`
  - [x] Add a version operating on `MeshData` (reduce the type size + add vertex offset)
- [x] Port plugins
  - [x] ObjImporter
     - [x] Needs an integer version of `removeDuplicates()` (I think a better variant than the overcomplicated `combineIndexedArrays()`?)
        - [x] And shouldn't compress the indices, leave that on the user (needs non-allocating `removeDuplicates()`)
    - [x] Adapt tests
  - [x] AssimpImporter (we'll need to interleave outselves, slow ugh)
  - [x] OgexImporter (just floats, none of the extra types, it's a text file so who cares), 
  - [x] StanfordImporter
    - [x] needs Vector2us etc types, need to add them all, ugh
    - [x] Adapt test data, test for new restrictions
    - [x] Per-face attribute import, optionally turning into per-vertex
      - [x] Needs an integer version of `MeshTools::removeDuplicates()`
      - [x] Needs `MeshTools::combineFaceAttributes()`
        - [x] ... which needs an allocation-less not-in-place variant of `removeDuplicates()`
  - [x] TinyGltfImporter (should be pretty straightforward)
- [ ] Allow `MeshAttributeData` to have views with negative/zero stride to allow more flexibility in MeshTools, but ban those for MeshData itself
  - [ ] Allow `MeshIndexData` to have strided views (and negative/zero stride) to allow more flex ... uh is there any mesh tool that needs this?
- [x] Adapt examples
- [x] Adapt Python bindings
- [x] Adapt DART integration
- [x] Proofread docs
  - [x] Ensure the old types are not referenced anymore
- [ ] Serialization basics (see below) <kbd>in progress</kbd>
- [x] Interleave fails for non-aligned data on Emscripten even though it shouldn't (I thought `copy()` was doing the right thing there? investigate & test!)

### Serialization

Basically making the `MeshData` (and later other `*Data` as well) easily serializable, allowing projects to convert arbitrary mesh formats to blobs that could be loaded directly via mmaping:

```sh
magnum-sceneconverter file.blend --mesh "chair" --serialize chair.blob
magnum-sceneconverter scene.glb --mesh "tree" --serialize tree.blob
...

cat chair.blob tree.blob car.blob > blobs.blob # because why not
```

```cpp
// Takes basically no time
auto blob = Utility::Directory::mapRead("blobs.blob");

// (yes really) and this also
const Trade::MeshData& chair = *reinterpret_cast<const Trade::MeshData*>(blob.data());

// tho this is safer as it does a bunch of extra checks
const Trade::MeshData* chair = Trade::MeshData::from(blob);
```

Key points:

-   Each `*Data` would contain the same header with some magic, version info and total size. Magic and version info for sanity checks, total size to be able to simply `cat` multiple blobs together without having to worry about metadata. Similar to [RIFF](https://en.wikipedia.org/wiki/Resource_Interchange_File_Format) chunks (~~can't be 100% compatible because it's limited to 4 GB max~~ RIFF64?, can look there for inspiration).

    ```cpp
    struct DataHeader {
        // some magic and version info (ideally stolen from PNG to detect broken encodings, CRLF conversion, bad endiannes etc.)
        DataType type;
        uint64_t size;
        // ... more stuffs

        const DataHeader* nextData() const {
            return reinterpret_cast<const char*>(this) + size;
        }
    };

    struct Trade::MeshData: DataHeader { ... };

    ...
    Containers::ArrayView<const char> data;
    for(auto i = static_cast<const DataHeader*>(data); i < data.end(); i = i->nextData()) {
        if(i->type == DataType::MeshData) {
            const auto& mesh = static_cast<const Trade::MeshData&>(*i);
            ...
        } else if(i-> type == DataType::AnimationData) {
            ...
        } else continue;
    }
    ```

    The concatenated blob can then be iterated like above, ignoring unknown chunks.
-   The `MeshData` need to support both self-contained operation (data bundled internally in an `Array`) and the above-shown "reinterpreted" operation. ~~Because in the reinterpreted case no destructors are called, this could be done by putting a tag into the `Array` deleter field --- ideally with an odd value to minimize the chance an actual function could have the same address (I assume functions are never on odd addresses). Both `M` and `!` satisfy this condition so the following would be an odd address on both BE and LE systems~~ Instead make the Arrays `nullptr` with size, which will make the default deleter no-op (in case someone actually manages to call the destructor) and all offset calculation working correctly as well -- when returning pointers, the actual memory location would be added to the offset.
-   ~~In order to avoid extra work, the data layout should be consistent between 32bit and 64bit systems -- otherwise it won't be possible to serialize data on a (64bit) desktop and use them on a (32bit) Emscripten, e.g.. `Array` is 12 bytes on 32bit and 24 bytes on 64bit (pointer, size and deleter) so I think it could be enough to add padding around it on 32bit systems; plus the above "deleter tag" logic needs to be extended to check for both a 32bit and a 64bit case.~~ Too complicated, the serialization code needs to be able to create 32bit, 64bit, LE and BE variants, "reinterpret" deserialization will work only with data made for given platform. (But there needs to be some way to decode data from different platforms also...)
-    Some assets might have one huge buffer and particular meshes being just views on it. Ideally the buffer should be uploaded as-is in a single piece, with meshes then referring subranges. In this case, the serialized MeshData need to have a way to reference an external "data chunk" somehow -- some flag saying the offset not internal but to a different chunk + a data chunk "ID"? Or having that implicit -- it'll always be the first data chunk after a couple of data-less MeshData chunks?

Discussion points:

- make it possible to record a command-line used to produce a particular chunk? (e.g. [as proposed in gcc](https://www.phoronix.com/scan.php?page=news_item&px=GCC-Record-Command-Line-Info)) ... or other comments? might be useful for repeatability


src/Magnum/Trade/MeshData.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+MeshData::MeshData(MeshData&&) noexcept = default;
+
+MeshData& MeshData::operator=(MeshData&&) noexcept = default;
+
+UnsignedInt MeshData::indexCount() const {
+    switch(_indices.type) {
+        case MeshIndexType::UnsignedByte: return _indices.data.size();
+        case MeshIndexType::UnsignedShort: return _indices.data.size()/2;
+        case MeshIndexType::UnsignedInt: return _indices.data.size()/4;
```
Squareys: How about `return _indices.data.size()/indexSize(_indices.type);` ? Maybe that can be used for the `indices()` function also and removes those switches...

If the MeshIndexType enum coincidentally would be it's own length, i.e. Byte = 1, Short = 2, Int = 4, this would be only a cast even

src/Magnum/Trade/MeshData.cpp
```diff
@@ -0,0 +1,150 @@
...
+namespace {
+    template<class T> Containers::Array<UnsignedInt> convertIndices(const Containers::ArrayView<const char> data) {
+        const auto input = Containers::arrayCast<const T>(data);
+        Containers::Array<UnsignedInt> output{input.size()};
+        for(std::size_t i = 0; i != input.size(); ++i) output[i] = input[i];
+        return output;
+    }
+}
+
+Containers::Array<UnsignedInt> MeshData::indices() const {
```
Squareys: I wonder if there could be additional API to pass the index Data as is without conversion. I think GL doesn't need Integers, right? (Ah, I see `releaseIndices` is just the thing 👍 ) 

That would be a first step to allowing the loading of pre-`MeshTools::compile(...)`-ed meshes :)

src/Magnum/Trade/MeshData.h
```diff
@@ -0,0 +1,558 @@
...
+#include "Magnum/Trade/visibility.h"
+
+namespace Magnum { namespace Trade {
+
+/**
+@brief Mesh attribute type
+
+@see @ref MeshData, @ref MeshAttributeData, @ref MeshAttributeName
+*/
+enum class MeshAttributeType: UnsignedByte {
```
Squareys: Scalars?

src/Magnum/Trade/MeshData.cpp
mosra: Oh! Cool optimization. Though, a bit dangerous and undetectable when people pass garbage there :)

src/Magnum/Trade/MeshData.cpp
mosra: `indices<T>()` in the header is what you're looking for -- and together with the equivalent functionality for attribute data this is the *main* reason why I'm doing this rework, otherwise zero-copy import wouldn't be possible ;) 

This is just a convenience function for cases where you don't want to handle all three types.

src/Magnum/Trade/MeshData.h
mosra: Yup, definitely. This is an early version that matches the original `MeshData` feature set, it'll get extended after -- among other things also tangents, vertex weights, per-verted "object ID" and such.

Squareys:
How about making `DataType` a `char[4]` so that it can be used as "magic"?

E.g. `"Mg1m"` could be "Magnum, version 1, mesh", or `"Mg1a"` etc. Would a, make the blobs distinguishable and easily identify magnum files as such to help with detection in Any*Importer.

There could also be a file-header in addition to the blob header which suggests number of chunks making detection of broken chungs easier etc.

mosra:
> How about making DataType a char[4] so that it can be used as "magic"?

Ah, forgot to list the "magic" field in there. Updated. Putting all that into just four bytes seems a bit too crowded, would prevent extensibility with 3rd party chunk types for example.

> There could also be a file-header in addition to the blob header which suggests number of chunks 

That's what I explicitly do not want, as then the `cat` would be impossible and you'd need some tool to put them together, extract, or traverse. With the way above, a file is correct if each chunk points to a (non-OOB) position of the next chunk with correct magic or EOF and I think that could be enough :) Error detection could be further improved by having a CRC field for every chunk for example ... but then why not just zip all that anyway :)

mosra:
Merged in current state as the design is stable enough. Zero-copy import, mesh serialization, meshlet support and scene conversion plugin APIs are continued in #240 and #427.

PR #383, opened 2019-10-17
Plaftorm: Add cursor management support

Melix19:
Platform support:
- [x] ~AndroidApplication~ It's rarely used
- [x] GlfwApplication
- [x] Sdl2Application
- [x] ~GlxApplication~
- [ ] ~EmscriptenApplication~ It'll be made by mosra
- [x] ~XEglApplication~

The main usage for this (that is the main reason on why I implemented this) is for the ImGuiIntegration, because right now when hovering the edges of the windows/other draggable elements the cursor doesn't change due to the lack of this feature.

I'm not entirely sure about how good it's implemented, e.g. the way I constructed the array using the last enum element to get the size, and the whole switch-case that I needed to make (yeah, basically everything).

Also, I was thinking about adding the cursor creation in the Application constructor as well, using the Configuration class, but even there I was unsure if it was a good idea or not.

### DIFF
--- a/src/Magnum/Platform/GlfwApplication.h
+++ b/src/Magnum/Platform/GlfwApplication.h
@@ -487,6 +487,33 @@ class GlfwApplication {
         /** @{ @name Mouse handling */
 
     public:
+        /**
+         * @brief Cursor type
+         *
+         * @see @ref setCursor()
+         */
+        enum class Cursor : UnsignedInt {
+            Arrow,          /**< Arrow */
+            TextInput,      /**< Text input */
+            Crosshair,      /**< Crosshair */
+            ResizeWE,       /**< Double arrow pointing west and east */
+            ResizeNS,       /**< Double arrow pointing north and south */
+            Hand,           /**< Hand */
+            Hidden,         /**< Hidden */
+            HiddenLocked    /**< Hidden and locked */
+        };
+
+    public:
+        /**
+         * @brief Set the cursor to the @p type
+         */
+        void setCursor(Cursor cursor);
+
+        /**
+         * @brief Get the current cursor type
+         */
+        Cursor cursor();
+
         /** @brief Warp mouse cursor to given coordinates */
         void warpCursor(const Vector2i& position) {
             glfwSetCursorPos(_window, Double(position.x()), Double(position.y()));
@@ -576,6 +603,9 @@ class GlfwApplication {
         CORRADE_ENUMSET_FRIEND_OPERATORS(Flags)
 
         void setupCallbacks();
+
+        GLFWcursor* _cursors[8]{};
+        Cursor _cursor = Cursor::Arrow;
 
         /* These are saved from command-line arguments */
         bool _verboseLog{};
--- a/src/Magnum/Platform/Sdl2Application.cpp
+++ b/src/Magnum/Platform/Sdl2Application.cpp
@@ -685,6 +685,11 @@ Sdl2Application::~Sdl2Application() {
     #endif
 
     #ifndef CORRADE_TARGET_EMSCRIPTEN
+    for(auto& cursor: _cursors)
+        SDL_FreeCursor(cursor);
+    #endif
+
+    #ifndef CORRADE_TARGET_EMSCRIPTEN
     SDL_DestroyWindow(_window);
     #endif
     SDL_Quit();
@@ -876,6 +881,64 @@ void Sdl2Application::mainLoopIteration() {
     if(_flags & Flag::NoTickEvent) SDL_WaitEvent(nullptr);
     #endif
 }
+
+#ifndef CORRADE_TARGET_EMSCRIPTEN
+void Sdl2Application::setCursor(Cursor cursor) {
+    CORRADE_INTERNAL_ASSERT(UnsignedInt(cursor) < Containers::arraySize(_cursors));
+
+    if(cursor == Cursor::Hidden) {
+        SDL_ShowCursor(SDL_DISABLE);
+        SDL_SetWindowGrab(_window, SDL_FALSE);
+        SDL_SetRelativeMouseMode(SDL_FALSE);
+        return;
+    } else if(cursor == Cursor::HiddenLocked) {
+        SDL_SetWindowGrab(_window, SDL_TRUE);
+        SDL_SetRelativeMouseMode(SDL_TRUE);
+        return;
+    } else {
+        SDL_ShowCursor(SDL_ENABLE);
+        SDL_SetWindowGrab(_window, SDL_FALSE);
+        SDL_SetRelativeMouseMode(SDL_FALSE);
+    }
+
+    if(!_cursors[UnsignedInt(cursor)]) {
+        constexpr SDL_SystemCursor CursorMap[] {
+            SDL_SYSTEM_CURSOR_ARROW,
+            SDL_SYSTEM_CURSOR_IBEAM,
+            SDL_SYSTEM_CURSOR_WAIT,
+            SDL_SYSTEM_CURSOR_CROSSHAIR,
+            SDL_SYSTEM_CURSOR_WAITARROW,
+            SDL_SYSTEM_CURSOR_SIZENWSE,
+            SDL_SYSTEM_CURSOR_SIZENESW,
+            SDL_SYSTEM_CURSOR_SIZEWE,
+            SDL_SYSTEM_CURSOR_SIZENS,
+            SDL_SYSTEM_CURSOR_SIZEALL,
+            SDL_SYSTEM_CURSOR_NO,
+            SDL_SYSTEM_CURSOR_HAND
+        };
+
+        _cursors[UnsignedInt(cursor)] = SDL_CreateSystemCursor(CursorMap[UnsignedInt(cursor)]);
+    }
+
+    SDL_SetCursor(_cursors[UnsignedInt(cursor)]);
+}
+
+Sdl2Application::Cursor Sdl2Application::cursor() {
+    if(SDL_GetRelativeMouseMode())
+        return Cursor::HiddenLocked;
+    else if(!SDL_ShowCursor(SDL_QUERY))
+        return Cursor::Hidden;
+
+    SDL_Cursor* cursor = SDL_GetCursor();
+
+    if(cursor)
+        for(UnsignedInt i = 0; i < sizeof(_cursors); i++)
+            if(_cursors[i] == cursor)
+                return Cursor(i);
+
+    return Cursor::Arrow;
+}
+#endif
 
 void Sdl2Application::setMouseLocked(bool enabled) {
     /** @todo Implement this in Emscripten */
--- a/src/Magnum/Platform/Sdl2Application.h
+++ b/src/Magnum/Platform/Sdl2Application.h
@@ -830,6 +830,48 @@ class Sdl2Application {
         /** @{ @name Mouse handling */
 
     public:
+        #ifndef CORRADE_TARGET_EMSCRIPTEN
+        /**
+         * @brief Cursor type
+         *
+         * @see @ref setCursor()
+         */
+        enum class Cursor : UnsignedInt {
+            Arrow,          /**< Arrow */
+            TextInput,      /**< Text input */
+            Wait,           /**< Wait */
+            Crosshair,      /**< Crosshair */
+            WaitArrow,      /**< Small wait cursor */
+            ResizeNWSE,     /**< Double arrow pointing northwest and southeast */
+            ResizeNESW,     /**< Double arrow pointing northeast and southwest */
+            ResizeWE,       /**< Double arrow pointing west and east */
+            ResizeNS,       /**< Double arrow pointing north and south */
+            ResizeAll,      /**< Four pointed arrow pointing north, south, east, and west */
+            No,             /**< Slashed circle or crossbones */
+            Hand,           /**< Hand */
+            Hidden,         /**< Hidden */
+            HiddenLocked    /**< Hidden and locked */
+        };
+        #endif
+
+    public:
+        #ifndef CORRADE_TARGET_EMSCRIPTEN
+        /**
+         * @brief Set the cursor to the @p type
+         */
+        void setCursor(Cursor cursor);
+
+        /**
+         * @brief Get the current cursor type
+         */
+        Cursor cursor();
+        #endif
+
+        /** @brief Warp mouse cursor to given coordinates */
+        void warpCursor(const Vector2i& position) {
+            SDL_WarpMouseInWindow(_window, position.x(), position.y());
+        }
+
         /** @brief Whether mouse is locked */
         bool isMouseLocked() const { return SDL_GetRelativeMouseMode(); }
 
@@ -1014,6 +1056,10 @@ class Sdl2Application {
 
         typedef Containers::EnumSet<Flag> Flags;
         CORRADE_ENUMSET_FRIEND_OPERATORS(Flags)
+
+        #ifndef CORRADE_TARGET_EMSCRIPTEN
+        SDL_Cursor* _cursors[14]{};
+        #endif
 
         /* These are saved from command-line arguments */
         bool _verboseLog{};
--- a/src/Magnum/Platform/GlfwApplication.cpp
+++ b/src/Magnum/Platform/GlfwApplication.cpp
@@ -549,6 +549,8 @@ void GlfwApplication::setupCallbacks() {
 
 GlfwApplication::~GlfwApplication() {
     glfwDestroyWindow(_window);
+    for(auto& cursor: _cursors)
+        glfwDestroyCursor(cursor);
     glfwTerminate();
 }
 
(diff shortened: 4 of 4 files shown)


src/Magnum/Platform/GlfwApplication.h
```diff
@@ -436,6 +437,33 @@ class GlfwApplication {
...
 
+    public:
+        enum class Cursor : UnsignedInt {
+            Arrow,      /**< Arrow */
+            IBeam,      /**< I-Beam */
+            Crosshair,  /**< Crosshair */
+            ResizeWE,   /**< Double arrow pointing west and east */
+            ResizeNS,   /**< Double arrow pointing north and south */
+            Hand        /**< Hand */
+        };
```
mosra: Instead of having the switch, I think the values could go directly in here:

```cpp
Arrow = GLFW_ARROW_CURSOR,
IBeam = GLFW_IBEAM_CURSOR,
...
```

Same for SDL. All other Application enums (buttons, keys, ...) are done this way to avoid the extra switch overhead, so this would be then consistent :)

src/Magnum/Platform/GlfwApplication.h
```diff
@@ -577,6 +605,8 @@ class GlfwApplication {
 
         void setupCallbacks();
 
+        Containers::Array<GLFWcursor*> _cursors{Containers::ValueInit, static_cast<std::size_t>(Cursor::Hand) + 1};
```
mosra: The size is known at compile time, so this could be a plain C array -- `GLFWcursor* _cursors[6];` (and then you don't need the Array include).

And instead of taking the size from the enum (which would break once someone adds a value after the `Hand`), I'd put `CORRADE_INTERNAL_ASSERT(Int(cursor) < Containers::arraySize(_cursors))` in the setCursor / createCursor methods. That'll also guard it against things like `setCursor(Cursor(0xbad1dea))` :)

src/Magnum/Platform/GlfwApplication.h
```diff
@@ -436,6 +437,33 @@ class GlfwApplication {
...
+            Hand        /**< Hand */
+        };
+
+    public:
+        /**
+         * @brief Create a cursor that can then be used
+         * 
+         * @see @ref setCursor()
+         */
+        void createCursor(Cursor cursor);
```
mosra: What about having this function contained directly in `setCursor()` -- allocating a new cursor, if it's not there yet? I think that would be more user-friendly. Then you also wouldn't need anything in the `Configuration`.

src/Magnum/Platform/GlfwApplication.h
```diff
@@ -436,6 +437,33 @@ class GlfwApplication {
         /** @copydoc Sdl2Application::redraw() */
         void redraw() { _flags |= Flag::Redraw; }
 
+    public:
+        enum class Cursor : UnsignedInt {
+            Arrow,      /**< Arrow */
+            IBeam,      /**< I-Beam */
```
mosra: I had to think twice about what this means... CSS has it simply as `text`, let's use that instead? :)

Melix19:
Thanks for taking time to review this :D

> This of course depends on if you have Emscripten set up or not

Unfortunately I don't have Emscripten set up and I don't have really a clear idea on how to do it.

> Is it okay if this waits post-release?

Yeah, I'll be fine. I can always use the --Head, so no problem there.

mosra:
Okay, then I can do the Emscripten part myself, no worries :)

src/Magnum/Platform/GlfwApplication.h
Melix19: Well, initially I made this and it worked for SDL but not for glfw, the reason is that glfw stores these flags as #define 0x0003600x, so casting e.g. a Cursor::Arrow to UnsignedInt it didn't gave me a 0-based index like Sdl does so it was accessing unallocated memory.
But I think I can fix this doing `UnsignedInt index = 0x00036000 - std::static_cast<UnsignedInt>(Cursor::Arrow)` to get the index to save the value to the array

mosra: Ahhhhhh you're right, then the easy indexing into the array wouldn't work. Keep it like you have it now, then, and instead of a switch you could do

```cpp
constexpr Int CursorMap[] {
    GLFW_ARROW_CURSOR,
    GLFW_IBEAM_CURSOR,
    ...
};
```

and then do the conversion as `CursorMap[UnsignedInt(cursor)]`. I'm [doing a similar thing](https://blog.magnum.graphics/backstage/gfx-api-enum-mapping/) for the PixelFormat enum and conversion to GL/Vulkan.

mosra: Dang, too late to reply :sweat_smile:

Melix19: > Dang, too late to reply 😅

Literally when I clicked Enter to push the changes I heard the *bling* that notified me of your comment :D

Btw I'll change it again, no problem.

Melix19:
Well, I think that I made everything... almost.

I discovered that imgui also wants to show/hide the cursor as he needs to, so maybe I can do a showCursor() and hideCursor() as well

mosra:
Is that orthogonal to the cursor type? Since it could be done via `setCursor(Cursor::None)` and then show again as `setCursor(Cursor::WhateverWasThereBefore)`.

Btw., I spotted some cursor-related APIs in GlfwApplication, isn't this hiding/showing there already?

Melix19:
> Is that orthogonal to the cursor type?

Yeah, I can make it relative to the type adding a Cursor::Hidden (and a Cursor::Disabled for glfw that has this extra option that can hide the cursor and enable unlimited mouse movement).

> I spotted some cursor-related APIs in GlfwApplication, isn't this hiding/showing there already?

Yep, there is already a cursorMode() inside of GlfwApplication::Configuration, but you can only change the mode on creation and not in real time.

mosra:
Hmm. SDL has that as [setMouseLocked()](https://doc.magnum.graphics/magnum/classMagnum_1_1Platform_1_1Sdl2Application.html#add8dcb7f6d2a04afc25dc88f4761a1d8), fortunately that's not just on creation. (Having only a possibility to do that once in Configuration for GLFW seems kinda useless, heh.)

If you want to implement this too, what about the following, to have it consistent for both:

- have `Cursor::Hidden` that hides the cursor (and, I don't know, the GL app would draw its own -- laggy one -- for example)
- have `Cursor::HiddenLocked` (`Disabled` doesn't feel like a descriptive name to me) that on GLFW sets this different type, and on SDL hides the cursor calls `SDL_SetRelativeMouseMode()`

And with this, I think it'd be good to have a `cursor()` getter as well -- so the app can query the state again (and in case of SDL, it would call `SDL_GetRelativeMouseMode()` to differentiate between `Hidden` and `HiddenLocked`).

Then the `setMouseLocked()` / `isMouseLocked()` and GLFW's `Configuration::CursorMode` could get deprecated in favor of this unified API. (I can do all that labelling and documenting.)

Melix19:
SDL_SetRelativeMouseMode() is not exactly identical to glfw's Disabled state: glfw stucks the cursor on it's position and make it invisible (even moving the mouse, the cursor stays hidden and at the same position); SDL_SetRelativeMouseMode() makes the cursor position only be relative to the window, so it can be moved and it will interact with everything that's in the window.
I know that it's not a major difference but we should't consider this unified if this behavior is slightly different.

The closest we can get is using SDL_SetWindowGrab, it locks the cursor on it's current position as glfw's Disabled state does (or in the top left of the windows if the cursor was outside of the window when the function is called, instead of the center of the screen that glfw uses as default, but I think that this is an insignificant difference).

Melix19:
Oh, ok, nevermind, actually the Disabled state is the same as the SDL_SetRelativeMouseMode, but if it's called multiple time (I was testing it in an ImGui context) the cursor gets like resetted to the last position, but with SDL_SetRelativeMouseMode actually is not and it works well

mosra:
If you look at [setMouseLocked()](https://github.com/mosra/magnum/blob/6ebe264e611caa23eeb6fc3f4fc57aa095353344/src/Magnum/Platform/Sdl2Application.cpp#L880-L889), it calls both `SDL_SetWindowGrab()` and `SDL_SetRelativeMouseMode()` ... but it's been a long time since I did those and I don't really remember what's the desired behavior and if/how GLFW matches that. Also can't really test / verify on anything locally, lately it's been all headless rendering and algorithms, no games at all :)

We could also postpone this part to later, if you don't feel like digging deep into this stuff.

Melix19:
> I don't really remember what's the desired behavior and if/how GLFW matches that

After having documented better I found the real differences:
- GLFW_CURSOR_DISABLED hides the cursor and make it to behave like a camera control, so it has unlimited space to be moved (not expected from Cursor::HiddenLocked).
- GLFW_CURSOR_HIDDEN just makes the cursor hidden when it's hovering the window, it does not affect the default cursor behavior (so it's good for Cursor::Hidden).
- SDL_SetRelativeMouseMode() makes the cursor hidden but it locks the cursor to the window borders, so the cursor can't go over them (expected from Cursor::HiddenLocked).

So the problem here is Glfw, in their documentation I haven't found a real alternative to SDL_SetRelativeMouseMode, so the only thing that we could do is to fake it if we really want to.

> We could also postpone this part to later, if you don't feel like digging deep into this stuff.

Maybe for now I could add a Cursor::HiddenLocked for sdl using the SDL_SetRelativeMouseMode() and a Cursor::HiddenUnlimited (?) using the GLFW_CURSOR_DISABLED?

mosra:
Isn't then `GLFW_CURSOR_DISABLED` the same as `SDL_SetRelativeMouseMode()` together with `SDL_SetWindowGrab()`? That's what I'd expect from `HiddenLocked` -- cursor hidden, unlimited movement, never leaving the window (i.e., a FPS camera). I can't see a case where I would want a  (hidden) cursor that gets stuck on window edges instead of letting me move freely.

Melix19:
> That's what I'd expect from HiddenLocked -- cursor hidden, unlimited movement, never leaving the window

That's what I'd expect either from HiddenLocked, and it's what setMouseLocked() currently does so it's correct. GLFW_CURSOR_DISABLED has unlimited movement as well, but they do this as the absolute position of the cursor is unlimited and not limited to the window.

E.g. with setMouseLocked() using the MouseMoveEvent::position(), you'll get values within the window size, but with GLFW_CURSOR_DISABLED you'll get an absolute position of like if the cursor is moving outside of the window and even outside the screen. This is what I meant.

mosra:
`SDL_SetRelativeMouseMode()` doc says that

> Only relative motion events will be delivered, the mouse position will not change.

Which, as far as I understand, means `MouseMoveEvent::position()` is always the same (probably keeping the last coordinate from the time when relative mode was requested), and thus ~useless -- one should query `relativePosition()` instead.

Then, as of 89d4a754d53d02d02839b5b44ee3836ae8c9ff7f, `GlfwApplication` supports `relativePosition()` as well, which is currently calculated on the application side as a delta to previous move event position. Thus `relativePosition()` should behave the same for both (while `position()` not). I think that's consistent enough (as long as the behavior is documented, suggesting users to query `relativePosition()`) -- and if that's not enough for very specific use cases, one can always call the underlying GLFW/SDL APIs directly.

Melix19:
Ok then, now it should be good to go!

> Which, as far as I understand, means MouseMoveEvent::position() is always the same (probably keeping the last coordinate from the time when relative mode was requested)

Well, you're right, reading the description I understand that too, but actually, from my test, the absolute position actually changes, but remains within the size of the window (at least on macOS)
so maybe we should document that better than Sdl does :D

src/Magnum/Platform/Sdl2Application.cpp
```diff
@@ -275,6 +275,52 @@ void Sdl2Application::setWindowTitle(const std::string& title) {
...
+        };
+
+        _cursors[UnsignedInt(cursor)] = SDL_CreateSystemCursor(CursorMap[UnsignedInt(cursor)]);
+    }
+
+    SDL_SetCursor(_cursors[UnsignedInt(cursor)]);
+}
+
+Sdl2Application::Cursor Sdl2Application::cursor() {
+    return _cursor;
```
mosra: Instead of using a variable, this could be a query to `SDL_GetCursor()` (and `SDL_GetRelativeMouseMode()`), I think? Especially considering the interaction with the to-be-deprecated `setMouseLocked()` APIs that could enable relative mouse mode independently, the getter should return the real state and not what it *thinks* is active. This would just iterate over the `_cursor` array, trying to find an item that matches (and then, if nothing matches, `CORRADE_ASSERT_UNREACHABLE()`, at least for now).

Tried to find an equivalent getter for GLFW, but there doesn't seems to be (unless I overlooked some general `glfwGetValue()`), so there the variable is okay I guess.

src/Magnum/Platform/Sdl2Application.cpp
```diff
@@ -275,6 +275,52 @@ void Sdl2Application::setWindowTitle(const std::string& title) {
...
 
+#ifndef CORRADE_TARGET_EMSCRIPTEN
+void Sdl2Application::setCursor(Cursor cursor) {
+    CORRADE_INTERNAL_ASSERT(UnsignedInt(cursor) < Containers::arraySize(_cursors));
+
+    _cursor = cursor;
+
+    if(cursor == Cursor::Hidden) {
+        SDL_ShowCursor(SDL_FALSE);
+        return;
```
mosra: If I set the cursor to `HiddenLocked` and then to `Hidden`, it would stay locked. So I guess here it should also do

```cpp
        SDL_SetWindowGrab(_window, SDL_FALSE);
        SDL_SetRelativeMouseMode(SDL_FALSE);
```

src/Magnum/Platform/GlfwApplication.h
```diff
@@ -436,6 +436,31 @@ class GlfwApplication {
         /** @copydoc Sdl2Application::redraw() */
         void redraw() { _flags |= Flag::Redraw; }
 
+    public:
+        enum class Cursor : UnsignedInt {
```
mosra: Document this enum as well? :) Same for SDL.

I'm usually adding a reference to the setter there, so people can jump from the enum directly to where it gets used.

src/Magnum/Platform/Sdl2Application.cpp
Melix19: True, that's better.

> (and then, if nothing matches, CORRADE_ASSERT_UNREACHABLE(), at least for now)

Wouldn't it better to return Cursor::Arrow as default? Because if you query the cursor without having created one it would just crash, but I think that it's obvious that the default one is always the Cursor::Arrow.
A minor downside is that if someone wants to manage the cursor by itself by calling internal SDL functions, the cursor() always returns Cursor::Arrow instead of saying "Hey, there's nothing here", but I think that this cloud just be specified in the documentation.

mosra: > Because if you query the cursor without having created one it would just crash

Good point. Yeah, do that then :) 

> if someone wants to manage the cursor by itself by calling internal SDL functions

Yeah, that was my primary concern, didn't realize it would crash also if the user didn't set any cursor yet.

Melix19:
> Last three minor things and then I think it's good to go as well

Done

mosra:
Thanks a lot! Do you have the ImGuiIntegration counterpart as well? :) So I have something to test with when I get to implementing the Emscripten side.

Melix19:
> Do you have the ImGuiIntegration counterpart as well?

I made a really bad version inside my application just for testing (so not actually in the ImGuiIntegration). But I have some concerns on doing it inside ImGuiIntegration, like, how could I manage to access the Application::Cursor without having to include the Sdl2Application? Because I was thinking to make another CursorMap to map the Application::Cursor to the imgui's cursor flags (taking note that imgui uses some cursors that sdl has but glfw doesn't have).

Or maybe I'm thinking it in a wrong way?

src/Magnum/Platform/Sdl2Application.h
```diff
@@ -1015,6 +1053,10 @@ class Sdl2Application {
         typedef Containers::EnumSet<Flag> Flags;
         CORRADE_ENUMSET_FRIEND_OPERATORS(Flags)
 
+        #ifndef CORRADE_TARGET_EMSCRIPTEN
+        SDL_Cursor* _cursors[14];
```
mosra:
```suggestion
        SDL_Cursor* _cursors[14]{};
```

(I got a crash otherwise. Same for GLFW.)

src/Magnum/Platform/GlfwApplication.h
```diff
@@ -436,6 +436,36 @@ class GlfwApplication {
...
+            Text,           /**< Text input */
+            Crosshair,      /**< Crosshair */
+            ResizeWE,       /**< Double arrow pointing west and east */
+            ResizeNS,       /**< Double arrow pointing north and south */
+            Hand,           /**< Hand */
+            Hidden,         /**< Hidden */
+            HiddenLocked    /**< Hidden and locked */
+        };
+
+        Cursor _cursor = Cursor::Arrow;
```
mosra: Put this next to the `_cursors` maybe? Doesn't make sense having it publicly.

Melix19:
> I thought it would make sense to move the cursor-related stuff into the "Mouse handling" sections

Yep, it makes more sense

> can you tell how does this API fit into the rest? Expose some equivalent on SDL side?

I think that it's good, even because there is [SDL_WarpMouseInWindow](https://wiki.libsdl.org/SDL_WarpMouseInWindow) for SDL, so I could just add a warpCursor() in there as well and improving the unified API.

mosra:
Ohh! So warp is "mouse goes out on one side but comes back on the other", right, not related to mouse locking or cursor hiding or that. Yup, agree on adding the equivalent on SDL side.

mosra:
Merged in 40bfa7ae9105bdeb60b85e8cccc26f75efa20541, sorry for the delay caused by all the release management stuff.

In addition I implemented Emscripten support in 04827d374376688924846d0b61f85660b5477a35 and c091ed0e66c2d55a5cc8a1a3b13ebd91c077104c (wow, CSS has 36 cursor types), plus deprecated the old, inconsistent and useless APIs in favor of your new ones in 8d9d247573d0e123501faa94f1dbe369115f89ed and c4e3e3c0457f912b1a625598ff253f8011b5636b.

Next I'm going to merge the ImGui side and upload an updated example. Thank you for this!

PR #387, opened 2019-11-07
Modular applications

costashatz:
This PR adds a `mainLoopIteration()` function to `AbstractXApplication` and `GlfwApplication` similar to the one that `Sdl2Application` has.


mosra:
Thank you!

PR #388, opened 2019-11-07
GlfwApplication: Workaround for DPI scaling

costashatz:
This PR offers a workaround for DPI scaling when monitor size is equal to zero in GlfwApplication..

### DIFF
--- a/src/Magnum/Platform/GlfwApplication.cpp
+++ b/src/Magnum/Platform/GlfwApplication.cpp
@@ -181,6 +181,10 @@ Vector2 GlfwApplication::dpiScaling(const Configuration& configuration) const {
         const GLFWvidmode* const mode = glfwGetVideoMode(monitor);
         Vector2i monitorSize;
         glfwGetMonitorPhysicalSize(monitor, &monitorSize.x(), &monitorSize.y());
+        if(monitorSize.isZero()) {
+            Warning{verbose} << "Platform::GlfwApplication: the physical monitor size is zero? DPI scaling won't be used";
+            return Vector2{1.0f};
+        }
         auto dpi = Vector2{Vector2i{mode->width, mode->height}*25.4f/Vector2{monitorSize}};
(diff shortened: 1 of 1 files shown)


mosra:
Merged as 411e3493583b5692aa3749b21e5b79e70f10055e, thank you!

mosra:
@costashatz looking at release notes of glfw 3.3.1, I see "zero monitor size" being fixed: https://www.glfw.org/changelog.html .. should we remove this workaround again? (I assume distros that already have 3.3 would get 3.3.1 soon, and those that are on 3.2 and older won't upgrade to 3.3.0 because that would make no sense.)

costashatz:
> @costashatz looking at release notes of glfw 3.3.1, I see "zero monitor size" being fixed: https://www.glfw.org/changelog.html .. should we remove this workaround again? (I assume distros that already have 3.3 would get 3.3.1 soon, and those that are on 3.2 and older won't upgrade to 3.3.0 because that would make no sense.)

@mosra sorry for the late reply! I was on vacation..

I will check whether on my system this was fixed. Because this was a singularity container and maybe the version freezes to 3.3..

Does it hurt to keep it there? I mean performance-wise.

mosra:
No worries :)

Performance-wise it doesn't hurt, but it's a no-longer-needed cruft that enlarges the surface for potential bugs. If you can confirm this is no longer needed, I'll remove it, otherwise I'll at least add a note that this workaround is obsolete since 3.3.1.

mosra:
This got reverted in 8e31882f6e922406868653bde90a895a3f1d674f, as the underlying X11 issue is fixed in GLFW 3.3.1 and a related Wayland issue is fixed in 3.3.3. Additionally, with virtual DPI scaling being now the default, this issue should be significantly less likely to happen even with old GLFW versions. In any case, if this would reappear, a workaround is to run the app with `--magnum-dpi-scaling 1` or the equivalent environment variable setting.

PR #392, opened 2019-11-13
Fix for Glfw and Wayland crashing

costashatz:
Fix for Glfw and Wayland crashing when showing the window after being hidden..

### DIFF
--- a/src/Magnum/Platform/GlfwApplication.cpp
+++ b/src/Magnum/Platform/GlfwApplication.cpp
@@ -406,9 +406,11 @@ bool GlfwApplication::tryCreate(const Configuration& configuration, const GLConf
        succeeds, make the context current so we can query GL_VENDOR below.
        If we are on Wayland, this is causing a segfault; a blinking window is
        acceptable in this case. */
-    constexpr static const char waylandString[] = "wayland";
+    constexpr const char waylandString[] = "wayland";
     if(std::strncmp(std::getenv("XDG_SESSION_TYPE"), waylandString, sizeof(waylandString)) != 0)
         glfwWindowHint(GLFW_VISIBLE, false);
+    else if(_verboseLog)
+        Warning{} << "Platform::GlfwApplication: Wayland detected, GL context has to be created with the window visible and may cause flicker on startup";
     if((_window = glfwCreateWindow(scaledWindowSize.x(), scaledWindowSize.y(), configuration.title().c_str(), monitor, nullptr)))
         glfwMakeContextCurrent(_window);
 
--- a/src/Magnum/Platform/GlfwApplication.cpp
+++ b/src/Magnum/Platform/GlfwApplication.cpp
@@ -403,8 +403,12 @@ bool GlfwApplication::tryCreate(const Configuration& configuration, const GLConf
 
     /* Create window. Hide it by default so we don't have distracting window
        blinking in case we have to destroy it again right away. If the creation
-       succeeds, make the context current so we can query GL_VENDOR below. */
-    glfwWindowHint(GLFW_VISIBLE, false);
+       succeeds, make the context current so we can query GL_VENDOR below.
+       If we are on Wayland, this is causing a segfault; a blinking window is
+       acceptable in this case. */
+    constexpr static const char waylandString[] = "wayland";
+    if(std::strncmp(std::getenv("XDG_SESSION_TYPE"), waylandString, sizeof(waylandString)) != 0)
+        glfwWindowHint(GLFW_VISIBLE, false);
     if((_window = glfwCreateWindow(scaledWindowSize.x(), scaledWindowSize.y(), configuration.title().c_str(), monitor, nullptr)))
         glfwMakeContextCurrent(_window);
 


src/Magnum/Platform/GlfwApplication.cpp
```diff
@@ -403,8 +403,12 @@ bool GlfwApplication::tryCreate(const Configuration& configuration, const GLConf
...
     /* Create window. Hide it by default so we don't have distracting window
        blinking in case we have to destroy it again right away. If the creation
-       succeeds, make the context current so we can query GL_VENDOR below. */
-    glfwWindowHint(GLFW_VISIBLE, false);
+       succeeds, make the context current so we can query GL_VENDOR below.
+       If we are on Wayland, this is causing a segfault; a blinking window is
+       acceptable in this case. */
+    constexpr static const char waylandString[] = "wayland";
+    if(std::strncmp(std::getenv("XDG_SESSION_TYPE"), waylandString, sizeof(waylandString)) != 0)
+        glfwWindowHint(GLFW_VISIBLE, false);
```
mosra: Since this will cause the window to flicker when the context is recreated, I think it would make sense to print a (verbose) warning message in this case. Something like this added after:

```cpp
else if(_verboseLog) 
    Warning{} << "Platform::GlfwApplication: Wayland detected, GL context has to be created with the window visible and may cause flicker on startup";
```

If you can craft a better / shorter message than me, please do :)

src/Magnum/Platform/GlfwApplication.cpp
```diff
@@ -403,8 +403,12 @@ bool GlfwApplication::tryCreate(const Configuration& configuration, const GLConf
 
     /* Create window. Hide it by default so we don't have distracting window
        blinking in case we have to destroy it again right away. If the creation
-       succeeds, make the context current so we can query GL_VENDOR below. */
-    glfwWindowHint(GLFW_VISIBLE, false);
+       succeeds, make the context current so we can query GL_VENDOR below.
+       If we are on Wayland, this is causing a segfault; a blinking window is
+       acceptable in this case. */
+    constexpr static const char waylandString[] = "wayland";
```
mosra:
```suggestion
    constexpr const char waylandString[] = "wayland";
```

I have this bad stigma of function-local statics being wrapped in mutexes and `std::call_once` stuff, so trying to not use it to avoid raising a red flag whenever I see the code :D If I'm right, `constexpr const char` is put into readonly memory anyway, so it shouldn't have any perf disadvantages.

mosra:
Argh! I just realized this would introduce a regression when you specify `WindowFlags::Hidden` in the configuration. Is it possible for you to test this patch on Wayland on top of the above (and `WindowFlags::Hidden` passed in the config)? The window should briefly show and then stay hidden if all is good.

```patch
diff --git a/src/Magnum/Platform/GlfwApplication.cpp b/src/Magnum/Platform/GlfwApplication.cpp
index f4be6fd75..4b40495b5 100644
--- a/src/Magnum/Platform/GlfwApplication.cpp
+++ b/src/Magnum/Platform/GlfwApplication.cpp
@@ -407,8 +407,8 @@ bool GlfwApplication::tryCreate(const Configuration& configuration, const GLConf
        If we are on Wayland, this is causing a segfault; a blinking window is
        acceptable in this case. */
     constexpr const char waylandString[] = "wayland";
-    if(std::strncmp(std::getenv("XDG_SESSION_TYPE"), waylandString, sizeof(waylandString)) != 0)
-        glfwWindowHint(GLFW_VISIBLE, false);
+    const bool onWayland = std::strncmp(std::getenv("XDG_SESSION_TYPE"), waylandString, sizeof(waylandString)) != 0;
+    if(!onWayland) glfwWindowHint(GLFW_VISIBLE, false);
     else if(_verboseLog)
         Warning{} << "Platform::GlfwApplication: Wayland detected, GL context has to be created with the window visible and may cause flicker on startup";
     if((_window = glfwCreateWindow(scaledWindowSize.x(), scaledWindowSize.y(), configuration.title().c_str(), monitor, nullptr)))
@@ -487,9 +487,13 @@ bool GlfwApplication::tryCreate(const Configuration& configuration, const GLConf
         _window = nullptr;
     }
 
-    /* Show the window once we are sure that everything is okay */
+    /* Show the window once we are sure that everything is okay (and it's not
+       requested to be hidden). If we didn't hide it initially due to applying
+       the Wayland workaround and it's requested to be, hide it now. */
     if(!(configuration.windowFlags() & Configuration::WindowFlag::Hidden))
         glfwShowWindow(_window);
+    else if(onWayland)
+        glfwHideWindow(_window);
 
     /* Return true if the initialization succeeds */
     return true;

```

Thanks a lot!

costashatz:
> Argh! I just realized this would introduce a regression when you specify `WindowFlag::Hidden` in the configuration. Is it possible for you to test this patch on Wayland on top of the above (and `WindowFlag::Hidden` passed in the config)? The window should briefly show and then stay hidden if all is good.

If the visibility is set to hidden (i.e., `glfwWindowHint(GLFW_VISIBLE, false);`) and we never call `glfwShowWindow()` then we do not get a segfault. I just verified that if I select `WindowFlag::Hidden` without any additional change in the code (this PR as is), it runs with no problem (of course no window appears).

mosra:
Merged as c8d2c33ac60a01c17183f7d7f7d5a6cf0d81fc5c (and without my patch above, since it was needed). Thank you!

mosra:
Just for the record, this workaround got reverted in a0011a9e39b260197f38a4d8d6b028dfd339172d as the underlying issue is fixed in GLFW 3.3.7.

PR #397, opened 2019-11-18
Add euler conversion from Quaternion

Melix19:
I think that this method could be helpful in some situation using magnum, so I decided to make a PR.
I took the algorithm from [QT](https://github.com/qt/qtbase/blob/84056cdca11c63ddb9fac3d95fcfa952adaa6ea1/src/gui/math3d/qquaternion.cpp#L507) (that it's based on another algorithm as specified in their comments) and adapted to magnum using a Vector3 as return.

It's obvious that this method should not be abused because of the problems that eulers can have (and I think that this type of conversion isn't even always 100% perfect), so maybe we could specify that in the doc?

### DIFF
--- a/src/Magnum/Math/Quaternion.h
+++ b/src/Magnum/Math/Quaternion.h
@@ -400,6 +400,19 @@ template<class T> class Quaternion {
         Matrix3x3<T> toMatrix() const;
 
         /**
+         * @brief Convert to an euler vector
+         *
+         * Expects that the quaternion is normalized and that the values
+         * are combined in ZYX order.
+         * The returned vector of radians is in XYZ order.
+         * @see @ref rotation(), @ref DualQuaternion::rotation()
+         * To convert the euler angles back to the Quaternion you can use:
+         *
+         * @snippet MagnumMath.cpp Quaternion-fromEuler
+         */
+        Vector3<Rad<T>> toEuler() const;
+
+        /**
          * @brief Negated quaternion
          *
          * @f[
@@ -731,6 +744,37 @@ template<class T> Matrix3x3<T> Quaternion<T>::toMatrix() const {
             2*_vector.y()*_vector.z() - 2*_vector.x()*_scalar,
                 T(1) - 2*pow2(_vector.x()) - 2*pow2(_vector.y()))
     };
+}
+
+/* Algorithm from:
+   https://github.com/mrdoob/three.js/blob/6892dd0aba1411d35c5e2b44dc6ff280b24d6aa2/src/math/Euler.js#L197 */
+template<class T> Vector3<Rad<T>> Quaternion<T>::toEuler() const {
+    CORRADE_ASSERT(isNormalized(),
+        "Math::Quaternion::toEuler():" << *this << "is not normalized", {});
+
+    Vector3<Rad<T>> euler{NoInit};
+
+    Matrix3x3<T> rotMatrix = toMatrix();
+
+    T m11 = rotMatrix[0][0];
+    T m12 = rotMatrix[0][1];
+    T m13 = rotMatrix[0][2];
+    T m21 = rotMatrix[1][0];
+    T m22 = rotMatrix[1][1];
+    T m23 = rotMatrix[1][2];
+    T m33 = rotMatrix[2][2];
+
+    euler.y() = Rad<T>(std::asin(-Math::min(Math::max(m13, T(-1.0)), T(1.0))));
+
+    if(!TypeTraits<T>::equalsZero(m13 - T(1.0), T(1.0))) {
+        euler.x() = Rad<T>(std::atan2(m23, m33));
+        euler.z() = Rad<T>(std::atan2(m12, m11));
+    } else {
+        euler.x() = Rad<T>(0.0);
+        euler.z() = Rad<T>(std::atan2(-m21, m22));
+    }
+
+    return euler;
 }
 
 template<class T> inline Quaternion<T> Quaternion<T>::operator*(const Quaternion<T>& other) const {
--- a/doc/snippets/MagnumMath.cpp
+++ b/doc/snippets/MagnumMath.cpp
@@ -988,6 +988,17 @@ Math::Matrix2x2<Byte> integral{floatingPoint}; // {{1, 2}, {-15, 7}}
 }
 
 {
+/* [Quaternion-fromEuler] */
+Rad x, y, z;
+Quaternion a =
+    Quaternion::rotation(z, Vector3::zAxis())*
+    Quaternion::rotation(y, Vector3::yAxis())*
+    Quaternion::rotation(x, Vector3::xAxis());
+/* [Quaternion-fromEuler] */
+static_cast<void>(a);
+}
+
+{
 /* [unpack-template-explicit] */
 // Literal type is (signed) char, but we assumed unsigned char, a != 1.0f
 Float a = Math::unpack<Float>('\xFF');
--- a/src/Magnum/Math/Test/QuaternionTest.cpp
+++ b/src/Magnum/Math/Test/QuaternionTest.cpp
@@ -96,6 +96,8 @@ struct QuaternionTest: Corrade::TestSuite::Tester {
     void angleNotNormalized();
     void matrix();
     void matrixNotOrthogonal();
+    void euler();
+    void eulerNotNormalized();
 
     void lerp();
     void lerp2D();
@@ -171,6 +173,8 @@ QuaternionTest::QuaternionTest() {
               &QuaternionTest::angleNotNormalized,
               &QuaternionTest::matrix,
               &QuaternionTest::matrixNotOrthogonal,
+              &QuaternionTest::euler,
+              &QuaternionTest::eulerNotNormalized,
 
               &QuaternionTest::lerp,
               &QuaternionTest::lerp2D,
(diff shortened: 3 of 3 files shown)


src/Magnum/Math/Quaternion.h
```diff
@@ -399,6 +399,13 @@ template<class T> class Quaternion {
          */
         Matrix3x3<T> toMatrix() const;
 
+        /**
+         * @brief Convert to an euler vector
+         *
+         * @see @ref rotation(), @ref DualQuaternion::rotation()
+         */
+        Vector3<T> toRotation() const;
```
mosra: `toEuler()` instead? So it's clear that you're going to the not-recommended API areas :) Can it return `Vector3<Rad<T>>`? Would be good to mention in which order the returned values are, in which order they should be combined and also a snippet that shows how to convert the Euler angles back:

```cpp
Rad x, y, z;
Quaternion q = /* I hope this is correct, please check! */
    Quaternion::rotate(z, Vector3::zAxis())*
    Quaternion::rotate(y, Vector3::yAxis())*
    Quaternion::rotate(x, Vector3::xAxis());
```

To avoid syntax error in the documentation, the docs have compiled snippets, so instead of placing it directly here what would be great is to extend `doc/snippets/MagnumMath.cpp` with a block delimited by `/* [Quaternion-toRotation] */` and reference it via `@snippet MagnumMath.cpp Quaternion-toRotation` (check other places, e.g. in Matrix4.h, to see how it's usually done). Of course if you don't feel like doing that, I can do that post-merge (but that'll delay the merge, of course ... I'm a bit tight with time lately).

src/Magnum/Math/Quaternion.h
```diff
@@ -733,6 +740,57 @@ template<class T> Matrix3x3<T> Quaternion<T>::toMatrix() const {
     };
 }
 
+/* Algorithm from:
+   https://github.com/qt/qtbase/blob/84056cdca11c63ddb9fac3d95fcfa952adaa6ea1/src/gui/math3d/qquaternion.cpp#L507 */
```
mosra: Could this reference the original code Qt is based off, instead? Are there any differences? Because Qt is LGPL and GPL means problems :)

src/Magnum/Math/Quaternion.h
```diff
@@ -733,6 +740,57 @@ template<class T> Matrix3x3<T> Quaternion<T>::toMatrix() const {
...
+template<class T> Vector3<T> Quaternion<T>::toRotation() const {
+    Vector3<T> euler;
```
mosra:
```suggestion
    Vector3<T> euler{NoInit};
```

src/Magnum/Math/Quaternion.h
```diff
@@ -733,6 +740,57 @@ template<class T> Matrix3x3<T> Quaternion<T>::toMatrix() const {
...
+        if(euler.x() > -Constants<T>::pi()) {
+            euler.y() = std::atan2(2.0f*(xz + yw), 1.0f - 2.0f*(xx + yy));
+            euler.z() = std::atan2(2.0f*(xy + zw), 1.0f - 2.0f*(xx + zz));
+        } else {
+            euler.z() = 0.0f;
+            euler.y() = -std::atan2(-2.0f*(xy - zw), 1.0f - 2.0f*(yy + zz));
+        }
+    } else {
+        euler.z() = 0.0f;
+        euler.y() = std::atan2(-2.0f*(xy - zw), 1.0f - 2.0f*(yy + zz));
```
mosra: All the literals need to be `T(0.0)` and `T(2.0)` respectively to avoid compiler warnings when using the double-precision `Quaterniond`.

src/Magnum/Math/Quaternion.h
```diff
@@ -733,6 +740,57 @@ template<class T> Matrix3x3<T> Quaternion<T>::toMatrix() const {
...
+    T xw = xp*wp;
+    T yy = yp*yp;
+    T yz = yp*zp;
+    T yw = yp*wp;
+    T zz = zp*zp;
+    T zw = zp*wp;
+
+    const T lengthSquared = xx + yy + zz + wp*wp;
+    if(!TypeTraits<T>::equalsZero(lengthSquared - 1.0f, 100) &&
+        !TypeTraits<T>::equalsZero(lengthSquared, 100)) {
```
mosra: Not sure about the `100` here -- doesn't it make the comparison *too* fuzzy? If the quat is normalized (which is a prerequisite for extracting the angles?), the original magnitude is `T(1.0)`, so that's what should be there I think.

src/Magnum/Math/Quaternion.h
```diff
@@ -733,6 +740,57 @@ template<class T> Matrix3x3<T> Quaternion<T>::toMatrix() const {
     };
 }
 
+/* Algorithm from:
+   https://github.com/qt/qtbase/blob/84056cdca11c63ddb9fac3d95fcfa952adaa6ea1/src/gui/math3d/qquaternion.cpp#L507 */
+template<class T> Vector3<T> Quaternion<T>::toRotation() const {
```
mosra: Shouldn't this assert for the quaternion being normalized, similarly to how `angle()` / `axis()` have it? AFAIK Qt's quaternions are always normalized, so there they don't check this.

The docs should then mention 

> Expects the quaternion to be normalized.

Melix19:
Finally had time for this!

So, I changed the algorithm to the one found in [wikipedia](https://en.wikipedia.org/wiki/Conversion_between_quaternions_and_Euler_angles#Quaternion_to_Euler_Angles_Conversion), the reason is that the Qt's one was giving a "dancing" result when converting from/to eulers in every frame (I know that this is never a good idea anyway, but it was just for testing):
Dancing rect

With this new algorithm is basically perfect.
Then, for the tests I just added a random quaternion and compared the euler result to a result given from [this online converter](https://www.andre-gaschler.com/rotationconverter/) given the same quaternion.

If there's nothing else to do, I think that it can be merged :D

alanjfs:
Woho! Good timing! I also need this. :D

src/Magnum/Math/Quaternion.h
```diff
@@ -733,6 +746,39 @@ template<class T> Matrix3x3<T> Quaternion<T>::toMatrix() const {
...
+
+    /* Roll (x-axis rotation) */
+    T sinrCosp = T(2.0) * (qw * qx + qy * qz);
+    T cosrCosp = T(1.0) - T(2.0) * (qx * qx + qy * qy);
+    euler.x() = Rad<T>(std::atan2(sinrCosp, cosrCosp));
+
+    /* Pitch (y-axis rotation) */
+    T sinp = T(2.0) * (qw * qy - qz * qx);
+    if(std::abs(sinp) >= T(1.0))
+        euler.y() = Rad<T>(std::copysign(Constants<T>::pi() / T(2.0), sinp)); /* Use 90 degrees if out of range */
```
mosra:
```suggestion
        euler.y() = Rad<T>(std::copysign(Constants<T>::piHalf(), sinp)); /* Use 90 degrees if out of range */
```

Also, codecov is reporting this line is not covered. Can you craft a value for the test that hits this condition? :) Then it'll be perfect.

Melix19:
Sorry for the delay.
I added another test with other values trying to cover the remaining line, but for some reason it gives me the wrong result. For reference:

```
    Quaternion a2 = Quaternion({0.70f, 0.0f, 0.70f}, 0.0f).normalized();
    Math::Vector3<Rad> b2{0.0_radf, -1.57079_radf, -3.14159_radf};

    CORRADE_COMPARE(a2.toEuler(), b2);
```
```
    Values a2.toEuler() and b2 are not the same, actual is
    Vector(Rad(0), Rad(-1.57045), Rad(0))
    but expected
    Vector(Rad(0), Rad(-1.57079), Rad(-3.14159))
```

So, I changed the algorithm again, using directly the one that [this](https://www.andre-gaschler.com/rotationconverter/) online converter [uses](https://github.com/mrdoob/three.js/blob/6892dd0aba1411d35c5e2b44dc6ff280b24d6aa2/src/math/Euler.js#L197) that in my tests it seems the perfect one. The licence is MIT, so it should be good.

Hopefully the coverage is 100% now.

src/Magnum/Math/Quaternion.h
```diff
@@ -733,6 +746,37 @@ template<class T> Matrix3x3<T> Quaternion<T>::toMatrix() const {
...
+
+    T m11 = rotMatrix[0][0];
+    T m12 = rotMatrix[0][1];
+    T m13 = rotMatrix[0][2];
+    T m21 = rotMatrix[1][0];
+    T m22 = rotMatrix[1][1];
+    T m23 = rotMatrix[1][2];
+    T m33 = rotMatrix[2][2];
+
+    euler.y() = Rad<T>(std::asin(-Math::min(Math::max(m13, T(-1.0)), T(1.0))));
```
Melix19: Oh, and here I tried to use the `Math::clamp()` instead of the `Math::min(Math::max())`, but for some reason it says that it does not exist (?)

mosra: It's defined in `Functions.h` that are not included here. This is fine :)

mosra:
Merged as 097289850aeebb9283e20de7b060d9585457c2e5, thank you! Interesting that there's so many algos out there but only some of them work as expected :D

PR #403, opened 2019-11-24
GL: Add ArrayView<AbstractTexture*> overrides to bind and bindImages

Squareys:
Hi @mosra !

As promised, here the PR for the ArrayView overrides in AbstractTexture.
I checked building with doxygen, that is still running, but seems to still be able to resolve references correctly. I removed a couple of todos that referenced doxygen in the process, though, so there might be errors further down the line.
I'll ping once that has run through.

Best, Jonathan

### DIFF
--- a/src/Magnum/GL/AbstractTexture.cpp
+++ b/src/Magnum/GL/AbstractTexture.cpp
@@ -156,7 +156,7 @@ void AbstractTexture::unbind(const Int firstTextureUnit, const std::size_t count
 }
 
 /** @todoc const std::initializer_list makes Doxygen grumpy */
-void AbstractTexture::bind(const Int firstTextureUnit, std::initializer_list<AbstractTexture*> textures) {
+void AbstractTexture::bind(const Int firstTextureUnit, Containers::ArrayView<AbstractTexture* const> textures) {
     /* State tracker is updated in the implementations */
     Context::current().state().texture->bindMultiImplementation(firstTextureUnit, {textures.begin(), textures.size()});
 }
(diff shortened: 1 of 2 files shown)


src/Magnum/GL/AbstractTexture.cpp
```diff
@@ -288,8 +287,7 @@ void AbstractTexture::unbindImage(const Int imageUnit) {
 }
 
 #ifndef MAGNUM_TARGET_GLES
-/** @todoc const Containers::ArrayView makes Doxygen grumpy */
```
mosra: I think this comment is still valid, same above :P

src/Magnum/GL/AbstractTexture.h
```diff
@@ -306,8 +311,11 @@ class MAGNUM_GL_EXPORT AbstractTexture: public AbstractObject {
          * @requires_gl44 Extension @gl_extension{ARB,multi_bind}
          * @requires_gl Multi bind is not available in OpenGL ES and WebGL.
          */
+        static void bindImages(Int firstImageUnit, Containers::ArrayView<AbstractTexture* const> textures);
+
+        /** @overload */
         static void bindImages(Int firstImageUnit, std::initializer_list<AbstractTexture*> textures) {
-            bindImagesInternal(firstImageUnit, {textures.begin(), textures.size()});
+            AbstractTexture::bindImages(firstImageUnit, Containers::arrayView(textures.begin(), textures.size()));
```
mosra: I wonder why `AbstractTexture::` is needed here but not above?

Squareys:
@mosra Sorry for the very long delay. Fixed and rebased.

mosra:
Thanks a lot (and extra :+1: for the rebase, makes things easier)! No problem about the delay :)

PR #406, opened 2019-12-13
Stop Swizzle.h(75): warning C4389: '==': signed/unsigned mismatch

roig:
Tiny fix, to suppress that annoying warning.


mosra:
Thank you!

PR #418, opened 2020-01-15
Add clarification to Cmake documentation

JacksonCampolattaro:
This PR should improve the clarity of the documentation about the use of Magnum as a Cmake subproject.

I tweaked a couple of paragraphs to make it more obvious that the Modules directory is still necessary even when not using a shared installation of Magnum.

### DIFF
--- a/doc/cmake.dox
+++ b/doc/cmake.dox
@@ -65,16 +65,16 @@ separate them with semicolons.
 
 @section cmake-subproject Using Magnum as a CMake subproject
 
-A self-contained alternative to the above is adding Corrade and Magnum
-repositories into your project (as a Git submodule, bundling a downloaded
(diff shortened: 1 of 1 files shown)


JacksonCampolattaro:
@mosra Thanks for the help resolving my issue earlier! This should stop others from making the same mistake I did.

mosra:
Thank you! I'll be mirrorring this change in docs for corrade, plugins, ... so merging might take a while. In the meantime, if you find something else, feel free to add more to this PR :)

mosra:
Merged as d78ceb09abdfe3ed3e591c28942bdf86f40cb719, with a corresponding Corrade update in mosra/corrade@6a3875969f970ee2f3bceadc07064c61ef820d00. Thanks a lot!

PR #419, opened 2020-01-30
Platform: implement mouse capturing for SDL and GLFW

mosra:
Isn't as straightforward as it originally seemed, and I don't want to push something that'd be deprecated right after in favor of a more consistent API.

TODOs: 

- [ ] GLFW has this in a branch right now, scheduled for 3.4 -- https://github.com/glfw/glfw/issues/58
- [ ] GLFW has it together with normal/hidden/locked (which would mean extending `setCursor()`) and those options are mutually exclusive, so maybe do it that way in both instead of `setMouseCaptured()` that's done in case of SDL?
- [ ] while SDL reports move event coordinates outside of the window, GLFW docs in the branch seem to indicate that the mouse position won't change -- would this mean the mouse move events won't get reported? Would need to test that branch to see.

### DIFF
--- a/src/Magnum/Platform/GlfwApplication.h
+++ b/src/Magnum/Platform/GlfwApplication.h
@@ -625,6 +625,10 @@ class GlfwApplication {
             glfwSetCursorPos(_window, Double(position.x()), Double(position.y()));
         }
 
+        /** @todo GLFW_CURSOR_CAPTURED, currently only in a branch:
+            https://github.com/glfw/glfw/issues/58
+            https://github.com/glfw/glfw/compare/captured-cursor-mode-r2 */
+
     private:
         /** @copydoc Sdl2Application::mousePressEvent() */
         virtual void mousePressEvent(MouseEvent& event);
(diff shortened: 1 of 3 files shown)


LB--:
> GLFW docs in the branch seem to indicate that the mouse position won't change -- would this mean the mouse move events won't get reported? Would need to test that branch to see.

My understanding is that GLFW already has "disabled" mode where the mouse is locked to the center of the screen and hidden, intended for first person games where you move the mouse to look around and the speed of camera movement should not be limited at all by the window/screen size, whereas the PR you linked to leaves the mouse cursor visible and free to move anywhere, but prevents it from escaping the client area. So, maybe you were looking at the wrong docs or they were copypasted and not edited properly from the "disabled" docs. At a quick glance, I don't see any commits related to that PR in that branch that affect documentation at all.

mosra:
I know about the disabled/hidden mode and there it's a bit different than this ... from the discussion / branch docs I got the impression that the mouse coordinates would get clamped to the window area, i.e. not changing in any way when outside of the window: https://github.com/glfw/glfw/commit/99b55dd21bc64a1b1bc470b53f07df4cb6282ca1

I guess I'll just wait until this materializes.

PR #420, opened 2020-02-06
Add reflect and refract functions

ttnghia:
Here are two very useful functions: reflect and refract, added into Functions.h.
Enjoy.


mosra:
Merged (with various additions) as 1ef813fdade9f67a60970d166096c34b25e6dd87, thank you!

PR #425, opened 2020-02-26
Math: Add non-const overloads for Frustum::(left|right|top|bottom|nea…

Squareys:
Hi @mosra !

I need this to split the frustum into multiple sub-frustums. Here's a diff to open the discussion.

Best,
Jonathan

### DIFF
--- a/src/Magnum/Math/Frustum.h
+++ b/src/Magnum/Math/Frustum.h
@@ -146,10 +146,17 @@ template<class T> class Frustum {
 
         /**
          * @brief Plane at given index
+         * @m_since_latest
          *
          * Expects that @p i is less than @cpp 6 @ce.
          */
+        Vector4<T>& operator[](std::size_t i) {
+            CORRADE_ASSERT(i < 6, "Math::Frustum::operator[](): index" << i << "out of range",
+                _data[i]);
+            return _data[i];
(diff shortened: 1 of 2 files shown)


src/Magnum/Math/Frustum.h
```diff
@@ -198,21 +198,51 @@ template<class T> class Frustum {
...
+         */
+        Vector4<T> near() { return _data[4]; }
 
         /** @brief Far plane */
         constexpr Vector4<T> far() const { return _data[5]; }
+        /**
+         * @overload
+         * @m_since_latest
+         */
+        Vector4<T> far() { return _data[5]; }
```
mosra: It's usually that the order is non-const first, const marked as overload, so for consistency

```cpp
        /** 
         * @brief Far plane 
         * @m_since_latest
         */
        Vector4<T> far() { return _data[5]; }
        constexpr Vector4<T> far() const { return _data[5]; } /**< @overload */
```

Same for others. (Yes, it'll be slightly weird that the main function is documented as new and the overload not, but I don't think that's a serious problem.)

mosra:
Merged as a0e6ac1ed05411f45055aefbf5cc380b4c1b49d6, thanks!

PR #427, opened 2020-03-11
Mesh data serialization, meshlet support and scene conversion plugin APIs

mosra:
Continuation of #371. Things to do (list will grow):

- [ ] Create a new ~~`AbstractMeshConverter` plugin interface for operating on
    the `MeshData` ... or~~ `AbstractSceneConverter` for merging / splitting meshes etc <kbd>in progress</kbd>
    - most algorithms need to operate on the whole scene tree, mesh conversion alone has little use
- [ ] Create a new ~~`magnum-meshconverter`~~ `magnum-sceneconverter` utility that would operate in-place on imported data (needs #240 as well, tho) <kbd>in progress</kbd>
- [ ] Integrate meshoptimizer
- [ ] Meshlet / mesh shader support
    - [ ] ~~Two new views on the indexData, one a "running offset" saying how many elements is per a meshlet, second picking vertices from the global vertex array for given meshlet; the original index array is then for each meshlet telling which *local* vertices to pick (so it's probably just 8-bit always); how to name them? `meshletOffsets()` & `meshletIndices()`?~~ instead a completely new MeshData instance containing per-instance and two new `MeshPrimitive::TrianglesMeshlet` and `MeshPrimitive::Meshlet` primitives distinguishing the two from each other
    - [ ] Some `MeshTools` algo to combine the global and local index arrays back to a classic index array
- [ ] Pin `VertexFormat` enum values to ensure backwards compatibility of serialized data with new values added
  - [ ] Same for `MeshPrimitive`, `Trade::MeshAttribute`, `MeshIndexType`
  - [ ] And `PixelFormat`
- [ ] Serialization basics (see below) <kbd>in progress</kbd>

### Serialization

Basically making the `MeshData` (and later other `*Data` as well) easily serializable, allowing projects to convert arbitrary mesh formats to blobs that could be loaded directly via mmaping:

```sh
magnum-sceneconverter file.blend --mesh "chair" --serialize chair.blob
magnum-sceneconverter scene.glb --mesh "tree" --serialize tree.blob
...

cat chair.blob tree.blob car.blob > blobs.blob # because why not
```

```cpp
// Takes basically no time
auto blob = Utility::Directory::mapRead("blobs.blob");

// (yes really) and this also
const Trade::MeshData& chair = *reinterpret_cast<const Trade::MeshData*>(blob.data());

// tho this is safer as it does a bunch of extra checks
const Trade::MeshData* chair = Trade::MeshData::from(blob);
```

Key points:

-   Each `*Data` would contain the same header with some magic, version info and total size. Magic and version info for sanity checks, total size to be able to simply `cat` multiple blobs together without having to worry about metadata. Similar to [RIFF](https://en.wikipedia.org/wiki/Resource_Interchange_File_Format) chunks (~~can't be 100% compatible because it's limited to 4 GB max~~ RIFF64?, can look there for inspiration).

    ```cpp
    struct DataHeader {
        // some magic and version info (ideally stolen from PNG to detect broken encodings, CRLF conversion, bad endiannes etc.)
        DataType type;
        uint64_t size;
        // ... more stuffs

        const DataHeader* nextData() const {
            return reinterpret_cast<const char*>(this) + size;
        }
    };

    struct Trade::MeshData: DataHeader { ... };

    ...
    Containers::ArrayView<const char> data;
    for(auto i = static_cast<const DataHeader*>(data); i < data.end(); i = i->nextData()) {
        if(i->type == DataType::MeshData) {
            const auto& mesh = static_cast<const Trade::MeshData&>(*i);
            ...
        } else if(i-> type == DataType::AnimationData) {
            ...
        } else continue;
    }
    ```

    The concatenated blob can then be iterated like above, ignoring unknown chunks.
-   The `MeshData` need to support both self-contained operation (data bundled internally in an `Array`) and the above-shown "reinterpreted" operation. ~~Because in the reinterpreted case no destructors are called, this could be done by putting a tag into the `Array` deleter field --- ideally with an odd value to minimize the chance an actual function could have the same address (I assume functions are never on odd addresses). Both `M` and `!` satisfy this condition so the following would be an odd address on both BE and LE systems~~ Instead make the Arrays `nullptr` with size, which will make the default deleter no-op (in case someone actually manages to call the destructor) and all offset calculation working correctly as well -- when returning pointers, the actual memory location would be added to the offset.
-   ~~In order to avoid extra work, the data layout should be consistent between 32bit and 64bit systems -- otherwise it won't be possible to serialize data on a (64bit) desktop and use them on a (32bit) Emscripten, e.g.. `Array` is 12 bytes on 32bit and 24 bytes on 64bit (pointer, size and deleter) so I think it could be enough to add padding around it on 32bit systems; plus the above "deleter tag" logic needs to be extended to check for both a 32bit and a 64bit case.~~ Too complicated, the serialization code needs to be able to create 32bit, 64bit, LE and BE variants, "reinterpret" deserialization will work only with data made for given platform. (But there needs to be some way to decode data from different platforms also...)
-    Some assets might have one huge buffer and particular meshes being just views on it. Ideally the buffer should be uploaded as-is in a single piece, with meshes then referring subranges. In this case, the serialized MeshData need to have a way to reference an external "data chunk" somehow -- some flag saying the offset not internal but to a different chunk + a data chunk "ID"? Or having that implicit -- it'll always be the first data chunk after a couple of data-less MeshData chunks?

Discussion points:

- make it possible to record a command-line used to produce a particular chunk? (e.g. [as proposed in gcc](https://www.phoronix.com/scan.php?page=news_item&px=GCC-Record-Command-Line-Info)) ... or other comments? might be useful for repeatability


mosra:
Initial bits of `magnum-sceneconverter` pushed to master in 1c51b98d76ed4ee74c7d73a5aa929139d6893ccf.

janbajana:
Testing this out. 
I have difficulties to get to see any data from `FrameProfiler`.

I have checked out the same branch in `magnum-plugins` and `magnum-extracts`. So I guess `magnum-player` is using `FrameProfiler`? As you showed on your picture here:
https://files.gitter.im/mosra/magnum/jQ2e/image.png

For `magnum-plugins` do I have to build those plugins to have working `FrameProfiler`?
```
        -DWITH_MAGNUMIMPORTER=ON \
        -DWITH_MAGNUMSCENECONVERTER=ON \
        -DWITH_MESHOPTIMIZERSCENECONVERTER=ON \
```
I gues extenral dependency is `zeux/meshoptimizer` only?

But I still do not understand how `magnum-player` is using `FrameProfiler` in your branch to output data?

mosra:
@janbajana err, sorry for the confusion. 

The [FrameProfiler](https://doc.magnum.graphics/magnum/classMagnum_1_1DebugTools_1_1FrameProfiler.html) is in master now, no need to use this branch for it (it was briefly here, but because it got stable pretty fast, I put it in master). For the `magnum-player` it wasn't anywhere yet, I pushed a WIP to magnum-extras `next` just now: mosra/magnum-extras@551cbc05c4c810d609806a5535f8b896215133c2.

mosra:
Eh, forgot to say -- you need to press the <kbd>P</kbd> key to toggle it.

janbajana:
Super that worked. I have statistics now. 
```
Opening file ../../scenes/Buggy.gltf
Loading 0 textures
Loading 148 materials
Loading 148 meshes
Loading 251 objects
Loading scene 0 
Last 50 frames:
  Frame time: 16.68 ms
  CPU duration: 16.30 ms
  GPU duration: 11.68 ms
```

Thanks for the help.

mosra:
Partially merged to master:

- initial scene converter plugin interfaces -- 2dc47839ef0b347a660c834800b1e0e610d66e2e
- `AnySceneConverter` plugin -- 0ad5a8968fce888bfe33cd5f43b42a5c3b551649
- support in `magnum-sceneconverter` -- 0da8f89055a91b01b2fb41ec9a988510a8940e13

What's left in this PR is everything related to mesh serialization, which needs a few more iterations before it's ready.

PR #432, opened 2020-04-06
Math/Random and its tests implementation

sariug:
Hello hello,

Did some simple implementation; would be happy to develop it further with the suggestions.
Prepared for one engine ("mt19937")

### DIFF
--- a/src/Magnum/Math/Random.h
+++ b/src/Magnum/Math/Random.h
@@ -58,15 +58,10 @@ public:
 namespace Random
 {
 template <class T = Float>
-T randomUnsignedScalar() // range [0, 1]
+T randomScalar(T begin = 0.0f, T end = 1.0f)
 {
-    return Implementation::RandomGenerator::generate<T>(static_cast<T>(0.0f),
-                                                        static_cast<T>(1.0f));
-}
-template <class T = Float>
-T randomSignedScalar() // range [-1, 1]
-{
-    return Implementation::RandomGenerator::generate(static_cast<T>(-1.0f), static_cast<T>(1.0f));
+    return Implementation::RandomGenerator::generate(static_cast<T>(begin),
+                                                     static_cast<T>(end));
 }
 
 template <class T = Float>
@@ -82,7 +77,7 @@ Vector3<T> randomUnitVector3()
     // Better to have it "theta" and "z" than three random numbers.
     // https://mathworld.wolfram.com/SpherePointPicking.html
     auto a = Implementation::RandomGenerator::generate(0.0f, 2 * Math::Constants<T>::pi());
-    auto z = randomSignedScalar();
+    auto z = randomScalar(-1.0f, -1.0f);
     auto r = sqrt<T>(1 - z * z);
     return {r * std::cos(a), r * std::sin(a), z};
 }
@@ -91,9 +86,9 @@ template <class T = Float>
 Quaternion<T> randomRotation()
 {
     //http://planning.cs.uiuc.edu/node198.html
-    auto u{randomUnsignedScalar()};
-    auto v{2 * Math::Constants<T>::pi() * randomUnsignedScalar()};
-    auto w{2 * Math::Constants<T>::pi() * randomUnsignedScalar()};
+    auto u{randomScalar()};
+    auto v{2 * Math::Constants<T>::pi() * randomScalar()};
+    auto w{2 * Math::Constants<T>::pi() * randomScalar()};
     return Quaternion<T>({sqrt<T>(1 - u) * std::sin(v),
                           sqrt<T>(1 - u) * std::cos(v),
                           sqrt<T>(u) * std::sin(w)},
--- a/src/Magnum/Math/Random.h
+++ b/src/Magnum/Math/Random.h
@@ -80,48 +80,25 @@ Vector2<T> randomUnitVector2()
 template <class T = Float>
 Vector3<T> randomUnitVector3()
 {
+    // Better to have it "theta" and "z" than three random numbers.
+    // https://mathworld.wolfram.com/SpherePointPicking.html
     auto a = Implementation::RandomGenerator::generate(0.0f, 2 * Math::Constants<T>::pi());
     auto z = randomSignedScalar();
     auto r = sqrt<T>(1 - z * z);
     return {r * std::cos(a), r * std::sin(a), z};
 }
-template <class T = Float>
-Vector2<T> randomPointInACircle() // always length < 1
-{
-    while (true)
-    {
-        auto p = Vector2<T>(
-            randomSignedScalar(), randomSignedScalar());
-        if (p.length() >= 1)
-            continue;
-        return p;
-    }
-}
-
-template <class T = Float>
-Vector3<T> randomPointInASphere() // always length < 1
-{
-
-    while (true)
-    {
-        auto p = Vector3<T>(randomSignedScalar(),
-                            randomSignedScalar(),
-                            randomSignedScalar());
-        if (p.length() >= 1)
-            continue;
-        return p;
-    }
-}
-
-bool randomBool()
-{
-    return static_cast<bool>(randomUnsignedScalar<Int>());
-}
 
 template <class T = Float>
 Quaternion<T> randomRotation()
 {
-    return Quaternion<T>({randomSignedScalar<T>(), randomSignedScalar<T>(), randomSignedScalar<T>()}, randomSignedScalar<T>()).normalized();
+    //http://planning.cs.uiuc.edu/node198.html
+    auto u{randomUnsignedScalar()};
+    auto v{2 * Math::Constants<T>::pi() * randomUnsignedScalar()};
+    auto w{2 * Math::Constants<T>::pi() * randomUnsignedScalar()};
+    return Quaternion<T>({sqrt<T>(1 - u) * std::sin(v),
+                          sqrt<T>(1 - u) * std::cos(v),
+                          sqrt<T>(u) * std::sin(w)},
+                         sqrt<T>(u) * std::cos(w));
 }
 
 } // namespace Random
--- a/src/Magnum/Math/Random.h
+++ b/src/Magnum/Math/Random.h
@@ -86,9 +86,9 @@ template <class T = Float>
 Quaternion<T> randomRotation()
 {
     //http://planning.cs.uiuc.edu/node198.html
-    auto u{randomScalar()};
-    auto v{2 * Math::Constants<T>::pi() * randomScalar()};
-    auto w{2 * Math::Constants<T>::pi() * randomScalar()};
+    auto u = randomScalar();
+    auto v = 2 * Math::Constants<T>::pi() * randomScalar();
+    auto w = 2 * Math::Constants<T>::pi() * randomScalar();
     return Quaternion<T>({sqrt<T>(1 - u) * std::sin(v),
                           sqrt<T>(1 - u) * std::cos(v),
                           sqrt<T>(u) * std::sin(w)},
--- a/src/Magnum/Math/Random.h
+++ b/src/Magnum/Math/Random.h
@@ -17,78 +17,75 @@ namespace Magnum
 namespace Math
 {
 
-namespace Implementation
+namespace Random
 {
-static std::seed_seq seeds{{
-    static_cast<std::uintmax_t>(std::random_device{}()),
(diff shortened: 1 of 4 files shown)


src/Magnum/Math/Random.h
```diff
@@ -0,0 +1,131 @@
...
+    generate(T start = -Magnum::Math::Constants<T>::inf(),
+             T end = Magnum::Math::Constants<T>::inf())
+    {
+        return std::uniform_real_distribution<T>{start, end}(generator());
+    }
+
+public:
+    static std::mt19937 &generator()
+    {
+        static std::mt19937 g{seeds};
```
mosra, line 53: `static` is problematic because of thread safety issues, global constructors/destructors etc. What do you think about this?

```cpp
Math::RandomGenerator g; // the user is required to instantiate this first

auto a = Math::randomSignedScalar(g);
auto b = Math::randomUnitVector2(g);
```

The `RandomGenerator` thus wouldn't be in `Implementation` anymore, but will be exposed to the user, even though its only use would be to get passed to the `random*()` functions.

src/Magnum/Math/Random.h
```diff
@@ -0,0 +1,131 @@
...
+public:
+    RandomGenerator() = delete;
+
+    template <typename T>
+    static typename std::enable_if<std::is_same<Int, T>::value, T>::type
+    generate(T start = -Magnum::Math::Constants<T>::inf(),
+             T end = Magnum::Math::Constants<T>::inf())
+    {
+        return std::uniform_int_distribution<T>{start, end}(generator());
+    }
```
mosra, line 40: I don't think the integer version is used anywhere. Maybe instead add a `static_assert(IsFloatingPoint<T>::value, "");` to the floating-point version below? (The trait is in [Magnum/Math/TypeTraits.h](https://doc.magnum.graphics/magnum/structMagnum_1_1Math_1_1IsFloatingPoint.html).)

src/Magnum/Math/Random.h
```diff
@@ -0,0 +1,131 @@
...
+T randomSignedScalar() // range [-1, 1]
+{
+    return Implementation::RandomGenerator::generate(static_cast<T>(-1.0f), static_cast<T>(1.0f));
+}
+
+template <class T = Float>
+Vector2<T> randomUnitVector2()
+{
+    auto a = Implementation::RandomGenerator::generate(0.0f, 2 * Math::Constants<T>::pi());
+    return {std::cos(a), std::sin(a)};
```
mosra, line 76: I guess this is where things become *hard* :sweat_smile: 

I actually have no idea here -- will the distribution be still uniform if sin/cos is used? I guess it will? Ideally this would be without the extra overhead of trig functions, but I don't have any idea if that's doable ... [in this thread on SO](https://stackoverflow.com/questions/6283080/random-unit-vector-in-multi-dimensional-space) they use a Gaussian distribution as an input, but .. ¯\\\_(ツ)_/¯

If you have better references than me, mention them here please :)

src/Magnum/Math/Random.h
```diff
@@ -0,0 +1,131 @@
...
+    while (true)
+    {
+        auto p = Vector3<T>(randomSignedScalar(),
+                            randomSignedScalar(),
+                            randomSignedScalar());
+        if (p.length() >= 1)
+            continue;
+        return p;
+    }
+}
```
mosra, line 113: This has a non-deterministic runtime cost, which isn't great for realtime use -- I realized this could be doable as `randomUnitVector2()*randomUnsignedScalar()`, but then the user could do the same and so this function isn't really needed. So maybe remove both `randomPointInACircle()` and `randomPointInASphere()` and instead document that it can be achieved by combining `randomUnit*()` and `randomUnsignedScalar()`?

src/Magnum/Math/Random.h
```diff
@@ -0,0 +1,131 @@
...
+
+bool randomBool()
+{
+    return static_cast<bool>(randomUnsignedScalar<Int>());
+}
```
mosra, line 118: Hmm, do we actually need this? :) The `randomUnsignedScalar()` for integers isn't really useful either. Maybe instead add a `static_assert(IsFloatingPoint<T>::value, "");` to the RandomGenerator and remove everything related to integers? Those can be always added back later. (The trait is in [Magnum/Math/TypeTraits.h](https://doc.magnum.graphics/magnum/structMagnum_1_1Math_1_1IsFloatingPoint.html).)

src/Magnum/Math/Test/RandomTest.cpp
```diff
@@ -0,0 +1,112 @@
...
+    {
+        CORRADE_VERIFY((Math::Random::randomPointInASphere()).length()<1.0f);
+    }
+}
+void RandomTest::randomRotation(){
+    for (std::size_t i = 0; i < 10; i++)
+    {
+        CORRADE_COMPARE(Math::Random::randomRotation().length(), 1.0f);
+    }
+}
```
mosra, line 106: It would be great to have some sort of distribution verification in the tests, to ensure we're not accidentally skewing the distribution to something non-uniform -- how good are your statistics skills? :)

Found [this on SO](https://math.stackexchange.com/questions/2435/is-there-a-simple-test-for-uniform-distributions), it suggests using a Chi Squared test. I never did such a thing myself tho :D

mosra (commented):
I really appreciate that you did a test as well :+1: 

Since this is your first contribution, I'm not commenting on coding style -- let's get this working first :)

src/Magnum/Math/Test/RandomTest.cpp
```diff
@@ -0,0 +1,112 @@
...
+//                              Vector4(4.5f,  4.0f, 7.0f,  2.0f),
+//                              Vector4(1.0f,  2.0f, 3.0f, -1.0f),
+//                              Vector4(7.9f, -1.0f, 8.0f, -1.5f)};
+//     CORRADE_COMPARE(a, Matrix4x4(Vector4(3.0f,  5.0f, 8.0f, -3.0f),
+//                                  Vector4(4.5f,  4.0f, 7.0f,  2.0f),
+//                                  Vector4(1.0f,  2.0f, 3.0f, -1.0f),
+//                                  Vector4(7.9f, -1.0f, 8.0f, -1.5f)));
+
+//     CORRADE_VERIFY((std::is_nothrow_constructible<Matrix4x4, Vector4, Vector4, Vector4, Vector4>::value));
+// }
```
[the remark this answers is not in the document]
sariug: hahaha wow. Should have a better check for later !

src/Magnum/Math/Random.h
sariug: I thought about it, and tried to keep minimalistic(however you are right for the thread safety), to keep it as a 1-step function. 

if you have no other proposal and think that this would be the best, then why not :)

src/Magnum/Math/Random.h
sariug: For this case sin and cos shall not be get affected ? (sin (+ + - - )[50%],  cos (+ - - + )[50%] )
So this shall be as accurate as the generator itself ? 

Can you also comment about what I read [here](https://mathworld.wolfram.com/SpherePointPicking.html) ? It looks accurate according to this.

src/Magnum/Math/Random.h
sariug: This was my selfish implementation(not in your suggestions haha). I dont event think we need to document this; Get a unit sphere; multiply with a [0,1] length for example.

src/Magnum/Math/Test/RandomTest.cpp
sariug: We have easter incoming ! :)

src/Magnum/Math/Random.h
sariug: Well, like i said in the beginning, a initiation for a new magnum functionality :) Can grow later with demand :) 

Deleting/Fixing. :)

src/Magnum/Math/Random.h
mosra: *Yes* that's a great reference, thanks -- a link to it should go in the documentation. I didn't absorb it fully yet, but yeah it sounds like a good proof. 

> getting 3 random numbers and normalizing

this is what you do for quaternions tho .. can the two-/three-dimensional case be extended for those as well?

src/Magnum/Math/Random.h
mosra: Not sure if best, but I didn't come up with anything better yet.

I think this should be also robust in case I decide to implement my own random generator later.

src/Magnum/Math/Test/RandomTest.cpp
```diff
@@ -0,0 +1,112 @@
...
+void RandomTest::unsignedScalar()
+{
+    for (std::size_t i = 0; i < 10; i++)
```
mosra, line 58: Instead of a loop in every test you can replace `addTests()` with [addRepeatedTests()](https://doc.magnum.graphics/corrade/classCorrade_1_1TestSuite_1_1Tester.html#a06c40ddf62db02c1575d019b630d7f08) and set the repeat count there. I think repeating every test 100 times won't hurt, 10 is too little :)

src/Magnum/Math/Test/RandomTest.cpp
```diff
@@ -0,0 +1,112 @@
...
+    {
+        CORRADE_VERIFY(Math::Random::randomUnsignedScalar<Float>() < 1.000000001);
+        CORRADE_VERIFY(Math::Random::randomUnsignedScalar<Float>() > -.000000001);
```
mosra, line 61: For greater / less-than comparison there's [TestSuite::Compare::Less](https://doc.magnum.graphics/corrade/classCorrade_1_1TestSuite_1_1Compare_1_1Less.html) and friends. It does the same as the above but can actually print the two numbers if something fails, which is better than just saying "this failed".

sariug: Done.

src/Magnum/Math/Test/RandomTest.cpp
sariug: Top !

src/Magnum/Math/Random.h
sariug: I first deleted, and then put it back haha :D 
From our initial discussion:
```
Float randomUnsignedScalar(); // range [0, 1]
Float randomSignedScalar(); // range [-1, 1]
Vector2 randomPointOnACircle(); // always length = 1
Vector3 randomPointOnASphere(); // also
Quaternion randomRotation(); // always normalized
```

Maybe keeping it with integers to not to cast anything if we roll a dice?(`randomUnsignedScalar<int>(1,6)`)or to see if that girl is single?(`randomUnsignedScalar<int>(0,1)`[bool to cast, yes. maybe type traits for numerics?])

src/Magnum/Math/Random.h
sariug: Quaternion is link [here.](    //http://planning.cs.uiuc.edu/node198.html)

Added both to the code :)

src/Magnum/Math/Test/RandomTest.cpp
sariug: Well, I put a chi square test. In testing, I couldnt see a "tolerant" testing scheme(Chi square let 1 fail; like 99% ).

src/Magnum/Math/Random.h
sariug: Implemented some stuff. Wondering how you would find these. I feel quiteee doubtful.(Especially would love to hide the "generate" function;

mosra:
Okay, great job! Everything seems to compile and pass tests except a few cases where 

> Value error_count is not less than 2, actual is 2 but expected < 2

I'm not sure how bad this is tho, seems to be depending on how good the standard implementation is? :)

I'll need some time to write the docs etc., but I think this can go in :tada:

sariug:
Great news ! 

> > Value error_count is not less than 2, actual is 2 but expected < 2
> 
> I'm not sure how bad this is tho, seems to be depending on how good the standard implementation is? :)

So this happens in chisquare I guess, which the randomness in belong to std engine.. 
Let me know if i can support anyhow !

mosra:
TODOs for me (adding a comment so I don't forget):

- [ ] move `<random>` to a cpp file because it's *heavy* -- https://artificial-mind.net/projects/compile-health/
- [ ] investigate what's up with `std::seed_seq` -- https://twitter.com/horenmar_ctu/status/1251044181558124544

PR #433, opened 2020-04-08
[WIP] Added shared context support

aspioupiou:
Magnum provides boostrap classes to setup OpenGL context and windows (or not using  `WindowlessXXXXX`). These creates an OpenGL context from scratch.

Magnum provides also in-app multi-context support where several contexts can coexists (each 3D contexts beeing related to its magnum counterpart)

However in some application, we may need to have several coexisting OpenGL shared contexts. A potential use case is when we want to share datas between A and B (like a master-slave architecture) but we don't want to modify the state of the other (with the same context, binding a buffer in B will be bound for A and B).

Technical solutions proposed here is to add a `Configuration::setSharedContext(GLXContext ctx)` that will be passed to the platform dependent context creation function (here `glXCreateContextAttribsARB`).

This is a WIP since only WindowlessGLXApplication has been updated

### DIFF
--- a/src/Magnum/Platform/WindowlessGlxApplication.cpp
+++ b/src/Magnum/Platform/WindowlessGlxApplication.cpp
@@ -40,15 +40,6 @@ namespace Magnum { namespace Platform {
 WindowlessGlxContext::WindowlessGlxContext(const WindowlessGlxContext::Configuration& configuration, GLContext* const magnumContext) {
     _display = XOpenDisplay(nullptr);
 
-    /**
-     * The user can choose to create a windowless application with a dedicated OpenGL context
-     * or give an OpenGL context to create a shared Context instead.
-     * 
-     * Here, we ask the configuration whether a context has been specified by the user 
-     * (!= nullptr) or not.
-     */
-    GLXContext ctx = configuration.sharedContext();
-    
     /* Check version */
     int major, minor;
     glXQueryVersion(_display, &major, &minor);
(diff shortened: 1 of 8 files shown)


src/Magnum/Platform/WindowlessGlxApplication.cpp
```diff
@@ -94,7 +103,7 @@ WindowlessGlxContext::WindowlessGlxContext(const WindowlessGlxContext::Configura
         #endif
         0
     };
-    _context = glXCreateContextAttribsARB(_display, configs[0], nullptr, True, contextAttributes);
+    _context = glXCreateContextAttribsARB(_display, configs[0], ctx, True, contextAttributes);
```
mosra, line 106: i think just 

```suggestion
    _context = glXCreateContextAttribsARB(_display, configs[0], configuration.sharedContext(), True, contextAttributes);
```

everywhere is descriptive enough, no need for the explanatory comment.

src/Magnum/Platform/WindowlessGlxApplication.h
```diff
@@ -238,8 +238,34 @@ class WindowlessGlxContext::Configuration {
...
 
     private:
+        /**
+         * If the @ref Configuration opengl context
+         * is shared with another context, then _sharedContext points to 
+         * this context.
+         * 
+         * Otherwise = nullptr;
+         */
+        GLXContext _sharedContext{nullptr};
```
mosra, line 268: No need to document a private variable :)

mosra:
Merged as 4d1203d66e5c8fc145b5455a2bc92f50fad36622, thank you!

PR #437, opened 2020-04-30
EGLDisplay non-terminating when mixing WindowlessEGL and sharedContexts

aspioupiou:
When using `WindowlessEGLApplication` (context A) as main context with another WindowlessEGLApplication (context B) sharing context with A, everything runs fine until destroying B. In my case, recreating and destroying several time B in our testsuite crashes. 
Enabling debug output shows `eglMakeCurrent(): EGL_NOT_INITIALIZED error: In eglMakeCurrent: EGLDisplay (0x560efcf486c0) not intiialized` when trying to make current A's EGL context after destroying B.

My research shows that EGL has a specific way to handle the display variable: 
- When creating a GLXwindowless application, an associated display is created: https://github.com/mosra/magnum/blob/ec3e308c9be2fa3168954e5a6dc5ed87c0c70fbe/src/Magnum/Platform/WindowlessGlxApplication.cpp#L78. On destruction, it is destroyed : https://github.com/mosra/magnum/blob/ec3e308c9be2fa3168954e5a6dc5ed87c0c70fbe/src/Magnum/Platform/WindowlessGlxApplication.cpp#L243 . Eveyrhing is fine in this case
- When creating a EGLWindowless application, an associated display is created: https://github.com/mosra/magnum/blob/ec3e308c9be2fa3168954e5a6dc5ed87c0c70fbe/src/Magnum/Platform/WindowlessEglApplication.cpp#L153. However, from https://www.khronos.org/registry/EGL/extensions/EXT/EGL_EXT_platform_base.txt, Multiple calls made to eglGetPlatformDisplayEXT with the same <platform> and <native_display> will return the same EGLDisplay handle.. This means A and B both shares the same display which is destroyed when destroying one of them: https://github.com/mosra/magnum/blob/ec3e308c9be2fa3168954e5a6dc5ed87c0c70fbe/src/Magnum/Platform/WindowlessEglApplication.cpp#L312. Afterwards, it is not possible to make the context of the other current because the EGLDisplay is terminated. Reinitializing the EGLDisplay with `eglInitialize` does not work also with `EGL_BAD_CONTEXT` error (internal state mismatche certainly) 

From my discussion with @mosra, the proposed solution is : 
 - Any `WindowlessEGLApplication` not sharing contexts with another `eglInitialize()` and will  `eglTerminate()` the display
  - Any `WindowlessEGLApplication` sharing context with another won't `eglTerminate()` the display. `eglInitialize` is not needed to be reworked since from the documentation, `initializing an already initialized EGLDisplay is a no-op`

However, to properly shutdown all contexts and related-datas (i.e `EGLDisplay`), we must first shutdown all the `WindowlessEGLApplications` sharing contexts, and finally the one (or another 3rd party EGL-context) who don't.

### DIFF
--- a/src/Magnum/Platform/WindowlessEglApplication.cpp
+++ b/src/Magnum/Platform/WindowlessEglApplication.cpp
@@ -84,6 +84,10 @@ bool extensionSupported(const char* const extensions, Containers::ArrayView<cons
 WindowlessEglContext::WindowlessEglContext(const Configuration& configuration, GLContext* const magnumContext) {
     #ifndef MAGNUM_TARGET_WEBGL
     
+    /** early check to avoid shared EGL context issues */
+    CORRADE_CONSTEXPR_ASSERT((configuration.sharedContext() == EGL_NO_CONTEXT) == (configuration.sharedDisplay() == EGL_NO_DISPLAY),
+	"Platform::WindowlessEglContext::Configuration::setSharedContext(): either both context have to be valid or both invalid.");
+	
 	/**
 	 * The user provided a shared context and its associated display, so we don't query the display
 	 */
@@ -93,14 +97,6 @@ WindowlessEglContext::WindowlessEglContext(const Configuration& configuration, G
    		_display = configuration.sharedDisplay();
    	    _sharedContext = true; 
    	}
-   	/**
-     * The user provided only the shared context or display, or provided both but one of them is wrong
-     */
-    else if (configuration.sharedContext() == EGL_NO_CONTEXT && configuration.sharedDisplay() != EGL_NO_DISPLAY ||
-             configuration.sharedContext() != EGL_NO_CONTEXT && configuration.sharedDisplay() == EGL_NO_DISPLAY)
-   {
-     	Error{} << "Platform::WindowlessEglContext: Either context or display is not valid. Context is " << configuration.sharedContext() << " Display is " << configuration.sharedDisplay();  	    
-   }   
      
     /* If relevant extensions are supported, try to find some display using
        those APIs, as that works reliably also when running headless. This
@@ -109,7 +105,7 @@ WindowlessEglContext::WindowlessEglContext(const Configuration& configuration, G
        we'd need to have a code path for 1.4 *and* 1.5, plus do complicated
        version parsing from a string. Not feeling like doing that today, no. */
     const char* const extensions = eglQueryString(EGL_NO_DISPLAY, EGL_EXTENSIONS);
-    if(extensions &&
+    if(!_sharedContext && extensions &&
         /* eglQueryDevicesEXT(). NVidia exposes only EGL_EXT_device_base, which
            is an older version of EGL_EXT_device_enumeration before it got
            split to that and EGL_EXT_device_query, so test for both. */
@@ -169,14 +165,15 @@ WindowlessEglContext::WindowlessEglContext(const Configuration& configuration, G
         Containers::Array<EGLDeviceEXT> devices{configuration.device() + 1};
         CORRADE_INTERNAL_ASSERT_OUTPUT(eglQueryDevices(configuration.device() + 1, devices, &count));
 
-   	   if(!_sharedContext && !(_display = reinterpret_cast<EGLDisplay(*)(EGLenum, void*, const EGLint*)>(eglGetProcAddress("eglGetPlatformDisplayEXT"))(EGL_PLATFORM_DEVICE_EXT, devices[configuration.device()], nullptr))) {
+   	   if(!(_display = reinterpret_cast<EGLDisplay(*)(EGLenum, void*, const EGLint*)>(eglGetProcAddress("eglGetPlatformDisplayEXT"))(EGL_PLATFORM_DEVICE_EXT, devices[configuration.device()], nullptr))) {
             Error{} << "Platform::WindowlessEglApplication::tryCreateContext(): cannot get platform display for a device:" << Implementation::eglErrorString(eglGetError());
             return;
         }
-    } else
-    #endif
+    } 
+    else if (!_sharedContext)    
     /* Otherwise initialize the classic way. WebGL doesn't have any of the
        above, so no need to compile that at all. */
+    #endif
     {
         #ifndef MAGNUM_TARGET_WEBGL
         if(configuration.device() != 0) {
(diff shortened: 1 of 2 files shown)


mosra:
Thanks for the extensive research :+1: 

> eglInitialize is not needed to be reworked since from the documentation, initializing an already initialized EGLDisplay is a no-op

Looking at [this thread](https://community.amd.com/thread/156626), I'd actually skip `eglInitialize()` and the whole display finding logic -- because the way you have done it here, it might happen that each of those two shared contexts will get created on a different display, and that will most probably not work (different GPU, different driver etc.). Extracting the display from the other context should ensure the same GPU/driver being used (if you create the first context on a llvmpipe, the second context should be there also, and not an Intel iGPU for example).

This also needs a bit of written documentation explaining the behaivor with shared contexts -- when `eglInitialize()`/`Terminate()` is called, what the lifetime of dependent shared contexts has to be, interactions with GPU selections etc. There's a doc section on GPU selection, so this could go after.

> Everything is fine in this case

Not sure -- FWIW, `XOpenDisplay()` could be returning the same handle always (same as EGL) and `XCloseDisplay()` then invalidating it, breaking other contexts that share the same display. Not sure if documentation that proves either eay can be found for this (recommended reading: [I hate Xlib and so should you](https://www.remlab.net/op/xlib.shtml)), but I'd at least put a TODO comment in the code of WindowlessGlxApp stating that it's not doing the same as EGL when it comes to shared contexts, so when someone indeed hits this in the future, this is not forgotten.

aspioupiou:
Ok ! Tell me if you need something from myself on this, this is high priority for myself so I'll be glad to help ;)

src/Magnum/Platform/WindowlessEglApplication.cpp
```diff
@@ -84,6 +84,23 @@ bool extensionSupported(const char* const extensions, Containers::ArrayView<cons
...
+   	    _sharedContext = true; 
+   	}
+   	/**
+     * The user provided only the shared context or display, or provided both but one of them is wrong
+     */
+    else if (configuration.sharedContext() == EGL_NO_CONTEXT && configuration.sharedDisplay() != EGL_NO_DISPLAY ||
+             configuration.sharedContext() != EGL_NO_CONTEXT && configuration.sharedDisplay() == EGL_NO_DISPLAY)
+   {
+     	Error{} << "Platform::WindowlessEglContext: Either context or display is not valid. Context is " << configuration.sharedContext() << " Display is " << configuration.sharedDisplay();  	    
+   }
```
mosra, line 103: Is it just me or is the indentation drunk here? :D Same below.

src/Magnum/Platform/WindowlessEglApplication.h
```diff
@@ -276,16 +276,21 @@ class WindowlessEglContext::Configuration {
...
+         *
+         * @note @p display must be provided to avoid errors on destruction, since 
+         * the display is shared among shared-context on EGL.
          * @see @ref WindowlessEglContext::glContext(),
          *      @ref WindowlessEglApplication::glContext()
          * @requires_gles Context sharing is not available in WebGL.
          */
-        Configuration& setSharedContext(EGLContext context) {
+        Configuration& setSharedContext(EGLContext context, EGLDisplay display) {
             _sharedContext = context;
```
mosra, line 292: I'd add a 

```cpp
CORRADE_ASSERT((context == EGL_NO_CONTEXT) == (display == EGL_NO_DISPLAY),
    "Platform::WindowlessEglContext::Configuration::setSharedContext(): either both context has to be valid or both invalid.", *this);
```

here, moving the implementation to the cpp, and drop the same check inside the `WindowlessEglContext` constructor -- better check this as early as possible, and since it's a programmer error, it should be an assert and not an Error.

src/Magnum/Platform/WindowlessEglApplication.h
```diff
@@ -276,16 +276,21 @@ class WindowlessEglContext::Configuration {
...
          * no sharing.
+         *
+         * @note @p display must be provided to avoid errors on destruction, since 
+         * the display is shared among shared-context on EGL.
          * @see @ref WindowlessEglContext::glContext(),
          *      @ref WindowlessEglApplication::glContext()
          * @requires_gles Context sharing is not available in WebGL.
          */
-        Configuration& setSharedContext(EGLContext context) {
+        Configuration& setSharedContext(EGLContext context, EGLDisplay display) {
```
mosra, line 291: I'd make the display first and context second here, also reorder the getters and private members the same way -- those two are always passed as display first, context second to other EGL functions so this should follow that.

src/Magnum/Platform/WindowlessEglApplication.cpp
```diff
@@ -152,7 +169,7 @@ WindowlessEglContext::WindowlessEglContext(const Configuration& configuration, G
         Containers::Array<EGLDeviceEXT> devices{configuration.device() + 1};
         CORRADE_INTERNAL_ASSERT_OUTPUT(eglQueryDevices(configuration.device() + 1, devices, &count));
 
-        if(!(_display = reinterpret_cast<EGLDisplay(*)(EGLenum, void*, const EGLint*)>(eglGetProcAddress("eglGetPlatformDisplayEXT"))(EGL_PLATFORM_DEVICE_EXT, devices[configuration.device()], nullptr))) {
+   	   if(!_sharedContext && !(_display = reinterpret_cast<EGLDisplay(*)(EGLenum, void*, const EGLint*)>(eglGetProcAddress("eglGetPlatformDisplayEXT"))(EGL_PLATFORM_DEVICE_EXT, devices[configuration.device()], nullptr))) {
```
mosra, line 172: Why not skip the whole branch here? If we already have a display, none of this needs to be done.

src/Magnum/Platform/WindowlessEglApplication.h
aspioupiou: Corrected !

src/Magnum/Platform/WindowlessEglApplication.cpp
aspioupiou: Corrected !

src/Magnum/Platform/WindowlessEglApplication.h
aspioupiou: Corrected as well

mosra:
Merged with a bunch of changes and doc clarifications as 64d0702277ac9988c3b92a495466c75f5e931135.

PR #438, opened 2020-05-02
Trade: Add (ambient|specular|diffuse|normal)CoordinateSet() properties to PhongMaterialData

Squareys:
Hi @mosra !

As promised, here's the PR for the texture layer property for material data.
PR for TinyGltf incoming! :)

Cheers,
Jonathan

### DIFF
--- a/src/Magnum/Trade/PhongMaterialData.cpp
+++ b/src/Magnum/Trade/PhongMaterialData.cpp
@@ -31,21 +31,33 @@ namespace Magnum { namespace Trade {
 
 using namespace Math::Literals;
 
-PhongMaterialData::PhongMaterialData(const Flags flags, const Color4& ambientColor, const UnsignedInt ambientTexture, const Color4& diffuseColor, const UnsignedInt diffuseTexture, const Color4& specularColor, const UnsignedInt specularTexture, const UnsignedInt normalTexture, const Matrix3& textureMatrix, const MaterialAlphaMode alphaMode, const Float alphaMask, const Float shininess, const void* const importerState) noexcept: AbstractMaterialData{MaterialType::Phong, AbstractMaterialData::Flag(UnsignedShort(flags)), alphaMode, alphaMask, importerState}, _ambientColor{ambientColor}, _diffuseColor{diffuseColor}, _specularColor{specularColor}, _shininess{shininess} {
+PhongMaterialData::PhongMaterialData(const Flags flags, const Color4& ambientColor, const UnsignedInt ambientTexture, const UnsignedInt ambientCoordinateSet, const Color4& diffuseColor, const UnsignedInt diffuseTexture, const UnsignedInt diffuseCoordinateSet, const Color4& specularColor, const UnsignedInt specularTexture, const UnsignedInt specularCoordinateSet, const UnsignedInt normalTexture, const UnsignedInt normalCoordinateSet, const Matrix3& textureMatrix, const MaterialAlphaMode alphaMode, const Float alphaMask, const Float shininess, const void* const importerState) noexcept: AbstractMaterialData{MaterialType::Phong, AbstractMaterialData::Flag(UnsignedShort(flags)), alphaMode, alphaMask, importerState}, _ambientColor{ambientColor}, _diffuseColor{diffuseColor}, _specularColor{specularColor}, _shininess{shininess} {
     CORRADE_ASSERT(!(flags & Flag::TextureTransformation) || (flags & (Flag::AmbientTexture|Flag::DiffuseTexture|Flag::SpecularTexture|Flag::NormalTexture)),
         "Trade::PhongMaterialData: texture transformation enabled but the material has no textures", );
+    CORRADE_ASSERT((flags & Flag::TextureCoordinateSets) || (ambientCoordinateSet == 0 && diffuseCoordinateSet == 0 && specularCoordinateSet == 0 && normalCoordinateSet == 0),
+        "PhongMaterialData::PhongMaterialData: non-zero texture coordinate sets require Flag::TextureCoordinateSets to be enabled", );
 
-    if(flags & Flag::AmbientTexture)
+    if(flags & Flag::AmbientTexture) {
         _ambientTexture = ambientTexture;
-    if(flags & Flag::DiffuseTexture)
+        _ambientCoordinateSet = ambientCoordinateSet;
+    }
+    if(flags & Flag::DiffuseTexture) {
         _diffuseTexture = diffuseTexture;
-    if(flags & Flag::SpecularTexture)
+        _diffuseCoordinateSet = diffuseCoordinateSet;
+    }
+    if(flags & Flag::SpecularTexture) {
         _specularTexture = specularTexture;
-    if(flags & Flag::NormalTexture)
+        _specularCoordinateSet = specularCoordinateSet;
+    }
+    if(flags & Flag::NormalTexture) {
         _normalTexture = normalTexture;
-    if(flags  & Flag::TextureTransformation)
+        _normalCoordinateSet = normalCoordinateSet;
+    }
+    if(flags & Flag::TextureTransformation)
         _textureMatrix = textureMatrix;
 }
+
+PhongMaterialData::PhongMaterialData(const Flags flags, const Color4& ambientColor, const UnsignedInt ambientTexture, const Color4& diffuseColor, const UnsignedInt diffuseTexture, const Color4& specularColor, const UnsignedInt specularTexture, const UnsignedInt normalTexture, const Matrix3& textureMatrix, const MaterialAlphaMode alphaMode, const Float alphaMask, const Float shininess, const void* const importerState) noexcept: PhongMaterialData{flags, ambientColor, ambientTexture, 0, diffuseColor, diffuseTexture, 0, specularColor, specularTexture, 0, normalTexture, 0, textureMatrix, alphaMode, alphaMask, shininess, importerState} {}
 
 #ifdef MAGNUM_BUILD_DEPRECATED
 PhongMaterialData::PhongMaterialData(const Flags flags, const MaterialAlphaMode alphaMode, const Float alphaMask, const Float shininess, const void* const importerState) noexcept: PhongMaterialData{flags, 0x000000ff_rgbaf, {}, 0xffffffff_rgbaf, {}, 0xffffffff_rgbaf, {}, {}, {}, alphaMode, alphaMask, shininess, importerState} {}
@@ -69,6 +81,11 @@ UnsignedInt& PhongMaterialData::ambientTexture() {
 }
 #endif
 
+UnsignedInt PhongMaterialData::ambientCoordinateSet() const {
+    CORRADE_ASSERT(flags() & Flag::AmbientTexture, "Trade::PhongMaterialData::ambientCoordinateSet(): the material doesn't have an ambient texture", {});
+    return _ambientCoordinateSet;
+}
+
 UnsignedInt PhongMaterialData::diffuseTexture() const {
     CORRADE_ASSERT(flags() & Flag::DiffuseTexture, "Trade::PhongMaterialData::diffuseTexture(): the material doesn't have a diffuse texture", {});
     return _diffuseTexture;
@@ -80,6 +97,11 @@ UnsignedInt& PhongMaterialData::diffuseTexture() {
     return _diffuseTexture;
 }
 #endif
+
+UnsignedInt PhongMaterialData::diffuseCoordinateSet() const {
+    CORRADE_ASSERT(flags() & Flag::DiffuseTexture, "Trade::PhongMaterialData::diffuseCoordinateSet(): the material doesn't have a diffuse texture", {});
+    return _diffuseCoordinateSet;
+}
 
 UnsignedInt PhongMaterialData::specularTexture() const {
     CORRADE_ASSERT(flags() & Flag::SpecularTexture, "Trade::PhongMaterialData::specularTexture(): the material doesn't have a specular texture", {});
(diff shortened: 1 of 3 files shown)


src/Magnum/Trade/PhongMaterialData.cpp
```diff
@@ -47,6 +47,17 @@ PhongMaterialData::PhongMaterialData(const Flags flags, const Color4& ambientCol
         _textureMatrix = textureMatrix;
 }
 
+PhongMaterialData::PhongMaterialData(Flags flags, const Color4& ambientColor, UnsignedInt ambientTexture, UnsignedInt ambientTextureLayer, const Color4& diffuseColor, UnsignedInt diffuseTexture, UnsignedInt diffuseTextureLayer, const Color4& specularColor, UnsignedInt specularTexture, UnsignedInt specularTextureLayer, UnsignedInt normalTexture, UnsignedInt normalTextureLayer, const Matrix3& textureMatrix, MaterialAlphaMode alphaMode, Float alphaMask, Float shininess, const void* importerState) noexcept: PhongMaterialData{flags, ambientColor, ambientTexture, diffuseColor, diffuseTexture, specularColor, specularTexture, normalTexture, textureMatrix, alphaMode, alphaMask, shininess, importerState} {
```
mosra, line 50: I would switch the delegation order with the above constructor delegating into this one, supplying `0`s for the ~~layers~~ coordinate sets, and this constructor doing everything.

src/Magnum/Trade/PhongMaterialData.h
```diff
@@ -299,11 +375,15 @@ class MAGNUM_TRADE_EXPORT PhongMaterialData: public AbstractMaterialData {
...
         UnsignedInt _ambientTexture{~UnsignedInt{}};
+        UnsignedInt _ambientTextureLayer{0};
         Color4 _diffuseColor;
         UnsignedInt _diffuseTexture{~UnsignedInt{}};
+        UnsignedInt _diffuseTextureLayer{0};
         Color4 _specularColor;
         UnsignedInt _specularTexture{~UnsignedInt{}};
+        UnsignedInt _specularTextureLayer{0};
         UnsignedInt _normalTexture{~UnsignedInt{}};
+        UnsignedInt _normalTextureLayer{0};
```
mosra, line 386: Would do the same as for textures here, i.e.

```suggestion
        UnsignedInt _normalTextureLayer{~UnsignedInt{}};
```

so we catch accidental access of invalid values better.

src/Magnum/Trade/PhongMaterialData.h
```diff
@@ -281,6 +348,15 @@ class MAGNUM_TRADE_EXPORT PhongMaterialData: public AbstractMaterialData {
...
         UnsignedInt normalTexture() const;
 
+        /**
+         * @brief Normal texture layer
+         * @m_since_latest
+         *
+         * Available only if the material has @ref Flag::NormalTexture.
+         * @see @ref flags(), @ref AbstractImporter::texture()
+         */
+        UnsignedInt normalTextureLayer() const;
```
mosra, line 358: The "layer" is confusing as it might also mean a layer in a layer texture. What about a "coordinate set" instead? What term does glTF / Assimp use for this?

src/Magnum/Trade/PhongMaterialData.h
```diff
@@ -124,6 +124,49 @@ class MAGNUM_TRADE_EXPORT PhongMaterialData: public AbstractMaterialData {
...
+         * @param alphaMode            Alpha mode. Use
+         *      @ref MaterialAlphaMode::Opaque for a default value.
+         * @param alphaMask            Alpha mask value. Use @cpp 0.5f @ce for a
+         *      default value.
+         * @param shininess            Shininess. Use @cpp 80.0f @ce for a default
+         *      value.
+         * @param importerState        Importer-specific state
+         * @m_since_latest
+         */
+        explicit PhongMaterialData(Flags flags, const Color4& ambientColor, UnsignedInt ambientTexture, UnsignedInt ambientTextureLayer, const Color4& diffuseColor, UnsignedInt diffuseTexture, UnsignedInt diffuseTextureLayer, const Color4& specularColor, UnsignedInt specularTexture, UnsignedInt specularTextureLayer, UnsignedInt normalTexture, UnsignedInt normalTextureLayer, const Matrix3& textureMatrix, MaterialAlphaMode alphaMode, Float alphaMask, Float shininess, const void* importerState = nullptr) noexcept;
```
mosra, line 167: Because custom coordinate sets need explicit support in shaders the same way as texture matrix needs explicit support in a shader, I'm thinking that this should be an extra flag -- for example

```cpp
            /**
             * The material uses non-default texture coordinate sets
             * @m_since_latest
             */
            TextureCoordinateSets = 1 << 6
```

and then in the constructor asserting that non-zero sets are used only if this flag is present (hmm, the textureMatrix it's ignored if the corresponding flag is not there but I think an assert is better. Would be great if you could change that to an assert also.).

Besides that, there's currently no builtin support for this anywhere, not even in `compile()`. Don't you need that also? Back when I did attribute allocation I [reserved some space for additional texture coordinates](https://doc.magnum.graphics/magnum/structMagnum_1_1Shaders_1_1Generic.html), but those still need to be made real and then used in `compile()`. Without that, this PR has very little use on its own :)

Actually, how's the mapping meant to be done on the shader side? The shader gets a mesh with (let's say) 3 sets of texture coordinates, how can it know which is which? How do you do this in your shaders?

src/Magnum/Trade/PhongMaterialData.h
```diff
@@ -124,6 +124,49 @@ class MAGNUM_TRADE_EXPORT PhongMaterialData: public AbstractMaterialData {
...
+
+
```
mosra, line 169: Superfluous newline here.

src/Magnum/Trade/PhongMaterialData.h
```diff
@@ -124,6 +124,49 @@ class MAGNUM_TRADE_EXPORT PhongMaterialData: public AbstractMaterialData {
          */
```
mosra, line 124: At the top of the class definition, there's an enum marked with `/** @todo what is this for?! */`. It was probably an initial attempt at this. Can you delete that thing?

src/Magnum/Trade/PhongMaterialData.h
Squareys: I assume this would be an uninitialized PhongMaterialData? In Importers where this is not set explicitly, it should be 0 anyway... unless the flag suggested below should be preventing the access of these entirely?

mosra: > the flag suggested below should be preventing the access of these entirely

Yes, basically the same behavior as for `_normalTexture` etc.

src/Magnum/Trade/PhongMaterialData.cpp
```diff
@@ -31,22 +31,34 @@ namespace Magnum { namespace Trade {
...
+PhongMaterialData::PhongMaterialData(Flags flags, const Color4& ambientColor, UnsignedInt ambientTexture, UnsignedInt ambientCoordinateSet, const Color4& diffuseColor, UnsignedInt diffuseTexture, UnsignedInt diffuseCoordinateSet, const Color4& specularColor, UnsignedInt specularTexture, UnsignedInt specularCoordinateSet, UnsignedInt normalTexture, UnsignedInt normalCoordinateSet, const Matrix3& textureMatrix, MaterialAlphaMode alphaMode, Float alphaMask, Float shininess, const void* importerState) noexcept: AbstractMaterialData{MaterialType::Phong, AbstractMaterialData::Flag(UnsignedShort(flags)), alphaMode, alphaMask, importerState}, _ambientColor{ambientColor}, _diffuseColor{diffuseColor}, _specularColor{specularColor}, _shininess{shininess} {
```
mosra, line 34:
```suggestion
PhongMaterialData::PhongMaterialData(const Flags flags, const Color4& ambientColor, const UnsignedInt ambientTexture, const UnsignedInt ambientCoordinateSet, const Color4& diffuseColor, const UnsignedInt diffuseTexture, const UnsignedInt diffuseCoordinateSet, const Color4& specularColor, const UnsignedInt specularTexture, const UnsignedInt specularCoordinateSet, const UnsignedInt normalTexture, const UnsignedInt normalCoordinateSet, const Matrix3& textureMatrix, const MaterialAlphaMode alphaMode, const Float alphaMask, const Float shininess, const void* const importerState) noexcept: AbstractMaterialData{MaterialType::Phong, AbstractMaterialData::Flag(UnsignedShort(flags)), alphaMode, alphaMask, importerState}, _ambientColor{ambientColor}, _diffuseColor{diffuseColor}, _specularColor{specularColor}, _shininess{shininess} {
``` 

:P

src/Magnum/Trade/PhongMaterialData.h
```diff
@@ -299,11 +373,15 @@ class MAGNUM_TRADE_EXPORT PhongMaterialData: public AbstractMaterialData {
...
         UnsignedInt _ambientTexture{~UnsignedInt{}};
+        UnsignedInt _ambientCoordinateSet{UnsignedInt{}};
         Color4 _diffuseColor;
         UnsignedInt _diffuseTexture{~UnsignedInt{}};
+        UnsignedInt _diffuseCoordinateSet{UnsignedInt{}};
         Color4 _specularColor;
         UnsignedInt _specularTexture{~UnsignedInt{}};
+        UnsignedInt _specularCoordinateSet{UnsignedInt{}};
         UnsignedInt _normalTexture{~UnsignedInt{}};
+        UnsignedInt _normalCoordinateSet{UnsignedInt{}};
```
mosra, line 384:
```suggestion
        UnsignedInt _normalCoordinateSet;
```

Here it doesn't have to be set to anything anymore because it's initialized always (sorry)

src/Magnum/Trade/Test/MaterialDataTest.cpp
```diff
@@ -179,23 +197,27 @@ void MaterialDataTest::constructMovePhong() {
...
-        PhongMaterialData::Flag::AmbientTexture|PhongMaterialData::Flag::DiffuseTexture|PhongMaterialData::Flag::SpecularTexture|PhongMaterialData::Flag::NormalTexture|PhongMaterialData::Flag::TextureTransformation,
-        0x111111_rgbf, 1,
-        0xeebbff_rgbf, 42,
-        0xacabad_rgbf, 24, 17,
+        PhongMaterialData::Flag::AmbientTexture|PhongMaterialData::Flag::DiffuseTexture|PhongMaterialData::Flag::SpecularTexture|PhongMaterialData::Flag::NormalTexture|PhongMaterialData::Flag::TextureTransformation|PhongMaterialData::Flag::TextureCoordinateSets,
+        0x111111_rgbf, 1, 0,
+        0xeebbff_rgbf, 42, 1,
+        0xacabad_rgbf, 24, 2, 17, 3,
         Matrix3::rotation(90.0_degf),
         MaterialAlphaMode::Blend, 0.55f, 13.0f, &a};
```
mosra, line 205: Okay but why is this verified only in a test for move?

For completeness the `constructPhongTextured()` should check that accessing those doesn't result in assert and returns 0, and there should be `constructPhongTexturedCoordinateSets()` that's basically a copy plus it also enables the flag and fills in the coordinate sets.

src/Magnum/Trade/PhongMaterialData.h
```diff
@@ -124,6 +123,48 @@ class MAGNUM_TRADE_EXPORT PhongMaterialData: public AbstractMaterialData {
          */
         explicit PhongMaterialData(Flags flags, const Color4& ambientColor, UnsignedInt ambientTexture, const Color4& diffuseColor, UnsignedInt diffuseTexture, const Color4& specularColor, UnsignedInt specularTexture, UnsignedInt normalTexture, const Matrix3& textureMatrix, MaterialAlphaMode alphaMode, Float alphaMask, Float shininess, const void* importerState = nullptr) noexcept;
 
+        /**
+         * @brief Constructor with texture layers
```
mosra, line 127:
```suggestion
         * @brief Construct with non-zero texture coordinate sets
```

src/Magnum/Trade/PhongMaterialData.h
```diff
@@ -124,6 +123,48 @@ class MAGNUM_TRADE_EXPORT PhongMaterialData: public AbstractMaterialData {
...
+         * @param textureMatrix        Texture coordinate transformation. Ignored
+         *      if @p flags doesn't have @ref Flag::TextureTransformation.
+         * @param alphaMode            Alpha mode. Use
+         *      @ref MaterialAlphaMode::Opaque for a default value.
+         * @param alphaMask            Alpha mask value. Use @cpp 0.5f @ce for a
+         *      default value.
+         * @param shininess            Shininess. Use @cpp 80.0f @ce for a default
+         *      value.
+         * @param importerState        Importer-specific state
+         * @m_since_latest
```
mosra, line 164:
```suggestion
         * @m_since_latest
         *
         * If any `*CoordinateSet` is non-zero, expects 
         * @ref Flag::TextureCoordinateSets to be enabled as well.
```

src/Magnum/Trade/PhongMaterialData.h
```diff
@@ -200,6 +241,14 @@ class MAGNUM_TRADE_EXPORT PhongMaterialData: public AbstractMaterialData {
         UnsignedInt& ambientTexture();
         #endif
 
+        /**
+         * @brief Ambient texture layer
+         *
```
mosra, line 246:
```suggestion
        /**
         * @brief Ambient texture coordinate set
         * @m_since_latest
         *
```

Same below

src/Magnum/Trade/PhongMaterialData.h
```diff
@@ -124,6 +123,48 @@ class MAGNUM_TRADE_EXPORT PhongMaterialData: public AbstractMaterialData {
...
+         * @brief Constructor with texture layers
+         * @param flags                Material flags
+         * @param ambientColor         Ambient color. Use
+         *      @cpp 0x000000ff_rgbaf @ce for a default value for a
+         *      non-textured material and @cpp 0xffffffff_rgbaf @ce for a
+         *      default value for a textured material.
+         * @param ambientTexture       Ambient texture ID. Ignored if @p flags
+         *      doesn't have @ref Flag::AmbientTexture
+         * @param ambientCoordinateSet  Ambient texture layer. Ignored if @p flags
+         *      doesn't have @ref Flag::AmbientTexture
```
mosra, line 136:
```suggestion
         * @param ambientCoordinateSet  Ambient texture coordinate set. Ignored if 
         *      @p flags doesn't have @ref Flag::AmbientTexture
```

Same below

src/Magnum/Trade/PhongMaterialData.h
```diff
@@ -124,6 +123,48 @@ class MAGNUM_TRADE_EXPORT PhongMaterialData: public AbstractMaterialData {
          */
```
mosra, line 123: Here I would mention that

```suggestion
         *
         * All `*CoordinateSet`s are implicitly set to @cpp 0 @ce.
         */
```

mosra:
Merged as b8cc3f766af1c925cd336f4d7ab78a4616b1dbf4, thank you! :tada: 

(Hah, now I realized the commit still refers to texture *layers*. Well, too late.)

Squareys:
Thanks for merging! 🎉 

Sorry about the commit message 😅

PR #441, opened 2020-05-03
Skinned Mesh attributes

Squareys:
Hey @mosra !

here's some initial draft for Skinning stuff to start discussion.

Cheers,
Jonathan

**Open Questions**
 1.Naming of Attributes

**TODOs**
 - [x] Add `Weights` and `JointIds` to AttributeNames and Generic attributes.
 - [x] MeshData & Testing
 - [ ] Documentation?

### DIFF
--- a/src/Magnum/Shaders/Generic.h
+++ b/src/Magnum/Shaders/Generic.h
@@ -126,13 +126,13 @@ both bitangents and object ID for instancing, \n
 <tr>
 <td>6</td>
 <td colspan="3">
-* *Reserved* --- vertex weights
+@ref Weights
 </td>
 </tr>
 <tr>
 <td>7</td>
 <td colspan="3">
-* *Reserved* --- bone indices
+@ref JointIds
 </td>
 </tr>
 <tr>
@@ -150,13 +150,13 @@ both bitangents and object ID for instancing, \n
 <tr>
 <td>10</td>
 <td colspan="2">
-* *Reserved* --- 2nd vertex weights
+@ref SecondaryWeights
 </td>
 </tr>
 <tr>
 <td>11</td>
 <td colspan="2">
-* *Reserved* --- 2nd bone indices
+@ref SecondaryJointIds
 </td>
 </tr>
 <tr>
@@ -381,7 +381,21 @@ template<UnsignedInt dimensions> struct Generic {
      */
     typedef GL::Attribute<5, Vector3> Normal;
 
-    /* 6, 7 reserved for vertex weights / bone IDs */
+    /**
+     * @brief Skin weights
+     *
+     * @ref Magnum::Vector4 "Vector4", four weights of influence per @ref JointIndices.
+     * Corresponds to @ref Trade::MeshAttribute::Weights.
+     */
+    typedef GL::Attribute<6, Vector4> Weights;
+
+    /**
+     * @brief Joint IDs
+     *
+     * @ref Magnum::Vector4ui "Vector4ui", four joint indices that may affect the vertex.
+     * Corresponds to @ref Trade::MeshAttribute::JointIds.
+     */
+    typedef GL::Attribute<7, Vector4ui> JointIds;
 
     /**
      * @brief (Instanced) transformation matrix
@@ -478,6 +492,11 @@ struct BaseGeneric {
     #ifndef MAGNUM_TARGET_GLES2
     typedef GL::Attribute<4, UnsignedInt> ObjectId;
     #endif
+    typedef GL::Attribute<6, Vector4> Weights;
+    typedef GL::Attribute<7, Vector4> JointIds;
+
+    typedef GL::Attribute<10, Vector4> SecondaryWeights;
+    typedef GL::Attribute<11, Vector4> SecondaryJointIds;
 
     typedef GL::Attribute<15, Vector2> TextureOffset;
 
@@ -506,7 +525,6 @@ template<> struct Generic<3>: BaseGeneric {
     typedef GL::Attribute<3, Vector4> Tangent4;
     typedef GL::Attribute<4, Vector3> Bitangent; /* also ObjectId */
     typedef GL::Attribute<5, Vector3> Normal;
-    /* 6, 7 reserved for vertex weights / bone IDs */
 
     typedef GL::Attribute<8, Matrix4> TransformationMatrix;
     /* 9, 10, 11 occupied by TransformationMatrix */
--- a/src/Magnum/Trade/MeshData.h
+++ b/src/Magnum/Trade/MeshData.h
@@ -106,7 +106,7 @@ enum class MeshAttribute: UnsignedShort {
 
     /**
      * Normal. Type is usually @ref VertexFormat::Vector3, but can be also
-     * @ref VertexFormat::Vector3h. @ref VertexFormat::Vector3bNormalized or
+     * @ref VertexFormat::Vector3h, @ref VertexFormat::Vector3bNormalized or
      * @ref VertexFormat::Vector3sNormalized. Corresponds to
      * @ref Shaders::Generic::Normal.
      * @see @ref MeshData::normalsAsArray()
--- a/src/Magnum/Trade/MeshData.h
+++ b/src/Magnum/Trade/MeshData.h
@@ -148,6 +148,23 @@ enum class MeshAttribute: UnsignedShort {
     ObjectId,
 
     /**
+     * Weights. Type is usually @ref VertexFormat::Vector4, but can be also
+     * @ref VertexFormat::Vector4h, @ref VertexFormat::Vector4ubNormalized
+     * or @ref VertexFormat::Vector4usNormalized.
+     * Corresponds to @ref Shaders::Generic::Weights.
+     * @see @ref MeshData::weightsAsArray()
+     */
+    Weights,
+
+    /**
+     * Joint IDs. Type is usually @ref VertexFormat::Vector4ui, but can be also
+     * @ref VertexFormat::Vector4us or @ref VertexFormat::Vector4ub.
+     * Corresponds to @ref Shaders::Generic::JointIds.
+     * @see @ref MeshData::jointIdsAsArray()
+     */
+    JointIds,
+
+    /**
      * This and all higher values are for importer-specific attributes. Can be
      * of any type. See documentation of a particular importer for details.
      * @see @ref isMeshAttributeCustom(MeshAttribute)
@@ -613,10 +630,11 @@ the @ref Primitives library.
 The simplest usage is through the convenience functions @ref positions2DAsArray(),
 @ref positions3DAsArray(), @ref tangentsAsArray(), @ref bitangentsAsArray(),
 @ref normalsAsArray(), @ref tangentsAsArray(), @ref textureCoordinates2DAsArray(),
-@ref colorsAsArray() and @ref objectIdsAsArray(). Each of these takes an index
-(as there can be multiple sets of texture coordinates, for example) and you're
-expected to check for attribute presence first with either @ref hasAttribute()
-or @ref attributeCount(MeshAttribute) const:
+@ref colorsAsArray(), @ref objectIdsAsArray(), @ref weightsAsArray() and
+@ref jointIdsAsArray(). Each of these takes an index (as there can be multiple
+sets of texture coordinates, for example) and you're expected to check for
+attribute presence first with either @ref hasAttribute() or
+@ref attributeCount(MeshAttribute) const:
 
 @snippet MagnumTrade.cpp MeshData-usage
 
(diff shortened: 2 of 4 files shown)


src/Magnum/Shaders/Generic.h
```diff
@@ -358,7 +358,26 @@ template<UnsignedInt dimensions> struct Generic {
...
+    /**
+     * @brief Joint indices
+     *
+     * @ref Magnum::Vector4 "Vector4", four weights of influence per @ref JointIndices.
+     * Corresponds to @ref Trade::MeshAttribute::Weights.
+     */
+    // TODO: E.g. gltf supports importing multiple sets of 4 indices (JOINTS_0 [0;3], JOINTS_1 [4;7])
+    // I suppose these would require special shaders and it's fine to only support 4 for the GenericShaders
+    // TODO: This is called "Joints" in gltf.
+    typedef GL::Attribute<7, Vector4> JointIndices;
```
mosra, line 380:
```suggestion
    typedef GL::Attribute<7, Vector4> JointIds;
```

Also isn't this an integer attribute, when it's a ID? All other attribs are singular, so maybe make this `JointId` and `Weight` also? Kinda on the fence about this tho, because there's `TextureCoordinates` also.

src/Magnum/Shaders/Generic.h
```diff
@@ -358,7 +358,26 @@ template<UnsignedInt dimensions> struct Generic {
...
 
-    /* 6, 7 reserved for vertex weights / bone IDs */
+    /**
+     * @brief Skin weights
+     *
+     * @ref Magnum::Vector4 "Vector4", four weights of influence per @ref JointIndices.
+     * Corresponds to @ref Trade::MeshAttribute::Weights.
+     */
+    // TODO: E.g. gltf supports importing multiple sets of 4 weights (WEIGHTS_0 [0;3], WEIGHTS_1 [4;7])
+    // I suppose these would require special shaders and it's fine to only support 4 for the GenericShaders
```
mosra, line 368: Yep, and that's why I reserved the "2nd vertex weights" etc above. This would be then a `SecondaryWeights` attribute, and yes the shader needs to explicitly support that (it's not fine to hard-cap the support to just 4, no :P).

Hmm, now i'm thinking if it's actually possible for the shader to access both primary and secondary weights through the same accessor (maybe through a `mat2x4` where the first column is bound to id 6 and the second to id 10?)

src/Magnum/Trade/MeshData.h
```diff
@@ -1717,6 +1736,54 @@ class MAGNUM_TRADE_EXPORT MeshData {
...
+         * Convenience alternative to @ref attribute(MeshAttribute, UnsignedInt) const
+         * with @ref MeshAttribute::JointIndices as the first argument. Converts
+         * the joint indices array from an arbitrary underlying type and returns it
+         * in a newly-allocated array. Expects that the vertex format is *not*
+         * implementation-specific, in that case you can only access the
+         * attribute via the typeless @ref attribute(MeshAttribute, UnsignedInt) const.
+         * @see @ref jointIndicesInto(), @ref attributeFormat(),
+         *      @ref isVertexFormatImplementationSpecific()
+         */
+        Containers::Array<Vector4us> jointIndicesAsArray(UnsignedInt id = 0) const;
```
mosra, line 1775: I think should always be the largest type (even if it will never be as big):

```suggestion
        Containers::Array<Vector4ui> jointIndicesAsArray(UnsignedInt id = 0) const;
```

src/Magnum/Shaders/Generic.h
Squareys: Since it's always four at a time (i.e. four joints that might influence a vertex at any given time), I think conveying it's "more than one" through the name would make sence in my oppinion.

src/Magnum/Shaders/Generic.h
Squareys: Why would it need to access both weight sets through the same accessor? The Gltf 2.0 overview seems to use a very "unrolled" way of multiplying weights with bone transforms, no matrix multiplications there:

```glsl

uniform mat4 u_jointMatrix[12];
in vec4 position;
in vec4 jointIds;
in vec4 weights;

void main(void) {
    mat4 skinMatrix =
        weights.x*jointMatrix[int(jointIds.x)] +
        weights.y*jointMatrix[int(jointIds.x)] +
        weights.z*jointMatrix[int(jointIds.x)] +
        weights.w*jointMatrix[int(jointIds.x)];

    gl_Position = modelViewProjection*skinMatrix*position;
}
```

mosra: Shouldn't that be `jointIds.xyzw` also? I mean the case with

```glsl
in vec4 jointIds;
in vec4 secondaryJointIds;
in vec4 weights;
in vec4 secondaryWeights;
```

is kinda ugly, so I'm wondering if this could be turned into some

```glsl
in float jointIds[8];
in float weights[8];
```

instead, with the `8` being possibly passed in during compilation (so e.g. we can support 5 influences without always doing calculations for 8.

Squareys: > Shouldn't that be jointIds.xyzw also?

Yes! => fixed

> pseudocode

This will lead to some ugly macro stuff if you do it this way. Instead with the duplicate attributes, it'd be just a block like `#ifdef MORE_BONES` which would be added rather than replacing the code that works for 4 joints/2 attributes.

mosra: > ugly macro

why? this is what I wanted to achieve

```glsl
mat4 skinMatrix;
for(int i = 0; i != HOW_MANY_THEM_BONES; ++i)
    skinMatrix += weights[i]*jointMatrices[int(jointIds[i])];
```

Squareys: Wouldn't `mat2x4 jointIds` return a vec4 on `jointIds[0]` though?

mosra: This is becoming endless :sweat_smile: 

Yes it would. I'm referring to the `in float jointIds[8];` above.

src/Magnum/Shaders/Generic.h
```diff
@@ -483,7 +497,8 @@ template<> struct Generic<3>: BaseGeneric {
     typedef GL::Attribute<3, Vector4> Tangent4;
     typedef GL::Attribute<4, Vector3> Bitangent; /* also ObjectId */
     typedef GL::Attribute<5, Vector3> Normal;
-    /* 6, 7 reserved for vertex weights / bone IDs */
+    typedef GL::Attribute<6, Vector4> Weights;
+    typedef GL::Attribute<7, Vector4> JointIds;
```
mosra, line 501:
```suggestion
    typedef GL::Attribute<7, Vector4ui> JointIds;
```

Also, where's the 2D variant?

src/Magnum/Trade/MeshData.h
```diff
@@ -147,6 +147,23 @@ enum class MeshAttribute: UnsignedShort {
...
+     * or @ref VertexFormat::Vector4usNormalized.
+     * Corresponds to @ref Shaders::Generic::Weights.
+     * @see @ref MeshData::weightsAsArray()
+     */
+    Weights,
+
+    /**
+     * Joint IDs. Type is usually @ref VertexFormat::Vector4ui, but can be also
+     * @ref VertexFormat::Vector4us or @ref VertexFormat::Vector4ub.
+     * @ref Shaders::Generic::JointIds.
```
mosra, line 162:
```suggestion
     * Corresponds to @ref Shaders::Generic::JointIds.
```

src/Magnum/Shaders/Generic.h
```diff
@@ -150,13 +150,13 @@ both bitangents and object ID for instancing, \n
 <tr>
 <td>10</td>
 <td colspan="2">
-* *Reserved* --- 2nd vertex weights
+@ref Weights
```
mosra, line 153: This is still wrong -- either add `SecondaryWeights`, `SecondaryJointIds` or revert this and the below change.

src/Magnum/Shaders/Generic.h
```diff
@@ -126,13 +126,13 @@ both bitangents and object ID for instancing, \n
...
 <td colspan="3">
-* *Reserved* --- vertex weights
+@ref Weights
 </td>
 </tr>
 <tr>
 <td>7</td>
 <td colspan="3">
-* *Reserved* --- bone indices
+@ref JointIndices
```
mosra, line 135:
```suggestion
@ref JointIds
```

mosra:
What's left for this to be useful is adding those to `MeshTools::compile()` -- but to test it there, you would need some shader that supports them, and at that point you have implemented skinning completely :)

src/Magnum/Shaders/Generic.h
Squareys: Why wouldn't this work for 2D?

src/Magnum/Shaders/Generic.h
Squareys: Will make it SecondaryWeights 👍

Squareys:
Sure, it looks straight forward enough to do so, and I'll have to eventually anyways, so why not rip off the bandage.

Squareys:
@mosra I started implementing Skinning in Phong to test the additions to compile(), but this would blow up the PR significantly. Could we split Phong shader implementation, compile() and test off into a separate PR?

src/Magnum/Shaders/Generic.h
mosra: Because it's defined only in the 3D specialization of the class ;)

mosra:
Yup, good idea -- the Phong implementation also needs tests on its own, so it makes sense to do all that separately.

Squareys:
Split Phong implementation into Squareys:phong-skinning, will create the PR for that once this is merged.
Rebased this and squashed some commits, I think this is ready for review :)

src/Magnum/Trade/Test/MeshDataTest.cpp
```diff
@@ -2500,6 +2522,90 @@ void MeshDataTest::objectIdsIntoArrayInvalidSize() {
...
+    MeshData data{MeshPrimitive::Points, std::move(vertexData), {MeshAttributeData{MeshAttribute::JointIds, Containers::arrayCast<Vector4ui>(vertexData)}}};
+
+    std::ostringstream out;
+    Error redirectError{&out};
+    Vector4ui destination[2];
+    data.jointIdsInto(destination);
+    CORRADE_COMPARE(out.str(),
+        "Trade::MeshData::jointIdsInto(): expected a view with 3 elements but got 2\n");
+}
+
```
mosra, line 2608: The `attributeNotFound()` and `implementationSpecificVertexFormatWrongAccess()` tests need to be extended to check asserts in the new accessors.

src/Magnum/Trade/MeshData.h
```diff
@@ -147,6 +147,23 @@ enum class MeshAttribute: UnsignedShort {
...
+     */
+    Weights,
+
+    /**
+     * Joint IDs. Type is usually @ref VertexFormat::Vector4ui, but can be also
+     * @ref VertexFormat::Vector4us or @ref VertexFormat::Vector4ub.
+     * Corresponds to @ref Shaders::Generic::JointIds.
+     * @see @ref MeshData::jointIdsAsArray()
+     */
+    JointIds,
```
mosra, line 165: I'm wondering -- would it make sense to support also one/two/three-component weights and joints? I guess it would be useful for file size savings. Can you check if there are example glTF files with less than 4 weights?

This won't complicate the implementation too much, a similar thing is already done for colors for example.

src/Magnum/Shaders/Generic.h
```diff
@@ -478,6 +492,11 @@ struct BaseGeneric {
     #ifndef MAGNUM_TARGET_GLES2
     typedef GL::Attribute<4, UnsignedInt> ObjectId;
     #endif
+    typedef GL::Attribute<6, Vector4> Weights;
+    typedef GL::Attribute<7, Vector4> JointIds;
+
+    typedef GL::Attribute<10, Vector4> SecondaryWeights;
+    typedef GL::Attribute<11, Vector4> SecondaryJointIds;
```
mosra, line 499: Seeing #444, this PR should contain also the `generic.glsl` changes and corresponding additions to `GenericTest::glslMatch()`

src/Magnum/Trade/MeshData.h
Squareys: The accessor is always a 4-component vector, so in case there is a file like this, it would have 4 but the unused components would be zero weight

mosra: In (unextended) glTF yes, but in general not. I don't see why glTF couldn't support `SCALAR`, `VEC2` and `VEC3` also.

mosra: It would be just JointIds always, but allowing also one/two/three-component types instead of just Vectot4ui.

`MeshData::jointIdsInto()` will then need to zero-fill the unused components like `colorsInto()` does for three-component colors or `positions3DInto()` does for two-component positions, yes.

mosra:
Merged with various changes as 7f94a6f914c834e0d693cf2a8d62eb1d40492699 and f447e994d271c8688da3cf7f04a1a2643f42cc21. The primary difference is that `MeshAttribute::JointIds` and `Weights` is now an array attribute in order to allow a single attribute to have any number of components -- two, four but also for example seven. This allows both for better packing in memory (compared to always having 4 or 8 components) as well as easier processing because with > 4 components the attribute isn't split into two. On the other hand it's still possible to have two `JointIds` and two `Weights` for example, and have the second set in a different (e.g., less precise) type.

The builtin shader bindings slots stay four component, with the intention that (a) the shader is either compiled for an appropriate per-vertex component count, (b) a dynamic per-vertex component count is passed to it, or (c) the mesh is repacked to always contain the maximum number of components the shader expects. 

A major change with the builtin attribute is that they have a flipped order compared to this PR -- `JointIds` and `SecondaryJointIds` are 6 and 10 instead of 7 and 11, `Weights` and `SecondaryWeights` are 7 and 11 instead of 6 and 10. This order made more sense to me, i.e. "joint i has a weight j" instead of "weight i is applied to joint j".

PR #442, opened 2020-05-07
Avoid using 100% CPU in an idle GlfwApplication

mosra:
Interestingly enough this went unnoticed since the very first implementation in 2016.

While I'm pretty sure this works, I'm not sure about possible bad interactions, so I'll merge to master only after I have enough confirmations that this still works as expected. Testers welcome, thanks in advance :+1: 

Cc: @Squareys


Squareys:
I just switched to GlfwApplication and didn't notice any issues related to this. I am redrawing the application because of DearImGui anyway, though, so don't think I would.

PR #444, opened 2020-05-11
Phong skinning

Squareys:
Hi @mosra !

As mentioned in #441 , here's the split-off of the Phong implementation.

Best,
Jonathan


src/Magnum/Shaders/Phong.vert
```diff
@@ -155,11 +186,25 @@ out highp vec3 cameraDirection;
 #endif
 
 void main() {
+    #ifdef SKINNING
+    mat4 skinMatrix;
+    int i = 0;
+    for(; i != JOINTS_PER_VERTEX && i != 4; ++i)
+        skinMatrix += weights[i]*jointMatrices[jointIds[i]];
+    #if JOINTS_PER_VERTEX > 4
+    for(i = 0; i != JOINTS_PER_VERTEX - 4 && i != 4; ++i)
```
Squareys, line 195: This should be `i != 8`.

src/Magnum/Shaders/Phong.vert
```diff
@@ -155,11 +186,25 @@ out highp vec3 cameraDirection;
 #endif
 
 void main() {
+    #ifdef SKINNING
+    mat4 skinMatrix;
```
Squareys, line 190: This should be initialized with `weights[0]*jointMatrices[jointIds[0]]`.

mosra:
The shader code is merged as part of 89a6eb22def083c4907add7d007184d868d27cd2. The shader code changed quite significantly since then so I was mostly just cherry-picking interesting bits from here, nevertheless -- thanks for the initial code, saved me a lot of time. The support is now extended to Flat and MeshVisualizer shaders as well, and works in instanced and multi-draw scenarios too.

The `compile()` support is merged as part of c026ac1cb5733c951e24d85868c742eb258e55dd. Compared to what was in this PR it had to deal with the additional complexity of the attributes being arbitrary arrays, and there's a new `compiledPerVertexJointCount()` helper for querying how many components in the builtin attribute slots get actually used in order to subsequently pass that info to the shader.

PR #445, opened 2020-05-17
Add missing UnsignedInt cast.

HaroldReyiz:
First PR so excited!
Hi @mosra, I'm gonna add these casts as I come across them here.


mosra:
Merged as c5574d89e7ea2f7ce5a79bba9675387b7e5004b7, as I assume you didn't run into any other case since :)

PR #449, opened 2020-06-09
CUDA device selection support

erikwijmans:
Add support for CUDA device selection to Magnum's EGL context.  When using `WindowlessEglApplication`, this can be specified via `--magnum-cuda-device`.  Otherwise, whether or not a device is a CUDA device is specified by `setDeviceIsCuda()`.

@mosra Does the CI have an NV gpu to test this with?

### DIFF
--- a/src/Magnum/Platform/WindowlessEglApplication.cpp
+++ b/src/Magnum/Platform/WindowlessEglApplication.cpp
@@ -47,6 +47,10 @@
 typedef void *EGLDeviceEXT;
 #endif
 
+#ifndef EGL_NV_device_cuda
+#define EGL_CUDA_DEVICE_NV 0x323A
+#endif
+
 #ifndef EGL_EXT_platform_device
 #define EGL_PLATFORM_DEVICE_EXT 0x313F
 #endif
@@ -68,6 +72,8 @@ namespace Magnum { namespace Platform {
 
 #ifndef MAGNUM_TARGET_WEBGL
 namespace {
+
+static constexpr UnsignedInt MaxEGLDevices = 128;
 
 bool extensionSupported(const char* const extensions, Containers::ArrayView<const char> extension) {
     CORRADE_INTERNAL_ASSERT(extensions);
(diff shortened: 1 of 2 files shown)


src/Magnum/Platform/WindowlessEglApplication.h
```diff
@@ -261,17 +262,44 @@ class WindowlessEglContext::Configuration {
...
+         * When using @ref WindowlessEglApplication, this is will be set
+         * to true if `--magnum-cuda-device` or `MAGNUM_CUDA_DEVICE` are
+         * specified
+         *
+         * @requires_gles Device selection is not available in WebGL.
+         */
+        Configuration& setDeviceIsCuda(const bool isCuda) {
+            _deviceIsCuda = isCuda;
+            return *this;
+        }
```
mosra, line 301: I would instead add a pair of `cudaDevice()` and `setCudaDevice()`. That makes it closer to the pair of `--magnum-device` / `--magnum-cuda-device` options and makes it easier on when populating the config from code as well, since all you have to do is call just `setCudaDevice()` instead of calling both `setDevice()` and `setCudaDevice()`.

src/Magnum/Platform/WindowlessEglApplication.h
```diff
@@ -313,6 +341,7 @@ class WindowlessEglContext::Configuration {
         UnsignedInt _device;
         EGLDisplay _sharedDisplay = EGL_NO_DISPLAY;
         EGLContext _sharedContext = EGL_NO_CONTEXT;
+        bool _deviceIsCuda = false;
```
mosra, line 344:
```suggestion
        UnsignedInt _device;
        UnsignedInt _cudaDevice = ~UnsignedInt{};
        EGLDisplay _sharedDisplay = EGL_NO_DISPLAY;
        EGLContext _sharedContext = EGL_NO_CONTEXT;
```

=> if `_cudaDevice` is not `~UnsignedInt{}`, it means the user set it, otherwise it'll use `_device`.

src/Magnum/Platform/WindowlessEglApplication.h
```diff
@@ -610,6 +646,7 @@ class WindowlessEglApplication {
         #ifndef MAGNUM_TARGET_WEBGL
         /* These are saved from command-line arguments */
         UnsignedInt _commandLineDevice;
+        bool _commandLineDeviceIsCuda = false;
```
mosra, line 649: Similarly here it would be 

```suggestion
        UnsignedInt _commandLineDevice;
        UnsignedInt _commandLineCudaDevice = ~UnsignedInt{};
```

to have the same handling in both

src/Magnum/Platform/WindowlessEglApplication.h
```diff
@@ -134,6 +134,7 @@ class WindowlessEglContext {
     private:
         #ifndef MAGNUM_TARGET_WEBGL
         bool _sharedContext = false;
+        static constexpr UnsignedInt MAX_DEVICES = 128;
```
mosra, line 137: This is used only in the cpp, so I would move that there into a unnamed namespace, named `MaxCudaDevices`.

src/Magnum/Platform/WindowlessEglApplication.cpp
```diff
@@ -157,12 +157,55 @@ WindowlessEglContext::WindowlessEglContext(const Configuration& configuration, G
...
+                    if(!magnumContext || !(magnumContext->internalFlags() & GL::Context::InternalFlag::GpuValidation))
+                        e << Debug::nospace << "; enable --magnum-gpu-validation to see additional info";
+                    return;
+                }
+
+                for (selectedDevice = 0; selectedDevice < UnsignedInt(count); ++selectedDevice) {
+                    if(magnumContext && (magnumContext->internalFlags() & GL::Context::InternalFlag::GpuValidation))
+                        Debug{} << "Platform::WindowlessEglApplication: eglQueryDeviceStringEXT(EGLDevice=" << Debug::nospace << selectedDevice << Debug::nospace << "):" << eglQueryDeviceStringEXT(devices[selectedDevice], EGL_EXTENSIONS);
+                    EGLAttrib cudaDeviceNumber;
+                    if (eglQueryDeviceAttribEXT(devices[selectedDevice], EGL_CUDA_DEVICE_NV, &cudaDeviceNumber) == EGL_TRUE && cudaDeviceNumber == configuration.device()) {
```
mosra, line 181: `EGL_CUDA_DEVICE_NV` might not be available on all platforms, to be sure I'd add it to the other defines at the top of the file:

```cpp
#ifndef EGL_NV_device_cuda
#define EGL_CUDA_DEVICE_NV 0x323A
#endif
```

src/Magnum/Platform/WindowlessEglApplication.cpp
```diff
@@ -157,12 +157,55 @@ WindowlessEglContext::WindowlessEglContext(const Configuration& configuration, G
...
             }
 
-            /* Assuming the same thing won't suddenly start failing when called
-               the second time */
-            Containers::Array<EGLDeviceEXT> devices{configuration.device() + 1};
-            CORRADE_INTERNAL_ASSERT_OUTPUT(eglQueryDevices(configuration.device() + 1, devices, &count));
+            UnsignedInt selectedDevice;
+            Containers::Array<EGLDeviceEXT> devices;
+
+            if (configuration.deviceIsCuda()) {
```
mosra, line 163: Following the API changes, this would be 

```suggestion
            if (configuration.cudaDevice() != ~UnsignedInt{}) {
```

src/Magnum/Platform/WindowlessEglApplication.cpp
```diff
@@ -157,12 +157,55 @@ WindowlessEglContext::WindowlessEglContext(const Configuration& configuration, G
...
+                auto eglQueryDeviceAttribEXT = reinterpret_cast<EGLBoolean (*)(EGLDeviceEXT, EGLint, EGLAttrib*)>(eglGetProcAddress("eglQueryDeviceAttribEXT"));
+                auto eglQueryDeviceStringEXT = reinterpret_cast<const char* (*)(EGLDeviceEXT, EGLint)>(eglGetProcAddress("eglQueryDeviceStringEXT"));
+                
+                if (!eglQueryDeviceAttribEXT || !eglQueryDeviceStringEXT) {
+                    Error e;
+                    e << "Platform::WindowlessEglApplication: CUDA device selection requires 'eglQueryDeviceAttribEXT' method, likely a driver issue";
+                    if(!magnumContext || !(magnumContext->internalFlags() & GL::Context::InternalFlag::GpuValidation))
+                        e << Debug::nospace << "; enable --magnum-gpu-validation to see additional info";
+                    return;
+                }
```
mosra, line 175: Like we discussed yesterday, test for either of `EGL_device_query` / `EGL_EXT_device_base` instead? Mentioning those exts in the error message would also be more googleable than the cryptic method name I think :)

src/Magnum/Platform/WindowlessEglApplication.cpp
```diff
@@ -157,12 +157,55 @@ WindowlessEglContext::WindowlessEglContext(const Configuration& configuration, G
...
+
+                for (selectedDevice = 0; selectedDevice < UnsignedInt(count); ++selectedDevice) {
+                    if(magnumContext && (magnumContext->internalFlags() & GL::Context::InternalFlag::GpuValidation))
```
mosra, line 178:
```suggestion
                    if(magnumContext && (magnumContext->internalFlags() & GL::Context::InternalFlag::DisplayVerboseInitializationLog))
```

The message is not really coming from KHR_debug or EGL_KHR_debug, this way it would get printed when you run with `--magnum-log verbose` (which already prints other things like device count, when enabled).

src/Magnum/Platform/WindowlessEglApplication.cpp
```diff
@@ -460,6 +505,11 @@ WindowlessEglApplication::WindowlessEglApplication(const Arguments& arguments, N
         _commandLineDevice = 0;
     else
         _commandLineDevice = args.value<UnsignedInt>("device");
+
+    if (!args.value("cuda-device").empty()) {
+        _commandLineDevice = args.value<UnsignedInt>("cuda-device");
+        _commandLineDeviceIsCuda = true;
+    }
```
mosra, line 512:
```suggestion
    if (!args.value("cuda-device").empty())
        _commandLineCudaDevice = args.value<UnsignedInt>("cuda-device");
    else
        _commandLineCudaDevice = ~UnsignedInt{};
```

src/Magnum/Platform/WindowlessEglApplication.cpp
```diff
@@ -476,7 +526,7 @@ bool WindowlessEglApplication::tryCreateContext(const Configuration& configurati
     Configuration mergedConfiguration{configuration};
     #ifndef MAGNUM_TARGET_WEBGL
     if(!mergedConfiguration.device())
-        mergedConfiguration.setDevice(_commandLineDevice);
+        mergedConfiguration.setDevice(_commandLineDevice).setDeviceIsCuda(_commandLineDeviceIsCuda);
```
mosra, line 529:
```suggestion
        mergedConfiguration.setDevice(_commandLineDevice).setCudaDevice(_commandLineCudaDevice);
```

ok sorry at this point i'm just using you as a remote keyboard :sweat_smile:

src/Magnum/Platform/WindowlessEglApplication.cpp
erikwijmans: Ah yeah, I see what you meant by that now :)

src/Magnum/Platform/WindowlessEglApplication.h
erikwijmans: Will do, although it will be `MaxEGLDevices` as non-cuda devices will also show up in the query of all devices.

mosra:
:tada:

PR #454, opened 2020-07-03
Math: add fmod function

pezcode:
This adds `Math::fmod()` to calculate a floating point division remainder for scalar and vector Math types by wrapping `std::fmod`.

Not entirely sure about the docs portion and especially the Latex (is `\operatorname` supported?).


mosra:
Thanks!

PR #461, opened 2020-07-18
Add binomial coefficient function

Janos95:
This patch introduces a binomial coefficient function.

### DIFF
--- a/src/Magnum/Math/Functions.h
+++ b/src/Magnum/Math/Functions.h
@@ -416,10 +416,10 @@ template<std::size_t size, class T> inline Vector<size, T> ceil(const Vector<siz
 @brief Computes a Binomial Coefficient
 
 Given n elements computes the number of ways to choose k elements
-without replacement from n elements. The number of ways to do this is
+without replacement. The number of ways to do this is
 given by
 @f[
-    \frac{n}{k} = \frac{n! \cdot (n-k)!}{k!} = \frac{n \cdot (n-1) \cdots (n - k + 1)}{k \cdots 1}.
+    \begin{pmatrix} n \\ k \end{pmatrix} = \frac{n! \cdot (n-k)!}{k!} = \frac{n \cdot (n-1) \cdots (n - k + 1)}{k \cdots 1}.
 @f]
 */
 template<class T> inline typename std::enable_if<IsIntegral<T>::value, T>::type binomialCoefficient(T n, T k) {
--- a/src/Magnum/Math/Functions.h
+++ b/src/Magnum/Math/Functions.h
@@ -34,6 +34,7 @@
 #include <cstdlib> /* std::div() */
 #include <type_traits>
 #include <utility>
+#include <limits>
 #include <Corrade/Utility/StlMath.h>
 
 #include "Magnum/visibility.h"
@@ -408,6 +409,33 @@ template<std::size_t size, class T> inline Vector<size, T> ceil(const Vector<siz
     for(std::size_t i = 0; i != size; ++i)
         out[i] = Math::ceil(a[i]);
     return out;
+}
+
+
+/**
+@brief Computes a Binomial Coefficient
+
+Given n elements computes the number of ways to choose k elements
+without replacement from n elements. The number of ways to do this is
+given by
+@f[
+    \frac{n}{k} = \frac{n! \cdot (n-k)!}{k!} = \frac{n \cdot (n-1) \cdots (n - k + 1)}{k \cdots 1}.
+@f]
+*/
+template<class T> inline typename std::enable_if<IsIntegral<T>::value, T>::type binomialCoefficient(T n, T k) {
+    CORRADE_ASSERT(k >= 0, "k must be positive to compute the Binomial Coefficient", T{0});
+    if (k > n) return 0;
+    if (k * 2 > n)
+        k = n-k;
+    if (k == 0) return 1;
+
+    T result = n;
+    for(T i = 2; i <= k; ++i ) {
+        CORRADE_ASSERT(result < std::numeric_limits<T>::max()/(n-i+1), "Overflow In Binomial Coefficient", T{0});
+        result *= (n-i+1);
+        result /= i;
+    }
+    return result;
 }
 
 /**
--- a/src/Magnum/Math/Functions.h
+++ b/src/Magnum/Math/Functions.h
@@ -34,7 +34,6 @@
 #include <cstdlib> /* std::div() */
 #include <type_traits>
 #include <utility>
-#include <limits>
 #include <Corrade/Utility/StlMath.h>
 
 #include "Magnum/visibility.h"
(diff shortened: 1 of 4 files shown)


src/Magnum/Math/Functions.h
```diff
@@ -410,6 +411,33 @@ template<std::size_t size, class T> inline Vector<size, T> ceil(const Vector<siz
...
+
+Given n elements computes the number of ways to choose k elements
+without replacement. The number of ways to do this is
+given by
+@f[
+    \begin{pmatrix} n \\ k \end{pmatrix} = \frac{n! \cdot (n-k)!}{k!} = \frac{n \cdot (n-1) \cdots (n - k + 1)}{k \cdots 1}.
+@f]
+*/
+template<class T> inline typename std::enable_if<IsIntegral<T>::value, T>::type binomialCoefficient(T n, T k) {
+    CORRADE_ASSERT(k >= 0, "k must be positive to compute the Binomial Coefficient", T{0});
```
mosra, line 426: Hmm, what about accepting just `UnsignedInt` here (i.e., making it non-templated)? That would also remove a need for this assert. 

When I was googling for nchoosek, i found [this in matlab docs](http://matlab.izmiran.ru/help/techdoc/ref/nchoosek.html):

> This function is only practical for situations where n is less than about 15.

So I assume there isn't much point in accepting 64-bit integers because the results wouldn't fit into the type anyway?

Oh, and with a non-templated version the definition could then probably go into `Functions.cpp` (don't forget to add `MAGNUM_EXPORT` to the declaration then).

src/Magnum/Math/Functions.h
```diff
@@ -410,6 +411,33 @@ template<std::size_t size, class T> inline Vector<size, T> ceil(const Vector<siz
     return out;
 }
 
+
+/**
+@brief Computes a Binomial Coefficient
```
mosra, line 416:
```suggestion
@brief [Binomial Coefficient](https://en.wikipedia.org/wiki/Binomial_coefficient)
```

for people like me, who would need to google this anyway :laughing:

src/Magnum/Math/Functions.h
```diff
@@ -34,6 +34,7 @@
 #include <cstdlib> /* std::div() */
 #include <type_traits>
 #include <utility>
+#include <limits>
```
mosra, line 37: This adds about 1k preprocessed lines which isn't too much but also isn't too little. With the function not being templated anymore, you could use `~UnsignedInt{}` instead for the second assert.

src/Magnum/Math/Functions.h
```diff
@@ -410,6 +411,33 @@ template<std::size_t size, class T> inline Vector<size, T> ceil(const Vector<siz
...
+template<class T> inline typename std::enable_if<IsIntegral<T>::value, T>::type binomialCoefficient(T n, T k) {
+    CORRADE_ASSERT(k >= 0, "k must be positive to compute the Binomial Coefficient", T{0});
+    if (k > n) return 0;
+    if (k * 2 > n)
+        k = n-k;
+    if (k == 0) return 1;
+
+    T result = n;
+    for(T i = 2; i <= k; ++i ) {
+        CORRADE_ASSERT(result < std::numeric_limits<T>::max()/(n-i+1), "Overflow In Binomial Coefficient", T{0});
```
mosra, line 434: I'm always prefixing asserts with the API name to make it clearer where it happened even without a backtrace, and it could probably make sense to print the inputs also:

```cpp
"Math::binomialCoefficient(): overflow for (" << Debug::nospace << n << k << Debug::nospace << ")"
```

src/Magnum/Math/Test/FunctionsTest.cpp
```diff
@@ -293,6 +297,15 @@ void FunctionsTest::ceil() {
...
 
+void FunctionsTest::binomialCoefficient() {
+    CORRADE_COMPARE(Math::binomialCoefficient(1, 1), 1);
+    CORRADE_COMPARE(Math::binomialCoefficient(1, 0), 1);
+    CORRADE_COMPARE(Math::binomialCoefficient(19, 11), 75582);
+    CORRADE_COMPARE(Math::binomialCoefficient(1000, 999), 1000);
+    CORRADE_COMPARE(Math::binomialCoefficient(0, 0), 1);
+    CORRADE_COMPARE(Math::binomialCoefficient(32, 11), 129024480);
+}
+
```
mosra, line 308: Thanks for the test!

I'm usually also testing assertion messages, to be sure that they make sense. An [example how to do that is here](https://github.com/mosra/magnum/blob/80a844c6c5256b1c77a04db22f8042891b521c0b/src/Magnum/Math/Test/VectorTest.cpp#L556-L568) -- note it has to be a separate test due to the skip (`binomialCoefficientOverflow()` e.g.).

src/Magnum/Math/Functions.h
```diff
@@ -410,6 +411,33 @@ template<std::size_t size, class T> inline Vector<size, T> ceil(const Vector<siz
     return out;
 }
 
+
+/**
+@brief Computes a Binomial Coefficient
+
+Given n elements computes the number of ways to choose k elements
+without replacement. The number of ways to do this is
+given by
```
mosra, line 420: Stealing further from the same [matlab docs](http://matlab.izmiran.ru/help/techdoc/ref/nchoosek.html), I think this would be the most concise explanation:

```
Returns the number of combinations of @f$ n @f$ things taken @f$ k @f$ at a time.
```

src/Magnum/Math/Functions.h
Janos95: Hm I wonder how they arrived at 15. If we compute n choose k the biggest temporary result will 
be (n choose n/2) *  n/2. This overflows for
32 bit unsigned integers for n >= 31
64 bit unsigned integers for n >= 63
128 bit unsigned integer for n >= 126
So I do think there is some value in having this as a template. But since I only need n <= 30 I can untemplate it if you still think this is the better way :)

mosra: Okay, what about `UnsignedInt` inputs and `UnsignedLong` output (and temporaries)? :)

I was thinking about something like "factorials REALLY HARD to compute for n > 15" but that doesn't really seem to make sense either. Maybe in 1994 it was and that's why the 15?

Janos95: ok :).
If they really compute the factorial that would already get problematic for n = 13.
Another possibility is that they had 16 bit unsigned integers that actually gets you to [15](https://www.wolframalpha.com/input/?i=%2815+choose+15%2F2%29+*++15%2F2+%3E+2%5E16+-+1+) if am not mistaken.

mosra: Yep, that sounds likely, maybe the documentation is really from the time where 32bit calculations were expensive and just hasn't been updated since.

mosra:
Hmm, one remaining question -- is the `k > n` case actually valid? I'd kinda expect this to be asserting as well. I can add that but want your confirmation first :)

mosra:
Merged in 8013afc28cd2f372bada51bd4f1d5f57209834e0, thank you!

PR #470, opened 2020-09-21
Fix: can not create gl context with GLFW3 + ANGLE on Windows

linuxaged:
EGL Context can not be created correctly without `glfwWindowHint(GLFW_CONTEXT_CREATION_API, GLFW_EGL_CONTEXT_API);`:
```
GL::Context: unsupported OpenGL ES version (0, 0)
```

Have tested on Win64 and passed.

### DIFF
--- a/src/Magnum/Platform/GlfwApplication.cpp
+++ b/src/Magnum/Platform/GlfwApplication.cpp
@@ -451,6 +451,7 @@ bool GlfwApplication::tryCreate(const Configuration& configuration, const GLConf
         }
         #else
         glfwWindowHint(GLFW_CLIENT_API, GLFW_OPENGL_ES_API);
+        glfwWindowHint(GLFW_CONTEXT_CREATION_API, GLFW_EGL_CONTEXT_API);
         #endif
 
     /* Request usable version otherwise */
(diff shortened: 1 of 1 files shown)


mosra:
Thanks a lot, just when I was about to do the same here :)

There's one special case for "desktop ES" where it's GLES + WGL/GLX instead of EGL, I'll check locally how that behaves and then merge.

mosra:
Did a bunch of testing on Linux, and while I can `LD_PRELOAD` ANGLE to test with windowless apps, it doesn't let me do that for SDL/GLFW apps, failing on `GL::Context: unsupported OpenGL ES version (0, 0)` in both cases. So I can't really verify this doesn't break anything. Without ANGLE preloaded, everything works and this option doesn't seem to change anything (which is strange, I would at least expect that I can't load entrypoints via EGL on a context created using GLX, but even that works :thinking:). Will need to investigate further.

Reading GLFW docs, setting `GLFW_CONTEXT_CREATION_API` to EGL on macOS would fail, I'll need to check what to do there in order to support ANGLE (if you have any experience on that side, please share :)).

---

For a conservative change, before I have a chance to verify all other cases, I'd suggest to wrap those in the following, that'll make it work for ANGLE on Windows but wouldn't break for desktop ES (where EGL isn't used) and wouldn't do anything on other systems, since on Linux this option doesn't seem to do anything and on macOS it'd break things.

```cpp
#if defined(CORRADE_TARGET_WINDOWS) && !defined(MAGNUM_TARGET_DESKTOP_GLES)
```

linuxaged:
Does Firefox/Chrome Linux version run on top of ANGLE? OpenGL seems considered as native on Linux.
i'll make it work for ANGLE on Windows only.

mosra:
On Linux ANGLE can act as a GL-over-Vulkan layer, the same on Mac.

Merged as 6a345ea9827aa83e99abd1c6e046d513ce93eb78, thank you! :+1:

PR #473, opened 2020-10-17
add Cool-Warm colormap

Janos95:
For a lot of scientific visualizations the [Cool-Warm color map](https://www.kennethmoreland.com/color-maps/) is very popular.
For instance it is the default color map in paraview.
Example using the cool-warm colormap (I used the doc image as a texture) for visualizing the curvature on a torus:
curvature


src/Magnum/DebugTools/ColorMap.h
```diff
@@ -204,6 +204,20 @@ MAGNUM_DEBUGTOOLS_EXPORT Containers::StaticArrayView<256, const Vector3ub> infer
...
+@brief Cool-Warm colormap
+@m_since{2020,06}
+
+@image html colormap-cool_warm.png width=100%
+
+@m_class{m-block m-primary}
+
+@thirdparty Created by [Kenneth Moreland](https://www.kennethmoreland.com/color-advice/).
+    It is a diverging (double-ended) color map with a smooth transition in the middle to prevent artifacts at the midpoint.
+    Although not isoluminant, this color map avoids dark colors to allow shading cues throughout.
```
mosra, line 217: Hmm, I wanted to point out that I don't like how it doesn't have uniform brightness, but this comment mentions that, so all good :)

I tried to find some licensing info, but didn't find any, so I guess I'll assume Public Domain, then.

doc/generated/colormaps.cpp
```diff
@@ -48,7 +48,8 @@ int main() {
         std::make_pair(DebugTools::ColorMap::magma(), "colormap-magma.png"),
         std::make_pair(DebugTools::ColorMap::plasma(), "colormap-plasma.png"),
         std::make_pair(DebugTools::ColorMap::inferno(), "colormap-inferno.png"),
-        std::make_pair(DebugTools::ColorMap::viridis(), "colormap-viridis.png")
+        std::make_pair(DebugTools::ColorMap::viridis(), "colormap-viridis.png"),
+        std::make_pair(DebugTools::ColorMap::coolWarm(), "colormap-cool_warm.png")
```
mosra, line 52: You're awesome, I didn't even expect you to make it *this* complete :)

src/Magnum/DebugTools/ColorMap.cpp
```diff
@@ -373,11 +373,80 @@ constexpr UnsignedByte Viridis[][3] = {
     {246, 230,  31}, {248, 230,  33}, {250, 230,  34}, {253, 231,  36}
 };
 
+/* Taken from https://www.kennethmoreland.com/color-advice/ */
```
mosra, line 376: Did you do this by hand? :) Like, even if you have a quick'n'dirty Python or whatever script, that'd be useful to preserve in case we'd like to add more from there, or the float versions and such.

In case you don't, there's no need to create it now, though.

src/Magnum/DebugTools/ColorMap.h
```diff
@@ -204,6 +204,20 @@ MAGNUM_DEBUGTOOLS_EXPORT Containers::StaticArrayView<256, const Vector3ub> infer
 */
 MAGNUM_DEBUGTOOLS_EXPORT Containers::StaticArrayView<256, const Vector3ub> viridis();
 
+/**
+@brief Cool-Warm colormap
```
mosra, line 208: This is the "Smooth" version, right? I'm wondering if it would make sense to add the other one ... or maybe just rename this to Smooth Cool-Warm to avoid confusion and reserve a potential place for the Bent one in the future.

mosra (commented):
Thanks, this is great :)

src/Magnum/DebugTools/ColorMap.cpp
Janos95: Sorry for the long delay. I used a python script, but its extremely hacky and just prints a pretty version of the csv file the website provides. Here it is in case you want to judge yourself : 

```

from numpy import genfromtxt
data = genfromtxt('cool_warm.csv', delimiter=',')
formatted = ''
data = data[1:,1:].astype(int)
formatted = ''
for i in range(256//4):
    line = ''
    for j in range(4):
        row = data[4*i+j]
        line += '{{{:>3}, {:>3}, {:>3}}}, '.format(row[0], row[1], row[2])
    formatted = formatted + line + '\n'

print(formatted)
```

src/Magnum/DebugTools/ColorMap.h
Janos95: Yes its the smooth version. I'll rename it. I feel like the bent version looks looks pretty similar to the smooth version. So until somebody actually wants to use it, I would suggest to leave it out?

src/Magnum/DebugTools/ColorMap.cpp
mosra: That's totally fine, actually ;) The main point of these is not to win code beauty competitions, but to *exist* so the next person doesn't have to write another one-off hacky script. Thank you.

src/Magnum/DebugTools/ColorMap.h
mosra: I can see a use case where you compare two color-mapped images and the small difference between the smooth and bent variant would significantly alter the result. So better be clear :)

Janos95: Agree. I added the bent version as well :)

mosra:
Merged in 81feec8f2d9ca569239fb6622f412ada0d836096, thank you!

PR #481, opened 2020-11-06
CSS changes to allow multiple Emscripten canvases on one page

mosra:
Prerequisite / related to #480.

- changes the CSS to use classes instead of IDs for styling, allowing to style more than one canvas on the same page without having to patch / copy `WebApplication.css`
- changes `setContainerCssClass()` to not use a hardcoded `#container` element but instead go up from the `<canvas>` to the first `.mn-container` (which, again, should make more canvases work)
- changes `Module` to not use hardcoded `#status` and `#status-description` IDs but instead fetch them from `Module.status` and `Module.statusDescription` that are easier to be overriden from the HTML

~~What's still not great:~~ All handled by #480 now:

- [x] The `EmscriptenApplication.js` driver *implicitly runs code* in order to populate application `argc`/`argv`. This should be deferred somehow because otherwise it's not really possible to have this code use externally-overriden `Module.status`
- [x] Can some changes be done to make it possible to reuse `EmscriptenApplication.js` to populate `SomeOtherModule` (instead of `Module`)? Currently the user has to copypaste the JS contents and modify them, which then makes upgrades harder
- [x] Docs on how to override `Module.canvas`, `status` etc. still need to be written

Cc: @pezcode

### DIFF
--- a/doc/platforms-html5.dox
+++ b/doc/platforms-html5.dox
@@ -129,11 +129,11 @@ variables.
 </head>
 <body>
   <h1>Magnum Emscripten Application</h1>
-  <div id="container">
-    <div id="sizer"><div id="expander"><div id="listener">
-      <canvas id="canvas"></canvas>
-      <div id="status">Initialization...</div>
-      <div id="status-description"></div>
+  <div class="mn-container">
+    <div class="mn-sizer"><div class="mn-expander"><div class="mn-listener">
+      <canvas class="mn-canvas" id="canvas"></canvas>
+      <div class="mn-status" id="status">Initialization...</div>
+      <div class="mn-status-description" id="status-description"></div>
       <script src="EmscriptenApplication.js"></script>
       <script async="async" src="{{ application }}.js"></script>
     </div></div></div>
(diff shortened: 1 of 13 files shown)


pezcode:
> The EmscriptenApplication.js driver implicitly runs code in order to populate application argc/argv. This should be deferred somehow because otherwise it's not really possible to have this code use externally-overriden Module.status

Can you elaborate on that? I don't see `Module.status` being used in relation to `Module.arguments`. The arguments can be overridden by the user just like the canvas etc., if that's what you're referring to.

> Can some changes be done to make it possible to reuse `EmscriptenApplication.js` to populate `SomeOtherModule` (instead of `Module`)? Currently the user has to copypaste the JS contents and modify them, which then makes upgrades harder

This is doable with `-s MODULARIZE` and `--pre-js`. PlatformMultipleEmscriptenApplicationTest is doing exactly that to override the HTML elements in `Module`. The alternative is to rewrite the .js as a module that exports the `Module` object but I have no idea how this interacts with Emscripten and browser compatibility...

Issue #453, opened 2020-07-03
2020.0b release


mosra:
Opening early mainly to avoid forgetting the tiny but important things to do.

- [ ] Prefix all CMake options with `CORRADE_` / `MAGNUM_` to make things less shitty for subprojects
  - and provide some backwards compatibility to avoid breaking *everything*

ttnghia:
More examples please 😋

mosra:
@ttnghia yeah I know, sorry that I didn't merge your latest one yet :see_no_evil:

ttnghia:
No problem  😄 Once I'm done with my job search (in another month or so), I may have more examples for you 😏

mgood7123:
for

 ⚠️ Assert redirection to default output makes the messages invisible on Android apps (basically need a variant of https://gist.github.com/mosra/566aedd25eadd07927d4f7e385aac931, but for Android apps only)

why dont you just redirect the output at compile time like i do?

HINT:
to add support for alternative output redirection on other platform's such as IOS, simply `#elif defined` the `DEFAULT_STREAM_*` definitions

for example, this redirects to android's `logcat` when compiling for `android`, and `std::cout` and `std::cerr` when compiling for everything else (i am not sure if IOS uses `std::cout` for printing or not)

```C++
#if defined(CORRADE_TARGET_ANDROID)
#include <Corrade/Utility/AndroidLogStreamBuffer.h>

const char * tag = "magnum";

Corrade::Utility::AndroidLogStreamBuffer bufferDebug {Corrade::Utility::AndroidLogStreamBuffer::LogPriority::Debug, tag};
std::ostream androidLogStreamDebug {&bufferDebug};

Corrade::Utility::AndroidLogStreamBuffer bufferWarning {Corrade::Utility::AndroidLogStreamBuffer::LogPriority::Warning, tag};
std::ostream androidLogStreamWarning {&bufferWarning};

Corrade::Utility::AndroidLogStreamBuffer bufferError {Corrade::Utility::AndroidLogStreamBuffer::LogPriority::Error, tag};
std::ostream androidLogStreamError {&bufferError};

#define DEFAULT_STREAM_D &androidLogStreamDebug
#define DEFAULT_STREAM_W &androidLogStreamWarning
#define DEFAULT_STREAM_E &androidLogStreamError
#else
#define DEFAULT_STREAM_D &std::cout
#define DEFAULT_STREAM_W &std::cerr
[...]

PR #482, opened 2020-11-17
Math: Fix Math::Intersection::sphereFrustum

Squareys:
Hi @mosra !

As per #453 (Fix sphere frustum intersection), here's the fix plus extensive tests.

I made a conscious descision to mark "touching" geometries as non-intersection. Even though mathematically, the volume is not empty, it's infinitely small and hence useless for the main use case, which will be frustum culling.

Best,
Jonathan

### DIFF
--- a/src/Magnum/Math/Test/IntersectionTest.cpp
+++ b/src/Magnum/Math/Test/IntersectionTest.cpp
@@ -181,12 +181,12 @@ void IntersectionTest::rangeFrustum() {
 
 void IntersectionTest::aabbFrustum() {
     const Frustum frustum{
-        {1.0f, 0.0f, 0.0f, 0.0f},
         {-1.0f, 0.0f, 0.0f, 5.0f},
-        {0.0f, 1.0f, 0.0f, 0.0f},
+        {1.0f, 0.0f, 0.0f, 0.0f},
         {0.0f, -1.0f, 0.0f, 1.0f},
-        {0.0f, 0.0f, 1.0f, 0.0f},
-        {0.0f, 0.0f, -1.0f, 10.0f}};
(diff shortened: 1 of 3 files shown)


src/Magnum/Math/Test/IntersectionTest.cpp
```diff
@@ -204,20 +204,50 @@ void IntersectionTest::aabbFrustum() {
...
+    /* Frustum spanning [(-10, -10, -10), (0, 0, 0)] */
     const Frustum frustum{
+        {-1.0f, 0.0f, 0.0f, -10.0f},
         {1.0f, 0.0f, 0.0f, 0.0f},
-        {-1.0f, 0.0f, 0.0f, 10.0f},
+        {0.0f, -0.5f, 0.0f, -5.0f}, /* Intentionally not normalized */
         {0.0f, 1.0f, 0.0f, 0.0f},
-        {0.0f, -1.0f, 0.0f, 10.0f},
         {0.0f, 0.0f, 1.0f, 0.0f},
-        {0.0f, 0.0f, -1.0f, 10.0f}};
```
mosra, line 213: Hmm, so if there's multiple ways how to represent the same plane and the planes can be in any order, maybe the test could be made instanced with testing a few equivalent variants and expecting the same points get always the same result? That could harden it quite nicely.

And that could be done for all `*Frustum` tests -- I see you already updated `planeFrustum()`, so the same instancing change could be done to all.

Squareys: Instanced test for permutating the plane order sounds like it would get out of hand in the test results pretty quickly, I think a loop with iteration description is probably enough?

Re the transformation: Do you mean offsetting/transforming the entire space, frustum and points, sphere's etc by a certain transform?

PR #483, opened 2020-11-18
EmscriptenApplication: use _malloc instead of internal allocate

pezcode:
Fix for `allocate` having a [different function signature starting with Emscripten 2.0.5](https://github.com/emscripten-core/emscripten/commit/fc67fbf94b6b815c619e75dfe59c84f3c81bf698).

I think this is cleaner than `#ifdef`-ing around `__EMSCRIPTEN_major/minor/tiny__` as that internal function could change again in the future, and duplicating `EM_ASM` blocks means even more code and maintenance.

I added `ASSERTIONS` and `SAFE_HEAP` to the test application compile flags so similar issues get caught in the future.

cc: @Squareys

### DIFF
--- a/src/Magnum/Platform/EmscriptenApplication.cpp
+++ b/src/Magnum/Platform/EmscriptenApplication.cpp
@@ -186,7 +186,11 @@ namespace {
         #pragma GCC diagnostic push
         #pragma GCC diagnostic ignored "-Wgnu-zero-variadic-macro-arguments"
         char* id = reinterpret_cast<char*>(EM_ASM_INT({
-            return allocate(intArrayFromString(Module['canvas'].id), 'i8', ALLOC_NORMAL);
+            const id = Module['canvas'].id;
+            const bytes = lengthBytesUTF8(id) + 1;
+            const memory = _malloc(bytes);
+            stringToUTF8(id, memory, bytes);
+            return memory;
         }));
         #pragma GCC diagnostic pop
         std::string str = id;
@@ -562,8 +566,12 @@ void EmscriptenApplication::setupCallbacks(bool resizable) {
 
         if(element === document) return 1; /* EMSCRIPTEN_EVENT_TARGET_DOCUMENT */
         if(element === window) return 2; /* EMSCRIPTEN_EVENT_TARGET_WINDOW */
-        if('id' in element)
-            return allocate(intArrayFromString(element.id), 'i8', ALLOC_NORMAL);
+        if('id' in element) {
+            const bytes = lengthBytesUTF8(element.id) + 1;
+            const memory = _malloc(bytes);
+            stringToUTF8(element.id, memory, bytes);
+            return memory;
+        }
 
         return 0;
     }));
--- a/src/Magnum/Platform/Test/CMakeLists.txt
+++ b/src/Magnum/Platform/Test/CMakeLists.txt
@@ -47,7 +47,9 @@ if(WITH_EMSCRIPTENAPPLICATION)
         # Test that the canvas and keylistener can be found with both the old
         # and the new Emscripten target behaviors
         # TODO: use target_link_options() once we can require CMake 3.13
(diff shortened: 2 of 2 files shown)


src/Magnum/Platform/Test/CMakeLists.txt
```diff
@@ -47,7 +47,9 @@ if(WITH_EMSCRIPTENAPPLICATION)
         # Test that the canvas and keylistener can be found with both the old
         # and the new Emscripten target behaviors
         # TODO: use target_link_options() once we can require CMake 3.13
-        "-s DISABLE_DEPRECATED_FIND_EVENT_TARGET_BEHAVIOR=0")
+        "-s DISABLE_DEPRECATED_FIND_EVENT_TARGET_BEHAVIOR=0"
+        # Enable memory runtime checks
+        "-s ASSERTIONS=2 -s SAFE_HEAP=1")
```
mosra, line 52: Are there any downsides of enabling assertions? I suppose it will increase the resulting JS file size quite a bit -- I occasionally use these tests to check changes in build sizes, so I guess I'll have to remember to temporarily comment this out when doing the comparisons.

I was thinking about enabling this only for Debug builds, but I personally almost never build for Emscripten in Debug, so that wouldn't help in catching new breakages.

src/Magnum/Platform/EmscriptenApplication.cpp
```diff
@@ -562,8 +566,12 @@ void EmscriptenApplication::setupCallbacks(bool resizable) {
...
         if(element === document) return 1; /* EMSCRIPTEN_EVENT_TARGET_DOCUMENT */
         if(element === window) return 2; /* EMSCRIPTEN_EVENT_TARGET_WINDOW */
-        if('id' in element)
-            return allocate(intArrayFromString(element.id), 'i8', ALLOC_NORMAL);
+        if('id' in element) {
+            const bytes = lengthBytesUTF8(element.id) + 1;
+            const memory = _malloc(bytes);
+            stringToUTF8(element.id, memory, bytes);
+            return memory;
+        }
```
mosra, line 574: Did you see any significant change in build output size with this change present? I hope malloc doesn't drag along tens of kilobytes of JS code that wouldn't be needed with plain allocate :)

mosra (commented):
This is great, thanks!

There's ~~one more~~ actually two more occurences of `allocate()` [in Corrade's Utility::Arguments](https://github.com/mosra/corrade/blob/92daf1a09985c45630613f2720162c16340022ba/src/Corrade/Utility/Arguments.cpp#L126) (and [here](https://github.com/mosra/corrade/blob/92daf1a09985c45630613f2720162c16340022ba/src/Corrade/Utility/Arguments.cpp#L514)), could I ask you to update that as well? Thanks :pray: To test, it should be enough to build the `UtilityArgumentTest`  (with `BUILD_TESTS` enabled in CMake) and run `node ./Release/bin/UtilityArgumentsTest.js` (possibly with `-s ASSERTIONS=1` enabled).

src/Magnum/Platform/Test/CMakeLists.txt
pezcode: You could enable it for the multiple canvas test app only, so it's tested but you still have the normal application to test with?

mosra: Eh, let's keep it there :) If it gets in the way of something, I'll remove it, but for now better safe than sorry.

src/Magnum/Platform/EmscriptenApplication.cpp
pezcode: 136kb before, 137kb after, 191kb with `ASSERTIONS` and `SAFE_HEAP`

mosra: Nice. That's quite okay.

mosra:
I *finally* managed to get at least some CI jobs running and merged as f94035805de34fd1822c943706ed6fb6b9dce5ef. Thank you!

PR #491, opened 2021-01-05
Add a workaround for eglQueryDeviceAttribEXT segfaulting

erikwijmans:
On nv driver 450.80.02, `eglQueryDeviceAttribEXT` segfaults when query GPUs that the user does not have access to (i.e. via cgroup).  Instead, always call `eglQueryDeviceStringEXT` as that doesn't segfault and sets an error that can be retrieved via eglGetError to see if the user has access to that device. On well behaved driver versions, `eglQueryDeviceAttribEXT` returns false instead of segfaulting.

There is of course the chance that `eglQueryDeviceStringEXT` will be the call with a problem in the future :)

### DIFF
--- a/src/Magnum/GL/Implementation/driverSpecific.cpp
+++ b/src/Magnum/GL/Implementation/driverSpecific.cpp
@@ -161,6 +161,13 @@ const char* KnownWorkarounds[]{
    possible to get driver version through EGL, so enabling this unconditionally
    on all EGL NV contexts. */
 "nv-egl-incorrect-gl11-function-pointers",
+
+/* On nv driver 450.80.02, eglQueryDeviceAttribEXT segfaults when query GPUs that 
+   the user does not have access to (i.e. via cgroup).  Instead, always call 
+   eglQueryDeviceStringEXT as that doesn't segfault and sets an error that can be 
+   retrieved via eglGetError to see if the user has access to that device. On well 
+   behaved driver versions, eglQueryDeviceAttribEXT returns false instead of segfaulting. */
+"nv-egl-crashy-query-device-attrib",
 #endif
 
 #ifndef MAGNUM_TARGET_GLES
--- a/src/Magnum/Platform/WindowlessEglApplication.cpp
+++ b/src/Magnum/Platform/WindowlessEglApplication.cpp
@@ -180,8 +180,17 @@ WindowlessEglContext::WindowlessEglContext(const Configuration& configuration, G
                 /* Go through the EGL devices and find one that has the desired
                    CUDA device number */
                 for(selectedDevice = 0; selectedDevice < UnsignedInt(count); ++selectedDevice) {
+                    /* On nv driver 450.80.02, eglQueryDeviceAttribEXT segfaults when query GPUs that 
+                       the user does not have access to (i.e. via cgroup).  Instead, always call 
+                       eglQueryDeviceStringEXT as that doesn't segfault and sets an error that can be 
+                       retrieved via eglGetError to see if the user has access to that device. On well 
+                       behaved driver versions, eglQueryDeviceAttribEXT returns false instead of segfaulting. */
+                    const char* const eglExtensions = eglQueryDeviceStringEXT(devices[selectedDevice], EGL_EXTENSIONS);
+                    if(eglGetError() == EGL_BAD_DEVICE_EXT && !magnumContext->isDriverWorkaroundDisabled("nv-egl-crashy-query-device-attrib"))
+                        continue;
+
                     if(magnumContext && (magnumContext->internalFlags() >= GL::Context::InternalFlag::DisplayVerboseInitializationLog))
-                        Debug{} << "Platform::WindowlessEglApplication: eglQueryDeviceStringEXT(EGLDevice=" << Debug::nospace << selectedDevice << Debug::nospace << "):" << eglQueryDeviceStringEXT(devices[selectedDevice], EGL_EXTENSIONS);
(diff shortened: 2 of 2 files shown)


src/Magnum/Platform/WindowlessEglApplication.cpp
```diff
@@ -180,8 +180,19 @@ WindowlessEglContext::WindowlessEglContext(const Configuration& configuration, G
...
+                       behaved driver versions, eglQueryDeviceAttribEXT returns false instead of segfaulting. */
+                    if (!magnumContext || (magnumContext->internalFlags() >= GL::Context::InternalFlag::DisplayVerboseInitializationLog) || !magnumContext->isDriverWorkaroundDisabled("nv-egl-query-device-attrib"))
+                        queryResult = eglQueryDeviceStringEXT(devices[selectedDevice], EGL_EXTENSIONS);
                     if(magnumContext && (magnumContext->internalFlags() >= GL::Context::InternalFlag::DisplayVerboseInitializationLog))
-                        Debug{} << "Platform::WindowlessEglApplication: eglQueryDeviceStringEXT(EGLDevice=" << Debug::nospace << selectedDevice << Debug::nospace << "):" << eglQueryDeviceStringEXT(devices[selectedDevice], EGL_EXTENSIONS);
+                        Debug{} << "Platform::WindowlessEglApplication: eglQueryDeviceStringEXT(EGLDevice=" << Debug::nospace << selectedDevice << Debug::nospace << "):" << queryResult;
+
+                    if ((!magnumContext || !magnumContext->isDriverWorkaroundDisabled("nv-egl-query-device-attrib")) && (eglGetError() == EGL_BAD_DEVICE_EXT))
+                        continue;
```
mosra, line 195: +1 for going the extra steps to make this a disable-able driver workaround, thank you.

There's one internal undocumented peculiarity with `isDriverWorkaroundDisabled()` -- when you call it, it will add the workaround to the list in engine startup log. Which means, as far as I can see, the above code will cause the workaround to be listed even on well-behaving driver versions, which could result in confusion. The code thus has to be structured in a way that `isDriverWorkaroundDisabled()` gets called only when you know *for sure* you hit the problematic code path.

Assuming driver sanity with `eglQueryDeviceStringEXT()` returning a pointer to static memory with minimal overhead, I think this could work:

```cpp
/* On nv driver 450.80.02, eglQueryDeviceAttribEXT segfaults when query GPUs that 
   the user does not have access to (i.e. via cgroup).  Instead, always call 
   eglQueryDeviceStringEXT as that doesn't segfault and sets an error that can be 
   retrieved via eglGetError to see if the user has access to that device. On well 
   behaved driver versions, eglQueryDeviceAttribEXT returns false instead of segfaulting. */
const char* const eglExtensions = eglQueryDeviceStringEXT(devices[selectedDevice], EGL_EXTENSIONS);
if(eglGetError() == EGL_BAD_DEVICE_EXT && !magnumContext->isDriverWorkaroundDisabled("nv-egl-query-device-attrib"))
    continue;

if(magnumContext && (magnumContext->internalFlags() >= GL::Context::InternalFlag::DisplayVerboseInitializationLog))
    Debug{} << "Platform::WindowlessEglApplication: eglQueryDeviceStringEXT(EGLDevice=" << Debug::nospace << selectedDevice << Debug::nospace << "):" << eglExtensions;
```

I also changed the verbose print to happen *after* the workaround because I don't think we should assume the return value of `eglQueryDeviceString()` is sane when it causes `EGL_BAD_DEVICE_EXT`.

Are you able to test that both cases (when the workaround is needed and when is not, for example when using all GPUs of a cluster) work? And that running with `--magnum-disable-workarounds nv-egl-query-device-attrib` still causes the expected crash :)

src/Magnum/GL/Implementation/driverSpecific.cpp
```diff
@@ -161,6 +161,13 @@ const char* KnownWorkarounds[]{
    possible to get driver version through EGL, so enabling this unconditionally
    on all EGL NV contexts. */
 "nv-egl-incorrect-gl11-function-pointers",
+
+/* On nv driver 450.80.02, eglQueryDeviceAttribEXT segfaults when query GPUs that 
+   the user does not have access to (i.e. via cgroup).  Instead, always call 
+   eglQueryDeviceStringEXT as that doesn't segfault and sets an error that can be 
+   retrieved via eglGetError to see if the user has access to that device. On well 
+   behaved driver versions, eglQueryDeviceAttribEXT returns false instead of segfaulting. */
+"nv-egl-query-device-attrib",
```
mosra, line 170: Could this be 

```suggestion
"nv-egl-crashy-query-device-attrib",
```

instead? :) Just so it's clearer what exactly is the workaround for.

src/Magnum/Platform/WindowlessEglApplication.cpp
erikwijmans: Makes sense.  The reason I set it up like this was to avoid calling `eglQueryDeviceString()` at all when the work around is disabled, but always doing that query should be fine.

Oddly enough, the return value of `eglQueryDeviceString()` is perfectly sane when it causes `EGL_BAD_DEVICE_EXT` and is exactly the same as when it doesn't.

erikwijmans: Confirmed that with the requested changes, `nv-egl-crashy-query-device-attrib` only shows up when the workaround is needed and things segfault when `nv-egl-crashy-query-device-attrib` is disabled.

mosra: Awesome, thanks a lot! :+1:

mosra:
Merged as 015985016c39c8e5917b4c677929d55effbc747b, thank you! :rocket:

PR #494, opened 2021-02-10
GL::Context: pass InternalFlags as parameter

### LINKED ISSUE #493: GL::defaultFramebuffer is neither multi-context nor multi-thread safe
Hey @mosra, I'm working on a VR application with low-latency JPEG streaming and I'm trying to reduce latency even further by using a background thread to upload the data. Not sure yet if I can gain something using that setup, but I noticed a problem along the way:

The `GL::defaultFramebuffer` instance is a global variable and it gets written to (at least) in https://github.com/mosra/magnum/blob/41b59d81760f8b8e601491193887d91ba05d76ab/src/Magnum/GL/DefaultFramebuffer.cpp#L116-L124, which is called during GL::Context creation.

So setting up a second GL::Context will mess up the main context's defaultFramebuffer. In a multi-threading scenario, it might even end up with some completely bogus viewport range.

Since the docs seem to suggest that multi-threading with multiple contexts is supported (e.g. https://doc.magnum.graphics/magnum/platform.html#platform-windowless-contexts) I guess we should fix that somehow.

Some thoughts on a fix:
* Will be rather difficult since `defaultFramebuffer` cannot be made thread-local and exported on windows. And even if we did that, we would need to overwrite it on `GL::Context::makeCurrent()`.
* We could make the default framebuffer a property of the context state (currentContext is already thread-local), and change `defaultFramebuffer` to a function returning `currentContext->state_->defaultFramebuffer` or similar). That would be a larger API change - is it worth it?
* Maybe write a wrapper class that forwards all method calls to the correct framebuffer instance? Involves duplicating all methods...

I might have time to work on this later this week, but I guess I'll wait for your input on this ;)

xqms:
Here's a possible way to resolve #493 (GL::defaultFramebuffer gets clobbered when you create a second context).

I'm using the already existent InternalFlags enum set, adding a new `NoFramebuffer` flag, which skips initialization of `defaultFramebuffer`. Additionally, the flags are exposed as additional parameters of the `GL::Context` and `Platform::GLContext` constructors, allowing users to pass in the new flag.

Coincidentally, this also allows to control the other flags without creating dummy argc/argv for `Platform::GLContext`, which feels much cleaner to me.

In detail, the passed `flags` are applied before parsing the command line arguments.

@mosra do you feel this is a good solution? Exposing a parameter called `InternalFlags` feels a bit daring, but I guess the Application classes already use that type ;)

### DIFF
--- a/src/Magnum/GL/Context.cpp
+++ b/src/Magnum/GL/Context.cpp
@@ -626,9 +626,9 @@ Context& Context::current() {
 
 void Context::makeCurrent(Context* context) { currentContext = context; }
 
-Context::Context(NoCreateT, Int argc, const char** argv, void functionLoader(Context&)): Context{NoCreate, Utility::Arguments{"magnum"}, argc, argv, functionLoader} {}
+Context::Context(NoCreateT, Int argc, const char** argv, void functionLoader(Context&), InternalFlags flags): Context{NoCreate, Utility::Arguments{"magnum"}, argc, argv, functionLoader, flags} {}
 
-Context::Context(NoCreateT, Utility::Arguments& args, Int argc, const char** argv, void functionLoader(Context&)): _functionLoader{functionLoader}, _version{Version::None} {
+Context::Context(NoCreateT, Utility::Arguments& args, Int argc, const char** argv, void functionLoader(Context&), InternalFlags flags): Context{NoCreate, functionLoader, flags} {
     /* Parse arguments */
     CORRADE_INTERNAL_ASSERT(args.prefix() == "magnum");
     args.addOption("disable-workarounds").setHelp("disable-workarounds", "driver workarounds to disable\n      (see https://doc.magnum.graphics/magnum/opengl-workarounds.html for detailed info)", "LIST")
@@ -644,8 +644,8 @@ Context::Context(NoCreateT, Utility::Arguments& args, Int argc, const char** arg
     /* Decide how to display initialization log */
     if(args.value("log") == "verbose" || args.value("log") == "VERBOSE")
         _internalFlags |= InternalFlag::DisplayVerboseInitializationLog;
-    else if(!(args.value("log") == "quiet" || args.value("log") == "QUIET"))
-        _internalFlags |= InternalFlag::DisplayInitializationLog;
+    else if(args.value("log") == "quiet" || args.value("log") == "QUIET")
+        _internalFlags &= ~InternalFlag::DisplayInitializationLog;
 
     /* Decide whether to enable GPU validation */
     if(args.value("gpu-validation") == "on" || args.value("gpu-validation") == "ON")
@@ -659,6 +659,8 @@ Context::Context(NoCreateT, Utility::Arguments& args, Int argc, const char** arg
     for(auto&& extension: Utility::String::splitWithoutEmptyParts(args.value("disable-extensions")))
         _disabledExtensions.push_back(extension);
 }
+
+Context::Context(NoCreateT, void functionLoader(Context&), InternalFlags flags): _functionLoader{functionLoader}, _version{Version::None}, _internalFlags{flags} {}
 
 Context::Context(Context&& other) noexcept: _version{other._version},
     #ifndef MAGNUM_TARGET_WEBGL
--- a/src/Magnum/GL/Context.cpp
+++ b/src/Magnum/GL/Context.cpp
@@ -912,7 +912,9 @@ bool Context::tryCreate() {
 
     /* Initialize functionality based on current OpenGL version and extensions */
     /** @todo Get rid of these */
-    DefaultFramebuffer::initializeContextBasedFunctionality(*this);
+    if(!(_internalFlags & InternalFlag::NoFramebuffer))
+        DefaultFramebuffer::initializeContextBasedFunctionality(*this);
+
     Renderer::initializeContextBasedFunctionality();
 
     /* Enable GPU validation, if requested */
--- a/src/Magnum/GL/Context.h
+++ b/src/Magnum/GL/Context.h
@@ -747,6 +747,8 @@ class MAGNUM_GL_EXPORT Context {
             DisplayVerboseInitializationLog = DisplayInitializationLog|(1 << 1),
             GpuValidation = 1 << 2,
 
+            NoFramebuffer = 1 << 3,
+
             Default = DisplayInitializationLog
         };
         typedef Containers::EnumSet<InternalFlag> InternalFlags;
--- a/src/Magnum/GL/Context.h
+++ b/src/Magnum/GL/Context.h
@@ -745,7 +745,9 @@ class MAGNUM_GL_EXPORT Context {
         enum class InternalFlag: UnsignedByte {
             DisplayInitializationLog = 1 << 0,
             DisplayVerboseInitializationLog = DisplayInitializationLog|(1 << 1),
-            GpuValidation = 1 << 2
+            GpuValidation = 1 << 2,
+
+            Default = DisplayInitializationLog
         };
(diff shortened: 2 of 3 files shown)


xqms:
Is there a way to see CircleCI results without giving them access to my entire GitHub account?

mosra:
> Is there a way to see CircleCI results without giving them access to my entire GitHub account?

You should only need to log in via GitHub. There's one a bit weird "first time use" behavior -- if you log in, it doesn't redirect you to the actual build log afterwards, but jumps to some "add watched repositories" page, which kinda looks like it wants access to your everything, yeah. Opening the link again when you're already logged in works.

(Unfortunately Travis CI [commited suicide](https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing) and doesn't bother answering my e-mails so I had to switch providers... but this login requirement is the only major downside of CircleCI so far, another is buggy console color rendering but that's a minor wart compared to what I had to regularly go through with Travis.)

The failed build was just some OOM error, and it works after a restart. I get 36 cores but anywhere between 8 and 32 GB RAM for each build, and *sometimes* the ultra-parallel jobs won't fit.

- - -

> Exposing a parameter called `InternalFlags` feels a bit daring

Yeah :sweat_smile: Because this API needs to be public and documented, it should be something else, and without the internal flags that are really not meant to be touched by users :) Give me some time to come up with acceptable naming. Apart from that, the change looks good in general.

xqms:
> You should only need to log in via GitHub

Nope AFAIK you actually need to give *write* access to every public repo you own just to log in:

```
By logging in, you are agreeing to our SaaS Agreement and Privacy Policy.
We ask for read/write access to make your experience seamless on CircleCI.
If you’re not ready to give us access to private projects, choose public repos.
```

Nevermind that I don't want to use it on my repos, I just wanted to view a build output on another user's repo... Anyway, I did that and revoked it immediately afterwards, no problem.

----------------------------------------------------

> Yeah :sweat_smile: Because this API needs to be public and documented, it should be something else, and without the internal flags that are really not meant to be touched by users :)

I actually thought it would be nice to be able to manipulate the logging flags. What I'm doing at the moment to be able to switch off logging for *one* context is:

```c++
// Hide Magnum context initialization for this context
std::vector<const char*> argv{ "dummy",
    "--magnum-log", "quiet"
};
int argc = argv.size();

Platform::GLContext context(argc, argv.data());
```

and I thought if I could directly un-set the DisplayInitializationLog flag that would be somehow cleaner:

```c++
// A silent context without framebuffer.
Platform::GLContext context(GL::Context::InternalFlag::NoFramebuffer);
```

So in my opinion renaming `InternalFlag` to `CreationFlag` or something like that (and documenting it properly) could be an option.

mosra:
I'm not ignoring you, just having a big pile of things to get through :sweat_smile: 

> I actually thought it would be nice to be able to manipulate the logging flags

Yep, that's one of the very common pain points. I realized the remaining `--magnum-blah` options could make sense to be exposed as well, especially the workarounds, gpu validation and extension disabling, and an API similar to application's `Configuration` / `GLConfiguration` might work (and in case of the applications the `GLConfiguration` would then derive from this to share the options).

So that expands the scope a bit. I'll try to implement and merge this until the end of the week, hope it doesn't stall you too much.

Additionally, while working on something unrelated, I realized there's another potential race here: https://github.com/mosra/magnum/blob/df6da797ab3c5a71e519438df68c8b8964b385b0/src/Magnum/GL/Framebuffer.cpp#L130-L134 

Will fix that as well, by just resetting the viewport to zero instead.

xqms:
> So that expands the scope a bit. I'll try to implement and merge this until the end of the week, hope it doesn't stall you too much.

Yes, using a Configuration-style setup sounds good! No worries, I'm fine working with my hacky patch for now :)

mosra:
Oookay, this is one of the "fixing the lightbulb" moments, because...

- Adding a `Context::Configuration` means that `Platform::*Application::Configuration` classes will subclass it, which means they need to pull in `<Magnum/GL/Context.h>`
  - But that header is rather huge, including `<vector>` and `<string>` and I don't want to drag all that nonsense along with every `Application`, so I proceeded in porting the public API away from these and to `Array[View]` and `StringView`
      - Which means I had to implement a bunch of missing APIs such as `StringView::contains()` 
        - And various whitespace trimming functions that used to be only on a `std::string` before
          - But after all it was a great thing, as I got rid of one `std::sort()` as well and just that reduced compile times by 5% and binary size by 10 kB, preprocessed size of the `Context.h` header to half; and the amount of allocations done on startup from about 9253 to 18
            - Which can be further reduced to maybe 2 or 3 using an `ArrayTuple`
              - And with `std::string`s going away there's a bunch of other low hanging fruits to take care of, like avoiding all the string copies when populating a `GL::Shader` because the inputs are mostly just global string literals...
            
Nevertheless, I stopped the recursion at level 5 for now, I'm done with the base work on the Configuration, so the next step should *finally* be introducing the new flag :sweat_smile:

xqms:
Wow, I feel honored to have kicked off *that* chain ;) Let me know if I can help in any way.

mosra:
Okay, finally landed the last bit -- 8aceb25854df1fc35b246c94f90d7bf9e08ee8ac. It's still in the `next` branch and not `master` as I need to be sure that it does the right thing first :) Can you check the commit and tell me if it's working how you expect (or point me to parts that I forgot to handle)? 

Thank you!

xqms:
Sorry for the lag ;) It seems to work just fine (tested `Windowless | QuietLog`), at least it fixes the startup crash I get when running unmodified master :+1:

I can test my full setup with streaming data tomorrow when I'm back in the lab, but I expect no issues there.

xqms:
... actually, there is something missing in `GL::Context::tryCreate()`:

```c++
     /* GPU validation is enabled if either enables it */
    [...]

    /* Same for windowless contexts */
    if(configuration.flags() & Configuration::Flag::Windowless)
        _configurationFlags |= Configuration::Flag::Windowless;
```

Some more info on my usage of the API, in case you need it (or I'm using something wrong):

Main thread:
```c++
        worker.glContext = Platform::WindowlessGlxContext{
            Platform::WindowlessGlxContext::Configuration{}
                .setSharedContext(glxContext)
        };
```

Worker thread:
```c++
    // Init our own GL context
    worker.glContext.makeCurrent();
    Platform::GLContext context(GL::Context::Configuration{}.addFlags(
        GL::Context::Configuration::Flag::Windowless | GL::Context::Configuration::Flag::QuietLog
    ));
```

If I add the snippet above, everything works fine.

mosra:
That omission caused races on `GL::Framebuffer` deletion, right? Other than that I can't see a place where it would matter.

Fixed in abd07b645c2bd2242227572bfc0dab3ee33ac7a0 (`next` branch again, force-pushed), can you check again? Thanks :pray:

xqms:
> That omission caused races on GL::Framebuffer deletion, right? Other than that I can't see a place where it would matter.

No, actually you check `_configurationFlags` for the Windowless flag, not `configuration`. So somehow it needs to be carried over to `_configurationFlags` ;)

https://github.com/mosra/magnum/blob/f357dde89a2263639bfcc24c21b0b60ce374e91e/src/Magnum/GL/Context.cpp#L990-L991

I'll test again.

xqms:
Works fine :+1:

mosra:
Ah, haha, stupid me. The fix uncovered another stupid bug in my change that broke ES2 builds, fixed that now as well. When the CIs get green, I'll merge to master.

mosra:
Okay, the `Windowless` flag is finally in master as 46df57f6d551d0961477b2e0192b5cb391401150.

PR #498, opened 2021-03-02
[EGL]: undef more Xlib nonsense

costashatz:
Undef a few more Xlib nonsense in `WindowlessEglApplication` (similar to `WindowlessGlxApplication`).


mosra:
Thanks!

Issue #293, opened 2018-11-02
Compilation time, CI time and executable size improvements


mosra:
For the 2018.1d release (and onwards) I'd like to focus on reducing the header and executable size, together with improving compile time (and, as a side effect, runtime performance). This was last done in 2013 (see the [blog article](http://blog.magnum.graphics/backstage/reducing-cpp-compilation-time-in-magnum-code-optimizations/)) and while current workflow enforces enough rules to prevent worsening of this problem, it's not actively improving it either.

The ultimate goal is being able to ship useful utilities as "single-header" libraries ~~without being laughed at for compile times~~ and having compile times competitive with C header-only libs, yet staying in C++. Which, of course, means much better compile times than other C++ projects (json.hpp and Eigen, I'm looking at you).

## Compile time improvements

- [x] ~~Add (and start using) `#pragma once` in all headers, as it [leads to measurable compile time improvement](https://twitter.com/ArvidGerstmann/status/996791515526512640)~~
  - below statistical error, extra noise, doesn't actually add any value if guards need to be present as well
- [x] Provide a `CORRADE_TARGET_LIBCXX` / `CORRADE_TARGET_LIBSTDCXX` macro that tells me whether libc++ or libstdc++ is used (needed by the things below) -- detection using `<ciso646>` and _LIBCPP_VERSION, https://stackoverflow.com/questions/31657499/how-to-detect-stdlib-libc-in-the-preprocessor, there needs to be an exception for GCC < 6.1: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=65473 (and a TODO to use `<version>` in C++20) -- mosra/corrade@b6b37fbd2150879d47e18fb9ba115e52ea6551a4
- [x] Provide (and use) a header wrapping platform-specific forward-declaration headers for `std::string` -- mosra/corrade@2345195d5daf704d628a2a7db9ff4ad1dcc977da
    - [x] [libc++](https://github.com/llvm-mirror/libcxx/blob/8c58c2293739d3d090c721827e4217c113ced89f/include/iosfwd#L190-L195),
    - [x] [libstdc++](https://github.com/gcc-mirror/gcc/blob/2a4787da69071b5d5bc178795accca264073b7e4/libstdc%2B%2B-v3/include/bits/stringfwd.h#L68-L73)
    - [x] MSVC STL has a full definition + typedef in `<xstring>`, ruling out any forward declaration
[...]

PR #499, opened 2021-03-03
Bye bye std::string in the GL library

mosra:
As usual, a Saturday afternoon task that turned out to cut much deeper than anticipated. Things to do:

- [x] port `label()` / `setLabel()` away from `std::string` -- bc884428f8c1d8f66ee0216c9a6a338eff31fca4
  - [x] all label tests need to be updated to explicitly check that we handle non-null-terminated strings properly (the GL label APIs have an explicit size, so it *should*, but just in case)
  - [x] duplicate `setLabel()/label()` tests for all Query and Texture subclasses in order to verify these are correctly implemented and exported (since those are all deinlined in new files now)
  - [x] changelog entries, mention all the breaking changes
- [x] port `DebugOutput`, `DebugGroup` away from `std::string`
  - [x] test for non-null-terminated strings
  - [x] explicit test for the deprecated debug output callback
  - [x] changelog entries, mention all the breaking changes
- [x] port `Shader` and `AbstractShaderProgram` away from `std::string`
  - [x] test for non-null-terminated strings
  - [x] `#line` statements added to shader sources need `Utility::format()` to return a `String` -- mosra/corrade@ea9f21790ff88e47a4e0d830165315ecf15644e9
  - [x] `Shader::addFile()` needs `Utility::Path::readString()` to return a `String` -- mosra/corrade@a211a75aa0de82e5d1984ecdb2ef220c209afdd4
  - [x] Shader tests need expanding quite a lot to verify we're correctly copying (or not) if and only if the input view isn't global
    - [x] needs propagating the `Global` flag from `String::nullTerminatedGlobalView()` -- mosra/corrade@f65c6521d5735db44671facf4227d21b7c7a7979
  - [x] changelog entries, mention all the breaking changes
- [x] port the nasty workaround for xfb outputs for NV on Windows
  - [x] test for non-null-terminated strings
  - [x] needs explicit testing from someone with a NV card *and* Windows
- [x] ~~various `<string>` includes may need to be added to various places~~ not anymore, 65a935cedc059bab49172b9a198e3ccc6b346bdc
  - [x] ideally I should first fix `CORRADE_SKIP()` to not need them, because that's the major reason for these -- mosra/corrade@8d077cc24f7fd93bafe3050fcbc72e87deb87b5a
- [x] Drop the `std::pair` from `AbstractShaderProgram::validate()` as it's a breaking change already anyway
  - [x] Take that as a chance to remove it from all internal places in GL as well
- [x] Update `Shader::compile()` to use an ArrayTuple for the output instead of two allocations
- [x] Update builtin shaders to use string view literals, avoiding temporary allocations
  - [x] The actual improvement will happen once `Utility::Resource` starts supporting null-terminated views -- mosra/corrade@9ef4b4aa89c1883529161724cb07bd0788541317
  - [x] Look at the improvements with Heaptrack to be actually sure about the improvement

Related to: #293

### DIFF
--- a/doc/changelog.dox
+++ b/doc/changelog.dox
@@ -1137,6 +1137,12 @@ See also:
 -   As part of the ongoing STL header dependency cleanup, the following
     breaking changes related to @ref std::string and a @ref std::vector are
     done:
+    -   @ref GL::AbstractShaderProgram::validate() now returns a
+        @relativeref{Corrade,Containers::String} instead of a @ref std::string;
+        @ref GL::Shader::sources() now returns a
+        @relativeref{Corrade,Containers::StringIterable} instead of a
+        @ref std::vector of a @ref std::string See also [mosra/magnum#499](https://github.com/mosra/magnum/pull/499)
+        and [mosra/magnum#608](https://github.com/mosra/magnum/pull/608).
     -   @ref GL::Context::vendorString(),
         @relativeref{GL::Context,rendererString()},
         @relativeref{GL::Context,versionString()},
--- a/doc/changelog.dox
+++ b/doc/changelog.dox
@@ -1135,11 +1135,12 @@ See also:
     significant positive effect on compile times of code using the @ref GL,
     @ref Audio, @ref Trade and @ref Text libraries
 -   As part of the ongoing STL header dependency cleanup, the following
-    breaking changes related to @ref std::string and a @ref std::vector are
-    done:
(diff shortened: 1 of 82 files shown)


LB--:
> the nasty workaround for xfb outputs for NV on Windows needs explicit testing from someone with a NV card

I have a GTX 1660 Ti, will that work?

mosra:
Yes, absolutely. I'll keep you in mind when I get back to this :)

PR #504, opened 2021-04-28
Fix cmake warnings when using BUILD_STATIC and AUTOGEN for Qt

asmaloney:
"Warning: Policy CMP0071 is not set: Let AUTOMOC and AUTOUIC process GENERATED files. ..."

Magnum allows really old cmake versions (3.4), so we check the version to ensure the SKIP_AUTOGEN property exists.

Also fix a spelling mistake in WITH_PRIMITIVES option.


asmaloney:
I think something's broken with codecov... There's no way my changes have this effect.

mosra:
Thank you, good to know. I'm not currently using Magnum in combination with Qt in any projects, so I was not aware of this warning.

> There's no way my changes have this effect.

For some reason the CircleCI build, which is responsible for testing GL and Vulkan code, didn't run, not sure why. I don't think that's a problem for this change, though :)

However, I don't really like adding Qt-specific options to random places in code. I was usually fixing these forward compatibility warnings by setting the corresponding CMake property. Can you instead try this, applied to the root CMakeLists?

```patch
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 593c31166..dfa87394c 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,6 +30,12 @@ if(HUNTER_ENABLED)
     include(${CMAKE_CURRENT_LIST_DIR}/package/hunter/HunterInit.cmake)
 endif()
 
+# If CMAKE_AUTOMOC is set, all uses of corrade_add_resource() would otherwise
+# complain on 3.10 that AUTOMOC is not processing GENERATED files
+if(POLICY CMP0071)
+    cmake_policy(SET CMP0071 NEW)
+endif()
+
 project(Magnum CXX)
 
 if(HUNTER_ENABLED)

```

Let me know if this fixes the warnings as well. Thanks!

asmaloney:
Made the change and it works - fixed in commit.

> Thank you, good to know. I'm not currently using Magnum in combination with Qt in any projects, so I was not aware of this warning.

I'm planning to experiment with magnum as a replacement for some Qt3D work I did a while ago. We'll see how it goes 😄 

Thanks for this library Vladimír!

mosra:
Commited as 90126a4ded3f67846f22cc53061d6d6d91ca54c7, thanks for the confirmation. For good measure I replicated this change on the other repos, since these might suffer from the same warning, if not already.

PR #525, opened 2021-07-10
SceneData rework

mosra:
Final bit in the series of MeshData / MaterialData / ... reworks that brings pure data-oriented joy to scene hierarchies in a way that should be flexible enough to be fed into / serialized from common ECS frameworks like entt or flecs.

"MVP":

- [x] Base implementation, tests
  - only unique fields
  - field arrays
  - custom fields
  - convenience type-converting functions
  - support for non-owned and offset-only data and `SceneFieldData` having the same layout on 32 and 64 bits (preparing for #427)
- [x] Per-field flags ("object mapping is sorted", "object mapping is ~~trivial~~ implicit", "field is offset-only")
  - [x] allow object mapping to be omitted if implicit
  - [x] "field is trivial", special-casing so one doesn't need to provide `-1` for scenes that contain just root nodes
- [x] OOP-like interface
  - [x] "get me a mesh/light/... for this object", returning an Optional for unique fields and `Array` otherwise (~~or -1 / identity transform for the "usual" properties like parent and transform? how to distinguish objects that are outside of a hierarchy & without transformations then?~~ postponed)
    - [x] that means one additional convenience API for every convenience accessor, ugh -- could these be handled by a single API, for example by one that takes a list of objects for which we want the property? yes, a shared `fooIntoInternal()` helper
    - [x] ~~OTOH we could provide an iterating API where one would have to switch & cast on the type, handle all possible field types and ...~~ ugh, nope, this is not the way
  - [x] ~~"get all fields that this object has", could be useful for debugging~~ slow as hell, users should instead directly ask for meshes/lights/...
  - [x] "get top-level objects", "get children of an object" ... basically goes through the Parents array and picks objects from there, is a O(n^2) operation, better ideas?
    - [x] "get parent of an object", returns `Optional<Int>` used for checking which scenes does an object belong to -- if `-1`, the object is top-level, if `NullOpt` the object does not belong to this scene
  - [x] speed up the lookup using the sorted/trivial per-field flags
- [x] Tighten and clarify naming ~~("object index" vs "object mapping" vs ??, "field" vs "entry of a field", `SceneObjectType` vs `SceneIndexType`, better idea for `objectCount(SceneField)` that describes "how many entries a field has", since one object can have more than one field entry it's not really "object count", entry count?)~~ object mapping and "field"
- [x] document / figure out what actually an "object ID" is, and how it's meant to be worked with when animations / skins reference it (do I go through all scenes to figure out which one it belongs to? what if multiple scenes reference it?), though it was the exact same issue before)
  - it should also be mentioned that a single object ID can reference both a 2D and a 3D object and the two can reference the same thing or not at all, and the ID can be shared among scenes or not (which affects how the output glTF will look like), and IT NEEDS SOME SANITY RULES
- [x] the `*Into()` / `*AsArray()` helpers should probably give out objects as well?
  - so one can loop over them in a single less-verbose expression
  - so the common assertions get done only once
  - allow the object output view be null in `*Into()`
- [x] docs for filling/usage including OOP-like access
- [x] `SceneFieldType::Pointer` and `SceneFieldType::MutablePointer`, with arbitrary `field<T*>()` accepted for those (same as with MaterialData)
- [x] `SceneField::ImporterState` for per-object importer state, of a `Pointer` type -- needed for compat with existing code
- [x] How do we actually export files with multiple scenes? Since each `SceneData` basically references the exact same node hierarchy, with all its data... 
  - it's just a specifics of glTF / COLLADA and not all file formats in general; glTF material variants extension makes it possible to have the same nodes appear different in different scenes, and what we have now is a further generalization of that
  - the glTF exporter will need to take care of this (e.g. by picking node info only from the first scene that references it, ignoring the (potentially conflicting) duplicate info)
- [x] One of the goals for this representation was to be able to efficiently store a state of a particle system (hundreds of thousands of positions + colors...), however with the restrictions that there can be only one instance of a particular field (so I can't have a separate packed 16-bit Translation field just for the particle system), or requiring that if there's a Scaling then all objects that have a Translation need to have a Scaling as well, it's not efficient anymore
  - ~~could the Archetype support mentioned below solve this (with some additional restrictions)?~~
  - ~~or should I just introduce custom fields for such case, missing out on the opportunity to deal with them through the common interfaces?~~
  - or eventually have ~~sub-scene references, where the sub-scene uses a different field set and encoding?~~ scene layers like with materials, where there can be a subrange of fields for each layer, and the fields can repeat?
  - ~~or some "<Whatever> field, variant 2" that allows certain sub-parts of the scene to use custom encoding?~~ that's no different than a sub-scene / sub-layer
  - or eventually lift the restriction of having the same object mapping and reduce just to sorted (so the combined transform etc can still be processed in linear time)
- [x] ~~More systematic way to deal with 64-bit object IDs and references. We need them to store arbitrary object handles, but the convenience APIs all have 32-bit as it's very unlikely that there actually will *ever* be >4G objects.~~ postponed
  - [x] Switching the `fooFor()` to 64-bit probably makes sense tho
  - One possible option would be to introduce some "object ID hash" field that would provide (contiguous?) remapping of the object ID to 32-bit values, and the `Into()` and `AsArray()` accessors would return this remapping instead of the original (and the data provider would be forced to provide such field?). But then again it'd cause issues with for example animation references that would still reference the original 64-bit value, there one would have to look up the hash first...
- [x] ~~change `mappingBound()` to a Range? / rename to `mappingRange()`?~~ postponed

Hooking up into the rest:

- [x] a standalone tool that converts the `SceneData` to have each object just one mesh/light/camera/skin, to support classic workflows and make the backwards-compat wrappers easier to write 
  - will get evenually publicly exposed in the new `SceneTools` library
- [x] Write deprecated backward-compat wrappers for `SceneData::children2D()` / `children3D()`
- [x] Deprecate `object2D()` / `object3D()` in `AbstractImporter`, write backwards-compat functions that extract this out of the new SceneData using the (slow) per-object convenience APIs
  - [x] also handle the expansion of multi-mesh objects using the tool above
- [x] `AbstractImporter::objectForName()` / `objectName()` (drop the 2D/3D distinction), works on objects up to ~~`SceneData::objectCount()`~~ `AbstractImporter::objectCount()`
  - [x] bump the interface string to avoid ABI-break-related crashes
  - [x] ~~huh... but how to distinguish between scenes and have it robust w/o having to retrieve scenedata first? or have `objectForName(scene, name)` and `objectCount(scene)` on the importer directly? ahhhh~~ there's a global object count (which is needed anyway since animations, skins etc reference objects globally also)
  - [x] ~~or drop this and store names in SceneData directly? might be wasteful if the users don't need that, doesn't allow searching... OTOH it'd be cool to have a builtin possibility to put them there so the SceneData can be made self-contained~~ nope, handle externally
- [x] `Trade::AnimationData` need to have a 64-bit object ID now
- [x] `AbstractImporter::sceneFieldName()` / `sceneFieldForName()` for custom fields
- [x] `SceneField::SkinId` needs to be distinguished for 2D and 3D, I suppose? Or have a `is2D()` / `is3D()` getter on the SceneData itself, ~~which looks for transformation fields and then decides? What if neither is available?~~ checked and populated in the constructor, skin fields require a transformation field to disambiguate the dimensionality
- [x] How do we deal with objects that have multiple meshes assigned, how does one understand skins there? is the same skin applied to all meshes? ~~or just to the first and it's expected that extra meshes have extra skins?~~ -- same skin for all meshes
  - [x] Make this field get copied when converting to single-functioning objects for backwards compatibility
- [x] `SceneField::MeshMaterial` needs to be signed because otherwise it's impossible to represent meshes without materials (as it has to have the same object mapping as meshes)
- [x] Adapt AnySceneImporter
- [x] ~~Adapt ObjImporter~~ it doesn't have any scene / material assignment yet, nothing to do
- [x] Adapt TinyGltfImporter
  - [x] ~~Currently it fails if there's no scene that'd list the root nodes, implement a fallback (with a warning) that collects all root nodes instead~~ postponed, needs a BitArray
  - [x] ~~And consequently warn also if there are scenes but also loose nodes not referenced by any of them~~ postponed, needs a BitArray
- [x] Adapt CgltfImporter
  - [x] ~~same handling as TinyGltf for scene-less files~~ postponed
- [x] Adapt AssimpImporter
  - [x] ~~might need the same handling as TinyGltf for scene-less files~~ nope, nodes are there only if a scene is
- [x] Adapt OpenGexImporter
  - [x] ~~might need the same handling as TinyGltf for scene-less files~~ nope, there's just one implicit scene
- [x] Adapt PrimitiveImporter (it has both 2D and 3D scenes!), a nice use case for constexpr SceneData
- [x] Adapt plugins to make use of per-field flags, ideally everything at least Ordered, implicit object mappings omitted
- [x] update docs for all plugins, mention which fields are always present, which not, what types, shared object mapping, ordering/triviality...
- [x] add to `magnum-sceneconverter --info` (listing what fields a scene has and how many objects is for each)
- [x] adapt `DartIntegration`
- [x] adapt the viewer example
- [x] adapt `magnum-player`
  
Followup tasks:
  
- [x] ~~`SceneFieldType::String`, internally represented as begin+size pairs, with the data being taken from the data blob, or as offsets if the field is offset-only; begin+size instead of implicit offset array in orde to allow the same string being shared between multiple fields~~ postponed
- [x] It should assert that all transformations are either 2D or 3D ~~or allow 2D transforms used for 3D and vice versa, like MeshData does for positions~~ no (at least for now), conversion complexity not worth the 0.001% use case
- [x] ~~Support the `Bool` type~~ postponed
- [x] ~~Support storing direct (non-owned) pointers to `MeshData`, `MaterialData` if one so desires (could be useful for the upcoming `SceneTools` library as well)~~ postponed
- [x] ~~Support sparse fields (highest bit of a 32-/64-bit object index used to distinguish "dead" entries), for easier (de)serialization of live systems~~ postponed
- [x] ~~Additional per-field flags ("field is sparse", "field is ordered", ...), flags for describing field data in additon to object data~~ postponed
- [x] ~~`SceneField::Present` / `Belongs` / ... ? which is a bit array and enumerates which objects belong to the scene~~ postponed
- [x] ~~`SceneField::InstanceMesh` to support glTF's [EXT_mesh_gpu_instancing](https://github.com/KhronosGroup/glTF/blob/master/extensions/2.0/Vendor/EXT_mesh_gpu_instancing/README.md)~~ postponed
  - [x] ~~`CollisionMesh`?~~ postponed
- [x] ~~Mesh view properties (index/vertex offset and count, instance offset and count), all should share the same object mapping; all should be retrievable using a single convenience API that fills a few separate views~~ currently doable via custom fields, postponed to when importer support is there
  - [x] Provide a corresponding overload to `GL::AbstractShaderProgram::draw()` that takes raw data instead of a ton of `GL::MeshView` references -- c2ecaa6abf641c7cf44668bb46afa53667770b75
    - [x] ~~Adapt the method chaining overload in all builtin shaders for that (make a macro!!)~~ postponed
- [x] ~~Material texture transform + layer properties that get applied on top of the referenced material~~ postponed
- [x] ~~Instead of failing, synthesize parent + transform info in the `AsArray()` helpers if not present -- `-1` / identity for all objects? need to think about how to distinguish transformation-less objects in this case~~ not really possible, transformless objects are common (plus the type distinguishes between 2D and 3D), non-node objects are also a thing now
  - [x] same for (now signed) `MeshMaterial`, all `-1`s
- [x] ~~`transformationObjectsAsArray()` helper that picks object mapping for `Transformation` / TRS fields depending on which of them is available, currently this has to be done by the user and is error-prone~~ solved by the `Into` / `AsArray` helpers returning the object mapping as well
- [x] ~~Real-world test sharing the data with entt/flecs?~~ i don't see why this new way would be worse than the cursed OOP API that was there before
  - [x] ~~`EnttIntegration` using https://github.com/skypjack/entt/wiki/Crash-Course:-entity-component-system#snapshot-complete-vs-continuous ?~~ postponed
  - [x] The `SceneField` probably should be 32-bit to have enough bits for arbitrarily allocated entity IDs (hopefully that's enough and we don't need 64)
  
Open Qs:

- [x] Storing octrees / kD trees -- the `Parent` field currently denotes that an object is a hierarchical node; so any other (custom) field could be assigned to objects to denote those are octree cells; encoding of the octree is up to the user
- [x] ~~Storing child information (inverse of the parent information) -- allows users to pick better representation for given task, `parentsAsArray()` / `childrenAsArray()` then falls back to querying the information from the other field if the primary is not available~~ postponed
- [x] ~~`SceneFieldType::SceneField`? e.g. for a potential accelerating `SceneField::InverseMapping` field that maps object IDs to fields (i.e., is sorted by object ID, or is trivial and sparse ...)~~ can't see why this would be useful anymore
- [x] ~~Think more about `SceneField::Parent`, if it really needs to be self-referencing (what else the object mapping view would be? the exact same view?), and if it could be restricted to transforms only~~ not self-referencing, but the Parent field implies the object is a node
  - [x] ~~rename `Parent` to `ParentNode`? to make room for parent octree cell etc~~ postponed
- [x] ~~Support sparse object IDs (for easier (de)serialization of live systems, again)~~ postponed
- [x] ~~Support more than one array per field (for easy storage of ["archetypes"-like ECS state](https://ajmmertens.medium.com/building-an-ecs-2-archetypes-and-vectorization-fe21690805f9))~~ postponed

### DIFF
--- a/doc/artwork/scenedata-dod.svg
+++ b/doc/artwork/scenedata-dod.svg
@@ -1,0 +1,1610 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
+   width="530"
+   height="240.00002"
+   viewBox="0 0 140.22916 63.500004"
+   version="1.1"
+   id="svg1701"
+   sodipodi:docname="scenedata-dod.svg"
+   inkscape:version="1.1.1 (3bf5ae0d25, 2021-09-20, custom)"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
+   xmlns="http://www.w3.org/2000/svg"
+   xmlns:svg="http://www.w3.org/2000/svg"
+   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+   xmlns:cc="http://creativecommons.org/ns#"
+   xmlns:dc="http://purl.org/dc/elements/1.1/">
+  <sodipodi:namedview
+     pagecolor="#ffffff"
+     bordercolor="#666666"
+     borderopacity="1"
+     objecttolerance="10"
+     gridtolerance="10"
+     guidetolerance="10"
+     inkscape:pageopacity="0"
+     inkscape:pageshadow="2"
+     inkscape:window-width="1920"
+     inkscape:window-height="1029"
+     id="namedview12294"
+     showgrid="true"
+     fit-margin-top="0"
+     fit-margin-left="0"
+     fit-margin-right="0"
+     fit-margin-bottom="0"
+     inkscape:zoom="1.3009434"
+     inkscape:cx="290.17404"
+     inkscape:cy="51.116751"
+     inkscape:window-x="0"
+     inkscape:window-y="0"
+     inkscape:window-maximized="1"
+     inkscape:current-layer="layer9"
+     inkscape:document-rotation="0"
+     inkscape:pagecheckerboard="0"
+     inkscape:document-units="px"
+     units="px"
+     showguides="true"
+     inkscape:guide-bbox="true">
+    <inkscape:grid
+       type="xygrid"
+       id="grid12857"
+       originx="-18.520834"
+       originy="-15.875007" />
+  </sodipodi:namedview>
+  <defs
+     id="defs1695" />
+  <metadata
+     id="metadata1698">
+    <rdf:RDF>
+      <cc:Work
+         rdf:about="">
+        <dc:format>image/svg+xml</dc:format>
+        <dc:type
+           rdf:resource="http://purl.org/dc/dcmitype/StillImage" />
+      </cc:Work>
+    </rdf:RDF>
+  </metadata>
+  <g
+     id="layer3"
+     style="display:none"
+     transform="translate(-25.135421,-55.562503)">
+    <path
+       style="font-variation-settings:normal;opacity:1;vector-effect:none;fill:none;fill-opacity:1;stroke:#2f83cc;stroke-width:0.132292;stroke-linecap:butt;stroke-linejoin:miter;stroke-miterlimit:4;stroke-dasharray:none;stroke-dashoffset:0;stroke-opacity:1;stop-color:#000000;stop-opacity:1"
+       d="m 44.979165,64.822918 c -0.04047,7.883497 32.952175,0.947985 33.07292,9.128124"
+       id="path2871" />
+    <path
+       style="font-variation-settings:normal;opacity:1;vector-effect:none;fill:none;fill-opacity:1;stroke:#2f83cc;stroke-width:0.132292;stroke-linecap:butt;stroke-linejoin:miter;stroke-miterlimit:4;stroke-dasharray:none;stroke-dashoffset:0;stroke-opacity:1;stop-color:#000000;stop-opacity:1"
+       d="m 44.979165,64.822918 c -0.04047,7.883497 14.431342,11.66361 14.552087,19.843749"
+       id="path2936" />
+    <path
+       style="font-variation-settings:normal;opacity:1;vector-effect:none;fill:none;fill-opacity:1;stroke:#2f83cc;stroke-width:0.132292;stroke-linecap:butt;stroke-linejoin:miter;stroke-miterlimit:4;stroke-dasharray:none;stroke-dashoffset:0;stroke-opacity:1;stop-color:#000000;stop-opacity:1"
+       d="M 44.979165,64.822918 C 44.938695,72.706415 23.691757,80.322987 23.812502,88.503126"
+       id="path2869" />
+    <path
+       style="font-variation-settings:normal;opacity:1;vector-effect:none;fill:none;fill-opacity:1;stroke:#2f83cc;stroke-width:0.132292;stroke-linecap:butt;stroke-linejoin:miter;stroke-miterlimit:4;stroke-dasharray:none;stroke-dashoffset:0;stroke-opacity:1;stop-color:#000000;stop-opacity:1"
+       d="M 23.812502,54.239583 C 23.772032,62.12308 17.077174,67.226111 17.197919,75.40625"
+       id="path2865" />
+    <path
+       style="font-variation-settings:normal;opacity:1;vector-effect:none;fill:none;fill-opacity:1;stroke:#2f83cc;stroke-width:0.132292;stroke-linecap:butt;stroke-linejoin:miter;stroke-miterlimit:4;stroke-dasharray:none;stroke-dashoffset:0;stroke-opacity:1;stop-color:#000000;stop-opacity:1"
+       d="m 46.55523,43.65625 c -0.04047,7.883497 19.469856,6.239654 19.590601,14.419793"
+       id="path2861" />
+    <path
+       style="font-variation-settings:normal;opacity:1;vector-effect:none;fill:none;fill-opacity:1;stroke:#2f83cc;stroke-width:0.132292;stroke-linecap:butt;stroke-linejoin:miter;stroke-miterlimit:4;stroke-dasharray:none;stroke-dashoffset:0;stroke-opacity:1;stop-color:#000000;stop-opacity:1"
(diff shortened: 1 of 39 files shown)


mosra:
Merged the current state, few bits got scrapped and put aside into #542 because they're too cursed. Items marked as postponed now live in the project Projects board and will be gradually implemented in the future.

PR #527, opened 2021-07-18
Android multitouch events support

nodoteve:
AndroidApplication::inputEvent() rework mainly


mosra:
Hey, sorry for totally neglecting your PR for so long. Only now I got to actually implementing multi-touch support everywhere, and it's done by unifying mouse, pen and touch input into a single `PointerEvent` abstraction. For Android the relevant changes are in b0abae88d1b44e0c4c5e33afea2dc7850f7fb0ff and 005d4c224fcf4e6eae972046d6a7fe43ce3172d3. I made use of some of your code, but limited the pointer position list to just 32 items at most to not need to use a generic growable array there. Which I think should be fine.

[High-level intro docs to the unified pointer events are here.](https://doc.magnum.graphics/magnum/platform.html#platform-windowed-pointer-events)

PR #537, opened 2021-10-17
RPM build

1b00:
To build RPM packages into the default location cd into package/rpm, read the docs and run build_rpm.sh (with or without parameter: pass either 'master' or any valid git tag|hash or no parameter for last tag)

### DIFF
--- a/package/rpm/README.md
+++ b/package/rpm/README.md
@@ -1,0 +1,28 @@
+# Install next packages
+
+## for usage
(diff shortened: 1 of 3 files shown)


mosra:
Sorry for neglecting your contribution for so long. This was now combined with #650 and merged as d3b83f090c15cc0e31ac182baa2b52996420a95a.

PR #539, opened 2021-10-24
GL: new "angle-instanced-attributes-always-draw-instanced" workaround

mosra:
Because the bugs keep piling up faster than I'm able to find solutions, sigh.

The original problem, as reported by @kleisauke: Pressing <kbd>S</kbd> and shooting the first sphere in the [web Bullet example](https://magnum.graphics/showcase/bullet/) gives a `Vertex buffer is not big enough for the draw call.` WebGL error -- an ANGLE-specific bug where instanced attributes don't get bounds-checked properly with non-instanced `glDraw*()`. The workaround is to use `glDraw*Instanced()` unconditionally if instanced attributes are present. 

I'm hesitant to do this unconditionally because who knows what other bugs or perf drops on other implementations / ANGLE backends this would uncover.

The workaround implementation is correct and was verified to work, however I'm stuck on crafting a regression test case to prevent this from resurfacing again in the future. Things left to do:

- [x] fix MeshGLTest to actually produce *something* on ANGLE+D3D, currently it's all zeros (should I set `gl_PointSize`?!)
- [x] the attempt to reproduce this problem in the test didn't really work (not leading to a GL error if workaround is not enabled), fix (was it due to the above issue? but i suppose the validation would still happen there, no?)

### DIFF
--- a/src/Magnum/GL/Implementation/driverSpecific.cpp
+++ b/src/Magnum/GL/Implementation/driverSpecific.cpp
@@ -520,32 +520,36 @@ auto Context::detectedDriver() -> DetectedDrivers {
     if(renderer.contains("ANGLE"_s))
         *_detectedDrivers |= DetectedDriver::Angle;
     #ifdef MAGNUM_TARGET_WEBGL
-    /* Otherwise assume ANGLE is present if WEBGL_debug_renderer_info isn't
-       present (which would tell us so already) and the ANGLE_instanced_arrays
-       is present on WebGL 1. Although e.g. Firefox exposes it even though it
-       renders directly through GL drivers on Linux, so this may catch more
-       drivers than just ANGLE. */
+    /* If the unmasked renderer string is not available, try other means */
     /** @todo this (and below) incorrectly detects ANGLE on FF 92+, as it has
         the WEBGL_debug_renderer_info disabled */
-    #ifdef MAGNUM_TARGET_GLES2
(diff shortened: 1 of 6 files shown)


mosra:
Finished with great help from @pezcode (thanks!).

src/Magnum/GL/Implementation/driverSpecific.cpp
```diff
@@ -520,32 +520,36 @@ auto Context::detectedDriver() -> DetectedDrivers {
...
             *_detectedDrivers |= DetectedDriver::Angle;
+        /* Otherwise try to detect a D3D ANGLE backend by querying line width.
+           It's always exactly just 1 on D3D, usually (but not always) more on
+           GL, not sure about Metal. So this is not a 100% match. Sources:
+           http://stackoverflow.com/a/20149090 and http://webglreport.com */
+        else {
+            Range1Di range;
+            glGetIntegerv(GL_ALIASED_LINE_WIDTH_RANGE, range.data());
+            if(range.min() == 1 && range.max() == 1 && vendor != "Internet Explorer"_s) {
+            !Debug{};
```
pezcode, line 561: 😄 Leftover from printf debugging?

mosra: Oh yeah this particular piece of code looked complex in the diff and I ignored it assuming I knew what was I doing back then. Heh. Thanks!

PR #547, opened 2022-01-15
Platform: add missing modifier and command keys to Sdl2Application

pezcode:
These keys already exist in `GlfwApplication` and `EmscriptenApplication`, enum names are taken from there. Key codes exist as far back as [SDL 2.0](https://github.com/libsdl-org/SDL/blob/release-2.0.0/include/SDL_keycode.h).


mosra:
Thank you! Merged as b1ec6aa5b86053b03f32cb71e9762a90bb60dc4a and d5db913a77cde07d1fe2cb9700c6da588e4a068f.

PR #549, opened 2022-01-20
Fix invalid iterator access in Device.cpp / Instance.cpp

Samulus:
Hi @mosra, thanks for the awesome library.

I've been using the Vulkan addons for Magnum and I ran into a crash with address sanitizer enabled on Visual Studio 2022. Pretty simple fix.

### DIFF
--- a/src/Magnum/Vk/Device.cpp
+++ b/src/Magnum/Vk/Device.cpp
@@ -861,7 +861,7 @@ template<class T> void Device::initializeExtensions(const Containers::ArrayView<
             const auto found = std::lower_bound(knownExtensions.begin(), knownExtensions.end(), extension, [](const Extension& a, const T& b) {
                 return a.string() < static_cast<const Containers::StringView&>(b);
             });
-            if(found->string() != extension) continue;
+            if(found == knownExtensions.end() || found->string() != extension) continue;
             _enabledExtensions.set(found->index(), true);
         }
     }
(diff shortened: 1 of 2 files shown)


mosra:
Hi, thank you! Oh yes, `std::lower_bound()`, an API that's just impossible to use correctly. Sorry about that, merged as 390d3fc06f6cb7982f3031a9bef544dedac29349. I should really create my own binary search function that's not so error-prone...

You're one of the very first users of the still-not-fully-finished Vulkan APIs, how do you like them? Even a drive-by feedback is valuable to me. Thanks! :)

Samulus:
I'm a big fan of the API! I had some issues with getting `flextVk` to work with KHR_Swapchain (the generator script had to be patched because it was failing on `vk_khr_maintenance1`) and issues with creating an SDL2 window with a Vulkan context (had to remove the `SDL_WINDOW_OPENGL` flag from Magnum code). I'll post issues for both of those when I get a chance.

mosra:
Ah, yes, those are precisely the missing parts :)

- for flextVk, there seems to be something unexpected happening with version 1.2.191+, if you set the `version` at the top to 1.2.190 and add `KHR_swapchain`, it works -- I'll look into that and fix it properly (I should enable recurring CI builds for the flextGL repo to catch those early)
- for SDL, I think going with [Configuration::WindowFlag::Contextless](https://doc.magnum.graphics/magnum/classMagnum_1_1Platform_1_1Sdl2Application_1_1Configuration.html#aaedcde29393bd4be1788b5d0ebb7a0bfa6ccd70de20161a61101cfd18027aa795) should make it not create the GL context, or at least that was the intention when the flag got added back in 2018

We'll be soon going back to work on the Vk library -- 1.3 features, swapchain creation, support in Application classes and builtin shaders are first things on the list.

PR #554, opened 2022-02-21
GL: add Framebuffer::InvalidationAttachment::DepthStencil

pezcode:
Small quality-of-life improvement. May not make any real difference in the driver since the [relevant extension](https://www.khronos.org/registry/OpenGL/extensions/EXT/EXT_discard_framebuffer.txt) mentions that you can invalidate a packed depth-stencil attachment by invalidating *either* of the two, but it's useful for code expressiveness.

### DIFF
--- a/src/Magnum/GL/Framebuffer.h
+++ b/src/Magnum/GL/Framebuffer.h
@@ -256,6 +256,21 @@ class MAGNUM_GL_EXPORT Framebuffer: public AbstractFramebuffer, public AbstractO
                  */
                 static const InvalidationAttachment Stencil;
 
+                #ifndef MAGNUM_TARGET_GLES2
+                /**
+                 * @brief Invalidate both depth and stencil buffer
+                 * @m_since_latest
+                 *
+                 * @m_keywords{GL_DEPTH_STENCIL_ATTACHMENT}
+                 * @requires_gles30 Combined depth and stencil attachment is
+                 *      not available in OpenGL ES 2.0. Invalidate both
+                 *      @ref InvalidationAttachment::Depth and
+                 *      @ref InvalidationAttachment::Stencil instead.
+                 * @todo Support this in ES2 (invalidate both depth and stencil internally)
+                 */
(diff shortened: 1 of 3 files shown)


src/Magnum/GL/Framebuffer.h
```diff
@@ -256,6 +256,21 @@ class MAGNUM_GL_EXPORT Framebuffer: public AbstractFramebuffer, public AbstractO
                  */
                 static const InvalidationAttachment Stencil;
 
+                #ifndef MAGNUM_TARGET_GLES2
+                /**
+                 * @brief Invalidate both depth and stencil buffer
```
Squareys, line 261: Should maybe be "Invalidate combined depth and stencil buffer" as in the `@requires_gles30` comment below?

pezcode: I copypasta'd from here: https://github.com/mosra/magnum/blob/5f54cc4702ac101c6d598933b2179d659b0c83b9/src/Magnum/GL/Framebuffer.h#L207 But I have no objection to either version 🌞

Squareys: @mosra I vote for updating the docs in Framebuffer.h to clarify the distiction between `Depth|Stencil` and `DepthStencil` :)

mosra: Ah wait, so if I invalidate `DepthStencil` but the framebuffer has separate depth and stencil attachment, it'll cause a GL error? If it does, then it should indeed say "Combined depth and stencil buffer" in both cases, yes. If not, then what's there is fine.

(Heh, it's been almost 3 months since this got opened? Sorry.)

pezcode: Added the extra tests, seems to work 🙆‍♂️

mosra:
Merged as 95f4124b5bce51156c1d100a8d7d6a7ea0822b59 (somehow my suggestions caused whitespace on EOL :eyes:, so I cleaned them up). Thank you!

PR #556, opened 2022-03-26
Include <string> to fix errors in AbstractSceneConverter and Abstract…

EhWhoAmI:
…ImageConverter when compiling when defined MAGNUM_BUILD_DEPRECATED.

### DIFF
--- a/src/Magnum/Trade/AbstractImageConverter.h
+++ b/src/Magnum/Trade/AbstractImageConverter.h
@@ -31,6 +31,8 @@
 
 #include <Corrade/PluginManager/AbstractManagingPlugin.h>
 
+#include <string>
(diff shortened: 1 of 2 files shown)


mosra:
Thank you! Sorry that you were the first to discover this, it was caused by a recent cleanup of STL string usage in the plugin manager and apparently somehow none of the CI builds use these interfaces without also explicitly including `<string>` or `<sstream>`.

I modified your fix to do a forward declaration of `std::string` on implementations that are able to do that, and including it also only if the deprecated APIs are needed so it doesn't unconditionally bloat the header again. Pushed as 54394e2c2f0029aa1f767ac895829d4070616328.

PR #557, opened 2022-04-06
Math: add AABB and bounding sphere algorithms

pezcode:
:wave: This PR adds a function for generating (approximate) bounding spheres to `MeshTools`. It lives in the new `BoundingVolume.h` header (+ matching `.cpp`).

The algorithm used is [*"Bouncing Bubble: A fast algorithm for Minimal Enclosing Ball problem"*](https://www.grin.com/document/204869), specifically the "Ritter substitution algorithm" proposed in the paper. Time complexity between Ritter and this algorithm is identical (3 passes over all position values) but it produces spheres with smaller radius. One side effect of this algorithm is that the radius is always at least some minimal hard-coded value, currently set to `TypeTraits<Float>::epsilon()`. Since the use case for this function is primarily expected to be some sort of culling, this seems acceptable to me.

Because it was pretty straight-forward, I also added a function for generating axis-aligned bounding boxes. It's just a wrapper around `minmax()` but it doesn't hurt to have.

### DIFF
--- a/src/Magnum/MeshTools/BoundingVolume.cpp
+++ b/src/Magnum/MeshTools/BoundingVolume.cpp
@@ -1,0 +1,79 @@
+/*
+    This file is part of Magnum.
+
+    Copyright © 2010, 2011, 2012, 2013, 2014, 2015, 2016, 2017, 2018, 2019,
+                2020, 2021, 2022 Vladimír Vondruš <mosra@centrum.cz>
+    Copyright © 2022 Pablo Escobar <mail@rvrs.in>
+
+    Permission is hereby granted, free of charge, to any person obtaining a
+    copy of this software and associated documentation files (the "Software"),
+    to deal in the Software without restriction, including without limitation
+    the rights to use, copy, modify, merge, publish, distribute, sublicense,
+    and/or sell copies of the Software, and to permit persons to whom the
+    Software is furnished to do so, subject to the following conditions:
+
+    The above copyright notice and this permission notice shall be included
+    in all copies or substantial portions of the Software.
+
+    THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+    IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+    FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+    THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+    LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+    FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+    DEALINGS IN THE SOFTWARE.
+*/
+
+#include "BoundingVolume.h"
+
+#include <Corrade/Containers/StridedArrayView.h>
+
+#include "Magnum/Math/FunctionsBatch.h"
+
+namespace Magnum { namespace MeshTools {
+
+Range3D boundingBoxAxisAligned(const Containers::StridedArrayView1D<const Vector3>& points) {
+    return Math::minmax(points);
+}
+
+Containers::Pair<Vector3, Float> boundingSphereBouncingBubble(const Containers::StridedArrayView1D<const Vector3>& points) {
+    /* See comment about radius below, this is done for consistency */
+    if(points.isEmpty()) return {{}, Math::TypeTraits<Float>::epsilon()};
+
+    /** @todo Skip NaNs here? To match behaviour of boundingBoxAxisAligned()
+        which uses minmax(). */
+    Vector3 center = points[0];
+    /* Radius ends up in the denominator in the first loop so we can't
+       initialize it to 0. Unfortunately this also means the returned radius is
+       always above or equal to epsilon. */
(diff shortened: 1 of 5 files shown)


src/Magnum/MeshTools/BoundingVolume.cpp
```diff
@@ -0,0 +1,80 @@
...
+       always above or equal to epsilon. */
+    Float radius = Math::TypeTraits<Float>::epsilon();
+    Float radiusSquared = radius*radius;
+
+    for(int i = 0; i < 2; ++i) {
+        for(const Vector3& p : points) {
+            const Float ds = (p - center).dot();
+            if(ds > radiusSquared) {
+                const Float alphaInv = radius/Math::sqrt(ds);
+                const Float alphaSqInv = alphaInv*alphaInv;
```
mosra, line 58: Wouldn't this lead to less precision loss, maybe? Not sure how important tho (and I got no experience with testing numerical stability of algorithms either).

```suggestion
                const Float alphaSqInv = radius*radius/ds;
```

src/Magnum/MeshTools/BoundingVolume.h
```diff
@@ -0,0 +1,71 @@
...
+
+namespace Magnum { namespace MeshTools {
+
+/**
+@brief Calculate an axis-aligned bounding box
+@param positions    Vertex positions
+@return Bounding box
+@m_since_latest
+
+This is a convenience wrapper around @ref Math::FunctionsBatch::minmax().
```
mosra, line 48:
```suggestion
Same as @ref Math::minmax(const Corrade::Containers::StridedArrayView1D<const T>&).
```

src/Magnum/MeshTools/Test/BoundingVolumeTest.cpp
```diff
@@ -0,0 +1,217 @@
...
+        constexpr Vector3 translation{1.0f, 2.0f, 3.0f};
+        MeshTools::transformPointsInPlace(Matrix4::rotationY(degrees)*Matrix4::translation(translation),
+            cubeMesh.mutableAttribute<Vector3>(Trade::MeshAttribute::Position));
+        const Containers::Pair<Vector3, Float> sphere =
+            MeshTools::boundingSphereBouncingBubble(cubeMesh.attribute<Vector3>(Trade::MeshAttribute::Position));
+        CORRADE_COMPARE(sphere.second(), Math::sqrt(3.0f));
+    }
+}
+
+void BoundingVolumeTest::benchmarkSphereBouncingBubble() {
```
mosra, line 201: Just to have a reference point to compare this to, maybe add a second benchmark with the ~~minmax~~ AABB?

mosra (commented):
This is great, thank you! 

Mostly just minor / style things, nothing serious.

src/Magnum/MeshTools/BoundingVolume.cpp
```diff
@@ -0,0 +1,80 @@
...
+       initialize it to 0. Unfortunately this also means the returned radius is
+       always above or equal to epsilon. */
+    Float radius = Math::TypeTraits<Float>::epsilon();
+    Float radiusSquared = radius*radius;
+
+    for(int i = 0; i < 2; ++i) {
+        for(const Vector3& p : points) {
+            const Float ds = (p - center).dot();
+            if(ds > radiusSquared) {
+                const Float alphaInv = radius/Math::sqrt(ds);
```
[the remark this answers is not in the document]
pezcode: Is this a pattern that compilers would pick up? I checked the implementation of `sqrtInverted()` in magnum and it was just `1/sqrt()` so it didn't feel too useful.

src/Magnum/MeshTools/BoundingVolume.cpp
pezcode: I... don't really know. My intention here was to avoid a division. Let me read up on that a little bit.

pezcode:
Thanks for the thorough review, much appreciated 😊 There's one open question regarding numerical stability, I'll get back to you on that after some reading

src/Magnum/MeshTools/BoundingVolume.cpp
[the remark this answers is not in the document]
mosra: Yeah ... there are platform-specific `rsqrt` intrinsics, and this function is meant to a placeholder for those. But until I get to actually start using those intrinsics everywhere (which needs mosra/corrade#115 and a good deal of extra testing), it'll be implemented using the standard interface, looking silly :)

src/Magnum/MeshTools/BoundingVolume.cpp
mosra: Not sure if this idea is helpful at all, but I suppose the smaller precision loss could make the calculated output more stable when tried on randomly shuffled position arrays. Don't know how to measure that better :sweat_smile:

src/Magnum/MeshTools/BoundingVolume.cpp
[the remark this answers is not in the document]
pezcode: 👌

src/Magnum/MeshTools/BoundingVolume.cpp
pezcode: Done 😊 I was overthinking this a little bit

mosra:
Thank you! Merged as 7f4500d9e02f74a1d27f9ba99a8c3a3e01e5a808...90b53798c22dec45fcd3a232b68adf071884d2cb. While cross-linking those with the algorithms in `Math::Intersection`, I realized the AABB calculating function could be named `boundingRange()` instead to be consistent with the intersection algos, so I renamed it. Besides that, I left the public header at just forward declarations to not drag along unneeded dependencies when people would want to use just one of the APIs.

Issue #507, opened 2021-05-25
Building WindowlessEmscriptenApplication on 2.0.17 fails with undefined symbol: malloc


werto87:
Hi, when I compile magnum to wasm i get the error "Could NOT find Corrade (missing: Utility)".
I setup the toolchain file ("set(EMSCRIPTEN_PREFIX "/usr/lib/emscripten")")
and build corrade for wasm.
Then I execute
"makepkg -fp PKGBUILD-emscripten-wasm" in "/magnum/package/archlinux"
and get the output:
==> Making package: emscripten-magnum dev.wasm-1 (2021-05-26T00:44:21 CEST)
==> Checking runtime dependencies...
==> Checking buildtime dependencies...
==> Retrieving sources...
==> Extracting sources...
==> Removing existing $pkgdir/ directory...
==> Starting build()...
-- The CXX compiler identification is Clang 13.0.0
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/lib/emscripten/em++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
CMake Error at /usr/share/cmake-3.20/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
  Could NOT find Corrade (missing: Utility)
Call Stack (most recent call first):
  /usr/share/cmake-3.20/Modules/FindPackageHandleStandardArgs.cmake:594 (_FPHSA_FAILURE_MESSAGE)
  modules/FindCorrade.cmake:605 (find_package_handle_standard_args)
  CMakeLists.txt:51 (find_package)

Configuring incomplete, errors occurred!
See also "/home/walde/aur/magnum/build-emscripten-wasm/CMakeFiles/CMakeOutput.log".
See also "/home/walde/aur/magnum/build-emscripten-wasm/CMakeFiles/CMakeError.log".
==> ERROR: A failure occurred in build().
    Aborting...

/usr/lib/emscripten/system/lib" contains:
libCorradeInterconnect.so, libCorradePluginManager.so, libCorradeTestSuite.so, libCorradeUtility.so

mosra:
Hi!

Corrade was built and installed using the `PKGBUILD-emscripten-wasm` in corrade as well, using the same toolchain file? It's strange that it installed `*.so`s, it should be generating static libraries with the `*.bc` extension. 

[...]

PR #560, opened 2022-05-03
GL/DebugTools: read buffer data on WebGL 2.0

pezcode:
This became a bit of a rabbit hole :eyes:

- `GL::Buffer::data()` and `subData()` are now supported on WebGL 2.0. This is exposed by Emscripten 2.0.17+ as a [`glGetBufferSubData()`](https://github.com/emscripten-core/emscripten/blob/24d442109f044fa2e0fc359916de8b8907ba1add/src/library_webgl2.js#L120-L133) convenience entrypoint. On older Emscripten versions, `data()`/`subData()` are callable but lead to an "unreachable" assert.
- `DebugTools::bufferData()` and `bufferSubData()` now forward to `GL::Buffer::subData()` where available, which now makes it work on desktop GL without requiring `ARB_map_buffer_range` and on WebGL 2.0. Essentially everywhere except WebGL 1.0 and GLES 2.0 without `EXT_map_buffer_range`
- I updated a few GL tests to verify buffer contents on GLES 3 using `DebugTools::bufferData()`. Technically they could also be verified on non-WebGL GLES 2.0 with the right extension, but I didn't feel like adding all the checks for that

Todos:
- [x] Some documentation about requiring Emscripten 2.0.17 on WebGL 2
- [x] _emscripten-webgl2_ doesn't actually find `glGetBufferSubData()`, I probably messed up the Emscripten version check
- [x] On some CIs `DebugTools/BufferData.cpp` isn't being compiled, leading to missing symbols. Investigate.
- [ ] Mac GLES3 CI has mysterious `GL::Context::current(): no current context` asserts in tests

----

Post-merge todos for @mosra:

- [x] Remove the template and de-inline the implementation of `DebugTools::buffer[Sub]Data()`. Originally Buffer used the same API pattern, but it got deprecated and eventually dropped in ecc0ab6a59cf49d09212bd6f69ba58a3a21275d0 for being too error-prone. Time to do that here too.


src/Magnum/DebugTools/BufferData.h
```diff
@@ -47,35 +47,45 @@ namespace Implementation {
...
-@requires_gles Buffer mapping is not available in WebGL.
+@requires_webgl20 Buffer data queries or buffer mapping are not available in
+    WebGL 1.0.
 */
-template<class T> Containers::Array<T> inline bufferSubData(GL::Buffer& buffer, GLintptr offset, GLsizeiptr size) {
+template<class T = char> Containers::Array<T> inline bufferSubData(GL::Buffer& buffer, GLintptr offset, GLsizeiptr size) {
+    if(!size) return {};
+    #if !defined(MAGNUM_TARGET_GLES) || defined(MAGNUM_TARGET_WEBGL)
+    Containers::Array<char> data = buffer.subData(offset, size*sizeof(T));
+    return Containers::Array<T>{reinterpret_cast<T*>(data.release()), std::size_t(size)};
```
pezcode, line 66: This code assumes that `GL::Buffer::subData()` always returns an `Array` with a default deleter. Is that a valid assumption? If so, should I assert for it here?

mosra: Good point -- yeah an `CORRADE_INTERNAL_ASSERT(!data.deleter());` would be great.

Hmm, thinking about this, it'd probably make more sense if I dropped the templated aspect of this. `Buffer::data()` used to have that as well, but it was *so* error-prone ("what is the size? bytes? elements??") that I eventually deprecated and removed that, telling people to do an `arrayCast()` instead. I'll add that as a post-merge TODO.

pezcode: The template parameter is pretty handy but I can see why it's confusing. I've already been using it in "[GL, Text: verify buffer contents on GLES3](https://github.com/mosra/magnum/pull/560/commits/9fa66f00cebe0a3b8a1d503e396dcd012f1e7729)", should I change those already or leave it for post-merge too?

src/Magnum/GL/Buffer.cpp
```diff
@@ -468,11 +468,17 @@ void Buffer::getParameterImplementationDSA(const GLenum value, GLint* const data
 }
 #endif
 
-#ifndef MAGNUM_TARGET_GLES
+#if !defined(MAGNUM_TARGET_GLES) || (defined(MAGNUM_TARGET_WEBGL) && !defined(MAGNUM_TARGET_GLES2))
 void Buffer::getSubDataImplementationDefault(const GLintptr offset, const GLsizeiptr size, GLvoid* const data) {
+    #if !defined(CORRADE_TARGET_EMPSCRIPTEN) || (__EMSCRIPTEN_major__*10000 + __EMSCRIPTEN_minor__*100 + __EMSCRIPTEN_tiny__ >= 20017)
```
mosra, line 473: ~~Heh. I can't see what's wrong with the check here. CI has 1.39.6 and yet it somehow evaluates as true :thinking:~~

Ah, you have `EMPSCRIPTEN` here, haha :)

src/Magnum/GL/Buffer.cpp
```diff
@@ -468,11 +468,17 @@ void Buffer::getParameterImplementationDSA(const GLenum value, GLint* const data
...
     glGetBufferSubData(GLenum(bindSomewhereInternal(_targetHint)), offset, size, data);
+    #else
+    CORRADE_ASSERT_UNREACHABLE("GL::Buffer::subData(): Buffer data queries in WebGL 2.0 possible only since Emscripten 2.0.17", );
```
mosra, line 476: As we discussed on Gitter, i'd make the whole thing fail at compile time, and not at runtime, to prevent an accidental footgun. So wrap this and also `Buffer::subData()` in `#if __EMSCRIPTEN_major__*...`.

Hmm but that means all test code using `DebugTools::bufferSubData()` now has to do `#if __EMSCRIPTEN_major__*...` as well, umm... OTOH, it'd have to do that anyway, to prevent a runtime assert, so I guess that's fine. Yeah, so make this function disabled at compile time instead.

src/Magnum/DebugTools/CMakeLists.txt
```diff
@@ -72,7 +72,7 @@ if(TARGET_GL)
         endif()
     endif()
 
-    if(NOT MAGNUM_TARGET_WEBGL)
+    if(NOT MAGNUM_TARGET_WEBGL OR NOT MAGNUM_TARGET_GLES2)
```
mosra, line 75: The following seems like a less confusing condition to me. Same elsewhere:

```suggestion
    if(NOT (MAGNUM_TARGET_WEBGL AND MAGNUM_TARGET_GLES2)
```

src/Magnum/DebugTools/BufferData.h
```diff
@@ -47,35 +47,45 @@ namespace Implementation {
...
     @ref building-features for more information.
 
 @requires_gles30 Extension @gl_extension{EXT,map_buffer_range} in OpenGL ES
     2.0.
-@requires_gles Buffer mapping is not available in WebGL.
+@requires_webgl20 Buffer data queries or buffer mapping are not available in
+    WebGL 1.0.
 */
-template<class T> Containers::Array<T> inline bufferSubData(GL::Buffer& buffer, GLintptr offset, GLsizeiptr size) {
+template<class T = char> Containers::Array<T> inline bufferSubData(GL::Buffer& buffer, GLintptr offset, GLsizeiptr size) {
```
mosra, line 62: Why the sudden `= char`? I want to de-template this eventually anyway, so the less API behavior changes in the meantime, the better I'd say.

src/Magnum/GL/Test/BufferGLTest.cpp
```diff
@@ -138,15 +142,14 @@ void BufferGLTest::constructFromData() {
     CORRADE_COMPARE(c.size(), 5*4);
     CORRADE_COMPARE(d.size(), 5*4);
 
-    /** @todo How to verify the contents in ES? */
-    #ifndef MAGNUM_TARGET_GLES
-    CORRADE_COMPARE_AS(Containers::arrayCast<Int>(a.data()),
+    #ifndef MAGNUM_TARGET_GLES2
+    CORRADE_COMPARE_AS(DebugTools::bufferData<Int>(a),
```
mosra, line 146: In the anticipation of the template removal, can you keep this as 

```suggestion
    CORRADE_COMPARE_AS(Containers::arrayCast<Int>(DebugTools::bufferData(a)),
```

and same elsewhere?

src/Magnum/GL/Test/BufferImageGLTest.cpp
```diff
@@ -33,6 +33,10 @@
 #include "Magnum/GL/PixelFormat.h"
 #include "Magnum/GL/OpenGLTester.h"
 
+#ifndef MAGNUM_TARGET_GLES2
+#include "Magnum/DebugTools/BufferData.h"
```
mosra, line 37: Ensure this library is also linked in everywhere except ES2 -- which is the actual root cause for the TODO you have ;)

> On some CIs DebugTools/BufferData.cpp isn't being compiled, leading to missing symbols. Investigate.

Currently because it's inlined on WebGL and desktop, you don't have a linker failure there, but with the anticipated template removal it'll be deinlined everywhere, so just link it everywhere except ES2.

Same for the other three tests.

mosra (commented):
Doesn't feel that much of a rabbit hole to me :)

> I didn't feel like adding all the checks for that

Sounds reasonable, ES2 is nowadays only important as a debugging-friendly alternative to WebGL 1 anyway. And it's not like WebGL 1 would be that heavily used either. Plus, I don't remember any suspicious ES2-specific behavior until now so it's unlikely that the tests would discover some new bug :)

src/Magnum/DebugTools/BufferData.h
pezcode: The intention was to unify `bufferData()` and `bufferSubData()`. It felt odd having to explicitly add `<char>` depending on whether I wanted part of or the entire buffer data. But I can drop that commit, just wasn't aware of your plans to remove templates.

src/Magnum/GL/Buffer.cpp
pezcode: 👀 Good catch

src/Magnum/DebugTools/BufferData.h
mosra: Ah, right. Okay, leave it in, then -- makes the usage closer to what the template-less version will be.

src/Magnum/GL/Buffer.cpp
pezcode: I was thinking... couldn't we fall back to using `EM_ASM` to call `getBufferSubData()` function on older Emscripten versions? There'd be some runtime overhead on those versions, but it's much better than not being able to read buffers at all. And it'd remove the need for ugly version checks around the two functions as well as in all tests.

pezcode:
The macos-gles3 failure seems to be a broken interaction with `MagnumDebugTools` and `MagnumOpenGLTesterTestLib`. I've read the comments about the latter, but it's not entirely clear to me why `MagnumDebugTools` breaks things.

src/Magnum/GL/Buffer.cpp
mosra: Realistically, is anybody using such old Emscripten anymore? (Except me on the CI, which I should upgrade anyway.)

If ancient Emscriptens would be regularly used by people because they're in LTS Linux distro repos (which is the case with old GCC and old CMake), then it would make sense to add some fallback path, but I don't think anybody really does that.

I'd say -- let's upgrade the WebGL 2 build on the CI to 2.0.17 to not have to ifdef each and every test (can you do that in the PR?) but keep the API in `GL::Buffer` ifdef'd to not cause a linker failure for *users* on older versions. I don't expect regular users to build tests as well, so this should be fine. Then support for < 2.0.17 is just on a best-effort basis (we aim to support but we don't explicitly test for it). To keep at least some variety in the testing, the WebGL 1 CI build would stay on 1.39 for now and the next upgrade round (in a year or so?) would bump WebGL 2 build to 3.0.x and WebGL 1 to 2.0.x.

Hm, now I remember we still have the compatibility code for pre-1.39 event handling behavior in the Application classes, right? Probably a good time to drop that now :)

src/Magnum/GL/Test/CMakeLists.txt
```diff
@@ -162,10 +170,10 @@ if(BUILD_GL_TESTS)
     endif()
 
     if(NOT MAGNUM_TARGET_GLES2)
-        corrade_add_test(GLBufferImageGLTest BufferImageGLTest.cpp LIBRARIES MagnumOpenGLTesterTestLib)
+        corrade_add_test(GLBufferImageGLTest BufferImageGLTest.cpp LIBRARIES MagnumDebugTools MagnumOpenGLTesterTestLib)
```
mosra, line 173: To fix the Mac CI failure, can you try swapping the order?

```suggestion
        corrade_add_test(GLBufferImageGLTest BufferImageGLTest.cpp LIBRARIES MagnumOpenGLTesterTestLib MagnumDebugTools)
```

Same below. The `*TestLib` has the GL library built with graceful assertions, while the regular `DebugTools` links to a regular `MagnumGL`. Meaning, there's two duplicate candidates for the (global) GL symbols and the linker picks the first it finds (from `DebugTools`), but then the code in `MagnumOpenGLTesterTestLib` relies on the second duplicate, causing an assertion due to a different copy of the global used in different parts of the code.

I hope that's enough to make this work, if not then I'd probably need to introduce `MagnumDebugToolsTestLib` that relies on `MagnumGLTestLib` instead of `MagnumGL`.

src/Magnum/GL/Buffer.cpp
pezcode: I upgraded to 2.0.17 for the emscripten-webgl2 job but it's failing with very funny errors. Not sure if we actually hit a compiler error there, but I'll try to build with that exact version locally.

src/Magnum/GL/Test/CMakeLists.txt
pezcode: Ah, thanks, I understand the issue now. Sadly changing the order doesn't seem to make a difference 😢 I assume CMake sorts the link order anyway

mosra:
Okay, seems we're at the cursed issues again. I'll take it from here :D

mosra:
Ah yes, 2.0.18 doesn't die anymore, but I started hitting #507 now. Curses!

Investigating...

mosra:
Ah well, and an attempted fix for the macOS global symbol mismatch, where I made `MagnumDebugToolsTestLib` depend on `MagnumGLTestLib` causes an ASan complaint due to duplicate symbols now appearing in different places. I suppose I just can't use `DebugTools::bufferData()` in the GL library tests, then...

pezcode:
Oh dear, what have I done 🙊 sorry about this mess

mosra:
All "good" for Emscripten now (see [#507](https://github.com/mosra/magnum/issues/507#issuecomment-1126975690) for details).

For the other problem, I managed to "fix" it so now it's not macOS or ASan complaining, but Linux tests failing due to asserts not being graceful anymore. [I suppose this is the culprit, gotta fix that properly I guess.](https://github.com/mosra/magnum/blob/0508e2932028d9c01118c074f5c5e771c11b3a81/src/Magnum/GL/CMakeLists.txt#L324-L332)

mosra:
Hm, reshuffled the library order once more and now it seems to work. Although it's a tower of brittle hacks and wishful thinking, hopefully not breaking again anytime soon.

Will merge once I implement the template-removal TODO as well. Tomorrow, enough suffering for a Sunday :skull:

mosra:
Finally merged a subset of this as 500e41e4a87ae4cecb08afb089e5aea3fbcd887d...4350f1358ac1b6c6d806f3af8bb055923a2368c3, sorry for the extremely long delay. CircleCI macOS builds started *also* acting up in the meanwhile, further delaying everything.

I ended up deprecating the templated `bufferData()` API in 0113b34852a42ce2a507dff6c8cd3dec11ccdf3d and then rebased your commits on top of that, as that order made more sense (and allowed me to reduce the deprecated code size quite a bit). Using DebugTools inside GL tests however proved to be too cursed, so I split that off into #565 for now. Will get back to fixing that later.

Thanks for your work!

PR #563, opened 2022-05-16
Fix tiny english issue in features.dox

cmey:
from most part --> for the most part


mosra:
Little things that matter. Thank you! :+1:

Issue #486, opened 2020-12-10
Plugin import breaks if CMAKE_RUNTIME_OUTPUT_DIRECTORY_* is set


Gollum999:
### Environment
* Windows 10
* Visual Studio 16.8.2 (MSVC 19.28.29334)
* CMake 3.18.1
* Magnum mosra/magnum@92d0609e33252c4e1f9e772dfb0825e96122aba3
* Corrade mosra/corrade@e5970c4190717a6425f1b17241f19a62341569ef

### Problem
If `CMAKE_RUNTIME_OUTPUT_DIRECTORY_<CONFIG>` is set, trying to load plugins will fail.  Example error message:

```
PluginManager::Manager::Manager(): none of the plugin search paths in {C:/code/magnum/bld/vs/Debug/bin/magnum-d/importers, magnum-d/importers} exists and pluginDirectory was not set, skipping plugin discovery
PluginManager::Manager::load(): plugin TgaImporter is not static and was not found in
```
As described [here](https://cmake.org/cmake/help/latest/variable/CMAKE_RUNTIME_OUTPUT_DIRECTORY_CONFIG.html#cmake-runtime-output-directory-config), if this variable is set, it is used to initialize the [`RUNTIME_OUTPUT_DIRECTORY_<CONFIG>`](https://cmake.org/cmake/help/latest/prop_tgt/RUNTIME_OUTPUT_DIRECTORY_CONFIG.html#runtime-output-directory-config) property of each target.  This property takes precedence over the [`RUNITME_OUPUT_DIRECTORY`](https://cmake.org/cmake/help/latest/prop_tgt/RUNTIME_OUTPUT_DIRECTORY.html#runtime-output-directory) property.  This means that the output directories that are set in each plugin's [CMakeLists](https://github.com/mosra/magnum/blob/92d0609e33252c4e1f9e772dfb0825e96122aba3/src/MagnumPlugins/TgaImporter/CMakeLists.txt#L50) end up being overridden.  This causes the plugin library to not be built into the expected location, causing it to fail to load at runtime.

### To reproduce
Build Magnum as a subproject as in [Example 1a](https://doc.magnum.graphics/magnum/getting-started.html#getting-started-setup-subproject) of the Getting Started guide.  Add the following lines to CMake before `add_subdirectory(magnum)`:

```
set(CMAKE_RUNTIME_OUTPUT_DIRECTORY_DEBUG ${CMAKE_RUNTIME_OUTPUT_DIRECTORY})
[...]

PR #570, opened 2022-06-13
CMake: Set plugins runtime output directory even when the rest are not set

hugoam:
Hello !
Our project only sets `CMAKE_RUNTIME_OUTPUT_DIRECTORY` but not the other two CMake variables, because those tend to mess with the way Visual Studio produces and consumes `.lib` so we don't want to open that can of worms.

This wasn't an issue before as we only consumed Magnum in a prebuilt way, but now I'm including magnum as part of our CMake tree for development convenience, and I'd still like the `.dll` to end up in the correct plugin subfolder without having to do some ugly stuff.

So this change proposes to set the `RUNTIME_OUTPUT_DIRECTORY` of the plugins when only `CMAKE_RUNTIME_OUTPUT_DIRECTORY` is set even if the other two variables are not set. In order to facilitate the change, I also factored the directory selection logic of the plugins to a single place since it was relatively complex code that was copied in more than 20 locations.

Let me know what you think and if you see issues with this approach, thanks :)

### DIFF
--- a/src/MagnumPlugins/CMakeLists.txt
+++ b/src/MagnumPlugins/CMakeLists.txt
@@ -33,17 +33,20 @@ macro(add_plugin plugin_name plugin_folder debug_install_dirs release_install_di
         corrade_add_static_plugin(${plugin_name} "${release_install_dirs}" ${metadata_file} ${ARGN})
         set_target_properties(${plugin_name} ${plugin_name}-dependencies PROPERTIES FOLDER "MagnumPlugins/${plugin_name}")
     endif()
-    # Modify only runtime output location if that one is set
-    if(CMAKE_RUNTIME_OUTPUT_DIRECTORY)
-        set_target_properties(${plugin_name} PROPERTIES
-            RUNTIME_OUTPUT_DIRECTORY ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/magnum$<$<CONFIG:Debug>:-d>/${plugin_folder})
-    endif()
-    # Modify the other output locations if all are set
-    if(CMAKE_RUNTIME_OUTPUT_DIRECTORY AND CMAKE_LIBRARY_OUTPUT_DIRECTORY AND CMAKE_ARCHIVE_OUTPUT_DIRECTORY)
-        set_target_properties(${plugin_name} PROPERTIES
-            LIBRARY_OUTPUT_DIRECTORY ${CMAKE_LIBRARY_OUTPUT_DIRECTORY}/magnum$<$<CONFIG:Debug>:-d>/${plugin_folder}
-            ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_ARCHIVE_OUTPUT_DIRECTORY}/magnum$<$<CONFIG:Debug>:-d>/${plugin_folder})
-    endif()
+    # For each type of output directory and each configuration, append the plugin subfolder only if the global CMAKE_ variable was set
+    foreach(output RUNTIME LIBRARY ARCHIVE)
+        if(CMAKE_${output}_OUTPUT_DIRECTORY)
+            set_target_properties(${plugin_name} PROPERTIES
+                ${output}_OUTPUT_DIRECTORY ${CMAKE_${output}_OUTPUT_DIRECTORY}/magnum$<$<CONFIG:Debug>:-d>/${plugin_folder})
+        endif()
+        # This doesn't handle custom build types as there's no way to know their names
+        foreach(config DEBUG RELEASE RELWITHDEBINFO MINSIZEREL)
+            if(CMAKE_${output}_OUTPUT_DIRECTORY_${config})
+                set_target_properties(${plugin_name} PROPERTIES
+                    ${output}_OUTPUT_DIRECTORY_${config} ${CMAKE_${output}_OUTPUT_DIRECTORY_${config}}/magnum$<$<CONFIG:Debug>:-d>/${plugin_folder})
+            endif()
+        endforeach()
+    endforeach()
 endmacro()
 
 if(MAGNUM_WITH_ANYAUDIOIMPORTER)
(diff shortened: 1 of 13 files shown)


mosra:
I saw you forking the repos so I went and saw this change already. Especially the massive deletion I like :D

Before I go any further, is #486 something you could fix as part of this PR as well? Should be hopefully easier now that it's not duplicated all over the place.

hugoam:
> Before I go any further, is #486 something you could fix as part of this PR as well? Should be hopefully easier now that it's not duplicated all over the place.

I'm thinking about it, but I'm unsure what would be the proper way to fix it. One way to see the problem is that it actually might be Conan's responsibility to set the `CMAKE_RUNTIME_OUTPUT_DIRECTORY` variable to a generator expression that discards the configuration, instead of setting `CMAKE_RUNTIME_OUTPUT_DIRECTORY_DEBUG` and `CMAKE_RUNTIME_OUTPUT_DIRECTORY_RELEASE` directly

pezcode:
> Maybe @pezcode could help here, for him it worked always... somehow.

I don't recall ever having to change a CircleCI configuration somewhere, but I do have an account there that's linked to my GitHub account if that makes any difference

hugoam:
Hmmm, CircleCI works now. I *think* what made it work is that `Allow Uncertified Orbs` was not checked in my CircleCI account settings, and Magnum needs this to be enabled. This is pretty silly though especially that there is no way to know that this is the cause. I just discovered it because I tried to enable CircleCI on my own Magnum fork, and the job failed because of this setting

src/MagnumPlugins/CMakeLists.txt
```diff
@@ -33,6 +33,21 @@ macro(add_plugin plugin_name debug_install_dirs release_install_dirs metadata_fi
...
+            set_target_properties(${plugin_name} PROPERTIES
+                ${OUTPUT_VAR} ${CMAKE_${OUTPUT_VAR}}/magnum$<$<CONFIG:Debug>:-d>/${plugin_folder})
+        endif()
+        foreach(CONFIG ${CONFIGS})
+            if(CMAKE_${OUTPUT_VAR}_${CONFIG})
+                set_target_properties(${plugin_name} PROPERTIES
+                    ${OUTPUT_VAR}_${CONFIG} ${CMAKE_${OUTPUT_VAR}_${CONFIG}}/magnum$<$<CONFIG:Debug>:-d>/${plugin_folder})
+            endif()
+        endforeach()
+    endforeach()
```
mosra, line 50:
```suggestion
    foreach(output RUNTIME LIBRARY ARCHIVE)
        if(CMAKE_${output}_OUTPUT_DIRECTORY)
            set_target_properties(${plugin_name} PROPERTIES
                ${output}_OUTPUT_DIRECTORY ${CMAKE_${output}_OUTPUT_DIRECTORY}/magnum$<$<CONFIG:Debug>:-d>/${plugin_folder})
        endif()
        # This doesn't handle custom build types as there's no way to know their names
        foreach(config DEBUG RELEASE RELWITHDEBINFO MINSIZEREL)
            if(CMAKE_${output}_OUTPUT_DIRECTORY_${config})
                set_target_properties(${plugin_name} PROPERTIES
                    ${output}_OUTPUT_DIRECTORY_${config} ${CMAKE_${output}_OUTPUT_DIRECTORY_${config}}/magnum$<$<CONFIG:Debug>:-d>/${plugin_folder})
            endif()
        endforeach()
    endforeach()
```

I moved the loop items directly to the `foreach()` calls to avoid polluting the scope with extra variables, and left the `_OUTPUT_DIRECTORY` part directly in the property call to make it clearer what's being modified.

Compared to the previous variant you had the difference is that before it was overriding the output only if either `CMAKE_RUNTIME_OUTPUT_DIRECTORY` was set, or if all three were set, now it treats each output directory setting individually and so covers more cases. I'm trying to imagine a case where this would be problematic, but I don't see any, so it's probably fine this way.

What bugs me a bit is the explicit list of configs, but no matter how hard I look, there doesn't seem to be a builtin way to ask CMake for a list of these. There's `CMAKE_CONFIGURATION_TYPES`, but that's just for multi-config generators (VS, Xcode) and not Ninja or Make, and not even the [official docs](https://cmake.org/cmake/help/latest/manual/cmake-buildsystem.7.html#default-and-custom-configurations) recommend anything about listing the new configs anywhere. I thought about maybe extracting the list from the `CMAKE_BUILD_TYPE` cache entry doc string, but that's brittle and it's unlikely that a project would update the doc string when adding custom build types. So not really a solution.

mosra (commented):
Yay, glad that you got CircleCI working. Sorry for the hassle, fortunately you need to do that only for the first PR. Just to be sure I looked into the repo config and "Build forked PRs" is checked, so it should theoretically work. No idea why it didn't.

Thanks for accounting for the config-specific locations. 

Just a single suggestion and then it can go :)

src/MagnumPlugins/CMakeLists.txt
hugoam: Incorporated your change :)

hugoam:
Incorporated your change and in the magnum-importer PR too
Most likely once this is merged we can close https://github.com/mosra/magnum/issues/486 too 😄

mosra:
Thanks! Yup, #486 would be solved by this.

One last *last* thing, just to make sure -- I see a `Fix Build NO_DEPRECATED` commit here, it disappeared before but now it's back. Is it still needed or are the changes I made in f049ba9012bd2f9291cba0cb01238f73c272941c enough to make the non-deprecated build work on MSVC? If it's not needed anymore, I'll skip it during merge.

mosra:
Ah so f049ba9012bd2f9291cba0cb01238f73c272941c didn't fix everything? If you get a chance, post the full error here and I'll attempt further.

Will merge this in a moment, together with the plugins PR.

mosra:
Merged as bcb96eba36c02afedaf860c26c205985a10226bc, thank you! 

And 45c1a7091f6386daf2b1075c9fb05d508d94f2b4 is in `master` now as well, so let me know what's left to make non-deprecated builds finally work :)

Issue #534, opened 2021-09-15
Async GL shader compilation


mosra:
The goal is figuring out a way how to use [`KHR_parallel_shader_compile`](https://www.khronos.org/registry/OpenGL/extensions/KHR/KHR_parallel_shader_compile.txt) without having to give up on the current way shaders are compiled, linked and initialized in a single constructor expression:

```cpp
Shaders::FooGL shader{Shaders::FooGL::UniformBuffers, 5, 3, 128};
```

Right now, shader compiler parallelization is only partially exploited in list-taking `GL::Shader::compile()` and `GL::AbstractShaderProgram::link()` which first submit compile/link command for all shaders and then check for errors. Since those are always called from the constructor, the max amount of parallelization is only 2 (or 3 when GS is used), which really isn't much. One option to improve this would be to call `compile()` and `link()` outside of the constructor, on multiple shader/program instances, but that means we lose the convenient single-expression RAII setup, and because the status is still checked right after all compilation jobs is submitted it won't help much on certain backends such as the _shitty_ old D3D compiler that can sometimes take several seconds to compile a single shader.

The requirements are thus:

-   keep RAII constructor initialization, because otherwise all shader interfaces would have to do additional checks for completeness
-   avoid having to implement two separate code paths for direct and async shader compilation, because that'd soon become a maintenance nightmare
-   remove existing limitations to the parallelism like implicit error checks or an inability to parallelize more than 2 or 3 compilations at the same time
-   don't complicate the existing shader creation APIs any further, as such simplicity is still desirable for tools and quick demos where startup time isn't the topmost priority

A proposal that could check out all of the above is:

- [ ] Add `void Shader::submitCompile()` and `void AbstractShaderProgram::submitLink()` (better naming?) that only submit the compilation/link and don't check for the status
- [ ] Add `bool Shader::checkCompile()` and `bool AbstractShaderProgram::checkLink()` (naming??) that returns `false` and prints an error message 
    - [ ] or maybe returns it in a string instead?
    - [ ] do we actually need to have a way to check compilation status at all, or only when link fails? the [WebGL extension spec](https://www.khronos.org/registry/webgl/extensions/KHR_parallel_shader_compile/) says `// Never check compile status unless subsequent linking fails.` so maybe the desired workflow is 
        1. check link status
        2. if it fails, check compile status
        3. if compile failed, print compile status message
        4. if compile didn't fail, print link status message
[...]

PR #576, opened 2022-07-08
Async GL shader compilation

dranikpg:
#534


dranikpg:
This is still a WIP, so no worries about incomplete doc comments, undefined constants for GLES targets, formatting, etc. :)

While implementing the proposed changes for Shader&Program is no more than splitting code into two functions, handling compile states requires some planning.

Compilation states, I suppose, should be only data bags. All logic should be in `FooGL::compile` or the final constructor. This way is is much cleaner and easier to manage. However `AbstractShaderProgram` constrainedness stands in the way. A shader program has to be created and submitted for linkage in `FooGL::compile` or the compilation state. Besides that, typical `FooGL::compile` requires a program to exist early on to call `bindAttributeLocation`/`bindFragmentDataLocation`, which have to be arranged strictly before linkage. So it'd make sense fully to reuse `AbstractShaderProgram`, which supports all operations.

I came up with the following solution and implemented it for FlatGL (and it works): 
* I assumed that `AbstractShaderProgram` (ASP) should remain abstract with protected members. After all, that was its intention. But to reuse its functionality, we will make `ASP::CompileState` extend it. 
* `FlatGL::ComplieState` will extend `ASP::CompileState` as planned. 
* `FlatGL::ComplieState` will declare `FlatGL::compile` and `FlatGL(CompileState&&)` as friend, to allow access to ASP functions. 
* `FlatGL(CompileState&&)` will simply move the `ASP` out of the compile state, to initialize its base class (which is another `ASP`)
* All common public faced helper functions (like `isLinkFinished`) will be in `ASP::CS`

```       
                          -- submit/check/... 
                          |
                          v
                       +++++++++++++   
                       +    ASP    +  <-------- moves out of
                       +++++++++++++          | to init its ASP
                            ||                |
                       +++++++++++++          |
FooGL::compile         +  ASP::CS  +       FooGL(FooGL::CS&&)
       |               +++++++++++++          |
       |                    ||                |
       |               +++++++++++++          |
       ------------->  + FooGL::CS +  <-------- friend to 
creates & returns      +++++++++++++            use ASP functions 
friend
```

This solution is compact and doesn't change the way `ASP` works, however it feels a bit tricky to me...

Things to consider when generalizing for more than FlatGL:
* Instead of using friend, we can reimplement all required functions for `ASP::CS`. More boilerplate, but less tricky?
* We can make `ASP::CS` store vertex&fragment shaders and call attach in constructor. However submitLink still has to be called manually after all bind calls.
* Instead of calling bind & submitLink manually, we can pass a function/templatize by a functor, which will be called after creation, but before linkage. This way `FooGL::compile` wont contain any `ASP` specific logic

What are your thoughts on this?

mosra:
Hi, apologies for a delayed response here -- I'm stuck deep in a certain other feature and its taking way more time than expected. 

Thanks for the code, and especially for the detailed description of your thought process. Very appreciated :+1: 

I'll post a more detailed reply once I manage ship that other feature (hopefully later this week), but after a brief look at your initial changes it seems my original propsal wasn't exactly great in terms of the amount of boilerplate inflicted on subclasses. (My fault, sorry.) I have to think more about that. The ideal state would be that subclasses wouldn't need to implement that much extra apart from what they did before (because all duplicated state would also mean twice as much testing effort), only splitting the initial setup somewhat, perhaps at the cost of some safety in case the async compilation is used. Which is fine I guess, it's an advanced feature after all.

I have a vague idea that the construction could get split via a chain of delegating constructors, which would either directly call into each other in the straightforward case, or an object gets "partially constructed", waited upon, and then moved to a fully-constructed state once the compilation finishes. I'll outline that better once I get my hands free. 

Until then, in 74f1778759d226fc3464f6adb045f370f4c14206 I added support for `KHR_parallel_shader_compile` at least, so you don't have to patch it in by hand. (It caused a conflict -- sorry.)

dranikpg:
Throwing it in as as suggestion before you get back to it.
Instead of having separate compile states like described in the issue, we might be just fine with one generic one that immediately constructs the ASP, but guards it unitl its not compiled.
```c++
template<typename FooGL>
struct CompileState {
private: // no access to shader until compiled
  FooGL asp; // moved in from FooGL::compile
  Shader vert, frag;
  optional<Shader> geom; // or parameterize by N - number of shaders?
}
``` 
then 
```c++
FooGL(CompileState<FooGL>&& cs)
```
blocks and finishes consturction.
All common functions can be handled by the compile state directly. Further more, there will be no duplicated fields like in the previous case

mosra:
Okay, that actually seems like the best middle ground :) 

The delegating constructors from my comment above wouldn't work on its own because there would be nowhere to store the temporary `GL::Shader` instances, while OTOH putting them into the shader program itself would be wasteful. But combined with your suggestion it could work:

```cpp
class FlatGL::CompileState: public FlatGL {
    private:
        friend FlatGL;
        using FlatGL::FlatGL; // low-effort way to get access to the NoCreate constructor to populate this
        GL::Shader frag, vert;
        GL::Version version; // + whatever else the constructor needs to preserve to avoid duplicated code
};
``` 

It's publicly derived, so the getter APIs like `isLinkFinished()` or `flags()` are easy to access if user code needs to. (Which means the dangerous APIs like `draw()` are accessible too, but that's fine, the same danger would be with a `NoCreate`'d instance.) I'm not 100% sure if it should be an inner class or defined externally as `FlatGLCompileState`, but having it as an inner class avoids further polluting [docs of the Shaders namespace](https://doc.magnum.graphics/magnum/namespaceMagnum_1_1Shaders.html#nested-classes) at least. It's also not a template because I think it's simpler to just type out all extra state you'd need (for compute shaders there would need to be `Optional<GL::Shader> compute`, etc, getting very ugly very fast).

Then, there would be `FlatGL::compile()` returning `FlatGL::CompileState`, and `FlatGL(CompileState&&)` that finishes construction, like you said, and like you already mostly have in your code.

Additionally:

- I would make `isLinkFinished()` public in `AbstractShaderProgram`, to avoid a need for an `AbstractShaderProgram::CompileState` -- for a finalized instance it would always return `true` which is useless but not dangerous, so it doesn't need to be hidden from the public API.
- I would make the `submitCompile()` / `submitLink()`, `checkCompile()` / `checkLink()` APIs operate just on a single shader/program, instead of a list. The list was there originally as a non-ideal parallelization attempt, and since you have a much better alternative now, calling `submitCompile({a, b})` doesn't really have any advantage over `a.submitCompile(), b.submitCompile()` (besides avoiding some temporary allocations, but that's my problem to solve). The original list-taking `compile()` / `link()` would stay (and sequentially call into the submit/check APIs) for backwards compatibility for 3rd party code.

Does this make sense? I hope I didn't forget something critical again :) And apologies for the delay.

dranikpg:
Makes sense and should work 😄 I'll try this out next week

dranikpg:
I _sketched_ the proposed changes. If it looks acceptable, I'll move on to tidying and implementing the same for other shaders.

`FlatGL` needs an additional constructor that will be called from its compile state. I suggest 
```c++
protected:
/** @brief Construct without running shader compilation and linking */
FlatGL(NoInitT, Flags flags, etc.)
```
that will create the ASP and set variables, but not trigger compilation (or else we're stuck in a loop 😆). `NoInit` seems to be the closest from the available tags.

--- 

> besides avoiding some temporary allocations, but that's my problem to solve

I saw the containers and it made me not go for the split initially. There is a todo to implement stack based allocation, I can try this out in this PR.

mosra (commented):
Amazing, thank you. I think the scaffolding essentials are quite close to completion, so I commented more thoroughly -- hope that's okay :)

What's left (apart from eventually stamping this out for the other shader variants once we have the base finalized) is some testing. I imagine this could be sufficient:

- In `GL/Test/ShaderGLTest.cpp` there's a `compile()` test case. I'd split it into `compile()`, `compileFailure()`, `compileAsync()` and `compileAsyncFailure()`, verifying that `isCompileFinished()` is always `true` after `compile()` or `checkCompile()`, and that the failure message is never printed after `submitCompile()` but only after `checkCompile()`. Given there would be the suggested fallback for drivers without `KHR_parallel_shader_compile`, the test wouldn't really need to behave differently depending on extension presence. On the other hand, drivers can expose the extension but not actually parallelize anything, so I don't see a way to verify if `isCompileFinished()` actually ever returns `false`. Though if you have any idea, feel free to try that out.
- Similarly in `AbstractShaderProgramGLTest.cpp` for the linking case.
- For the flat shader, there's `compile()` and `compileUniformBuffers()` in `Shaders/Test/FlatGLTest.cpp`. Those are instanced and test a ton of combinations. I don't think it's needed to test all that again, I'd only add `compileAsync()` and `compileUniformBuffersAsync()` that check a single case with non-trivial input arguments, use the `CompileState` workflow there, verify that the flags/material/draw count gets properly propagated right from the start and verify that `isLinkFinished()` returns `true` after.

Tests can be enabled with `MAGNUM_BUILD_TESTS` and `MAGNUM_BUILD_GL_TESTS` in CMake, then for easy execution they're all put into `Debug/bin` or `Release/bin`. I'm pretty sure you'll manage to get around, but just in case [here's a small tutorial for the test suite](https://doc.magnum.graphics/corrade/testsuite.html) and [here's how debug/warning/error output can be captured and verified](https://github.com/mosra/magnum/blob/e8e9aeec7faee941620a39ee016cdd0df880ddfc/src/Magnum/GL/Test/AbstractShaderProgramGLTest.cpp#L454-L460). You could use for example [Compare::StringHasPrefix](https://doc.magnum.graphics/corrade/classCorrade_1_1TestSuite_1_1Compare_1_1StringHasPrefix.html) to check that the output is reasonable without having to worry about random difference in the output across platform.

When you push, the GLES2 and GLES3 CI jobs as well as the Android GLES2 job run the GL tests with SwiftShader. I'm not sure if SwiftShader implements this extension (probably not?), but in any case most of the code gets verified. The other jobs don't run any GL tests, at the very least they'll check that the code compiles.

src/Magnum/Shaders/FlatGL.cpp
```diff
@@ -277,22 +289,44 @@ template<UnsignedInt dimensions> FlatGL<dimensions>::FlatGL(const Flags flags
...
+    #ifndef MAGNUM_TARGET_GLES2
+    , UnsignedInt materialCount, UnsignedInt drawCount
+    #endif
+):
+    #ifndef MAGNUM_TARGET_GLES2
+    FlatGL(compile(flags, materialCount, drawCount))
+    #else
+    FlatGL{compile(flags)}
+    #endif
+{}
```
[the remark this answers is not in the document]
dranikpg: This corresponds to two separate constructors in the header. Below is another one, that forwards (Flags) to (Flags, Int, Int). You might have placed them here on purpose to not clutter the header, but I can move them both if needed.

src/Magnum/GL/Shader.cpp
```diff
@@ -749,74 +749,71 @@ Shader& Shader::addFile(const std::string& filename) {
...
+    std::string message(logLength, '\0');
+    if(message.size() > 1)
+        glGetShaderInfoLog(_id, message.size(), nullptr, &message[0]);
+    message.resize(Math::max(logLength, 1)-1);
+
+    /* Some drivers are chatty and can't keep shut when there's nothing to
+       be said, handle that as well. */
+    Context::current().state().shader.cleanLogImplementation(message);
+
+    /** @todo update log messages */
```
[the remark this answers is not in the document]
dranikpg: > Can you elaborate on what this TODO is for?

Just to remember to ask you, what to do if we no longer can print which shader failed compilation 😅 No intentions of staying in code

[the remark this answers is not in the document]
mosra: Yeah I think I answered that above -- it's fine, it says the kind of the shader that failed (vertex/fragment/...), which should be enough to distinguish as programs always have at most one of each kind.

src/Magnum/Shaders/FlatGL.cpp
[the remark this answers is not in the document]
mosra: Ah, excellent point :+1: Yep, it would probably make sense to move both, since both are now just thin wrappers that eventually delegate to `compile()`.

(Just to be clear, inline them directly inside the class definition. But I guess you know what I meant ;))

src/Magnum/Shaders/FlatGL.h
```diff
@@ -1038,6 +1052,22 @@ template<UnsignedInt dimensions> class MAGNUM_SHADERS_EXPORT FlatGL: public GL::
         #endif
 };
 
+template<UnsignedInt dimensions> class FlatGL<dimensions>::CompileState : public FlatGL<dimensions> {
+public:
+    using FlatGL::isLinkFinished;
```
mosra, line 1057: Is this `using` even needed?

src/Magnum/Shaders/FlatGL.h
```diff
@@ -602,6 +603,16 @@ template<UnsignedInt dimensions> class MAGNUM_SHADERS_EXPORT FlatGL: public GL::
...
 
+        class CompileState;
+
+        explicit FlatGL(CompileState&& cs);
+
+        static CompileState compile(Flags flags
+        #ifndef MAGNUM_TARGET_GLES2
+        , UnsignedInt materialCount, UnsignedInt drawCount
+        #endif
+        );
```
mosra, line 614: I think a convenience 

```cpp
        #ifndef MAGNUM_TARGET_GLES2
        static CompileState compile(Flags flags) {
            return compile(flags, 1, 1);
        }
        #endif
```

for non-UBO use cases could make sense as well, to match the similar convenience constructor.

src/Magnum/GL/AbstractShaderProgram.h
```diff
@@ -1446,13 +1447,36 @@ class MAGNUM_GL_EXPORT AbstractShaderProgram: public AbstractObject {
...
+         * @ref link(std::initializer_list<Containers::Reference<AbstractShaderProgram>>)
          * for improved performance, see its documentation for more
          * information.
          */
         bool link();
 
+        /**
+         * @brief Submit for linking
+         *
+         * The attached shaders must be compiled with @ref Shader::compile() before linking.
```
mosra, line 1461: I get the sentence copied from the `link()` above, but now it should probably mention that `submitCompile()` is also enough? :)

mosra (commented):
Just minor things, and I'm tired so apologies if my sentences don't make sense :D

I'll give it another look when you get to the tests. Thank you!

src/Magnum/Shaders/FlatGL.h
dranikpg: I removed `using FlatGL::FlatGL` because we now construct the `FlatGL` instance separately and call `attachShaders`, `submitLink`, etc directly on it. The only public function that is really needed, is `isLinkFinished`, I guess.

dranikpg: Only if we extend `FlatGL` by `private`/`protected`. This would allow us to hide all redundant FlatGL/ASP functions, as `isLinkFinished` is the only function needed in the CompileState, I guess. Is this a viable idea?

mosra: It would be useful to have access to `flags()`, `materialCount()` etc., even if just for testing/diagnostic purposes (for example, if a certain async compilation takes *really long*, you could use these to discover it's due to the driver taking a lot of time to unroll a large loop).

Which means, you'd have to add `using FlatGL::flags;` etc., and then I feel the maintenance overhead is not worth it. Just keep everything public. If it ever becomes a footgun, we can reconsider, but right now I don't see a problem with having everything public :)

dranikpg:
As it turns out, querying `GL_LINK_STATUS` might return success before the completion query does so 😕 I've discovered this when I was updating the tests

This means that the following is possible:
```c++
FlatGL2D s{{}, 1,1}; // calls checkLink internally
s.isLinkFinished(); // returns false
```

I've made a [small example](https://gist.github.com/dranikpg/a49c4d4d29ba2ef239e77fd916e7da21) that works in this odd way on my pc:

```
Renderer: AMD RENOIR (DRM 3.41.0, 5.13.0-52-generic, LLVM 12.0.1) by AMD
OpenGL version: 4.6 (Core Profile) Mesa 21.2.6
```

I'm not sure if this is the norm, because I couldn't find anything in the khronos docs on whether it can't be used after linking.

src/Magnum/GL/Test/ShaderGLTest.cpp
```diff
@@ -40,6 +42,18 @@
...
+#ifndef MAGNUM_TARGET_GLES
+    #ifndef CORRADE_TARGET_APPLE
+    Version::GL210
+    #else
+    Version::GL310
+    #endif
+    ;
+#else
+Version::GLES200;
+#endif
```
dranikpg, line 55: Should I just move this back into each test case? Because code re-use is pretty minimal across what I've seen in the tests

dranikpg:
> I'll give it another look when you get to the tests

@mosra just a friendly ping in case you didn't see my message above. No hurries if you're busy 😅 😓  I implemented the proposed test changes. If they look ok, I'll move to the other shaders. However, the is a small issue with the query correctness. I've described it above

mosra:
I saw the notifications but didn't find time yet, sorry for the wait again. I'll try to check this later today or tomorrow.

src/Magnum/GL/Test/ShaderGLTest.cpp
mosra: Yes, please, that makes them less prone to breakages if certain cases would need a different version. The tests are mostly write-only anyway, so the duplication isn't that much of a problem.

src/Magnum/GL/Test/AbstractShaderProgramGLTest.cpp
```diff
@@ -257,6 +265,74 @@ void AbstractShaderProgramGLTest::create() {
 
     MAGNUM_VERIFY_NO_GL_ERROR();
     CORRADE_VERIFY(linked);
+    // TODO: decide on this.
+    //CORRADE_VERIFY(program.isLinkFinished());
```
mosra, line 269: To answer your question -- does this return `false` always, once `checkLink()` was called? Or is it random or happening only on some drivers? If you wait half a second ([Utility::System::sleep()](https://doc.magnum.graphics/corrade/namespaceCorrade_1_1Utility_1_1System.html#a7f93179c285d2667b97d9b38476d4cda)), does it eventually become `true`? :D 

If it's random or inconsistent between drivers, then I would do this to have the behavior documented in the test (update the message to say what's actually happening):

```cpp
{
    bool finished = program.isLinkFinished();
    CORRADE_EXPECT_FAIL_IF(!finished, "Some drivers return false GL_COMPLETION_STATUS after the link status has been queried.");
    CORRADE_VERIFY(finished);
} 
```

If it *eventually* becomes `true` after, I'd just add a sleep, if it's not several seconds.

src/Magnum/Shaders/Test/FlatGLTest.cpp
```diff
@@ -853,6 +860,37 @@ template<UnsignedInt dimensions> void FlatGLTest::construct() {
     MAGNUM_VERIFY_NO_GL_ERROR();
 }
 
+template<UnsignedInt dimensions> void FlatGLTest::constructAsync() {
+    setTestCaseTemplateName(Utility::format("{}", dimensions));
+
+    auto&& data = ConstructData[testCaseInstanceId()];
+    setTestCaseDescription(data.name);
```
mosra, line 867: No need to test all combinations here again (apologies if I haven't been clear on this). Just pick one case from `ConstructData` that has sufficiently non-trivial `flags` (but ideally doesn't need to check for any extensions, to reduce the checks) and replicate it in this test case.

Same for the uniform buffer variant below, there just pick something with non-trivial `materialCount` / `drawCount`.

src/Magnum/Shaders/Test/FlatGLTest.cpp
```diff
@@ -853,6 +860,37 @@ template<UnsignedInt dimensions> void FlatGLTest::construct() {
...
+    #ifndef MAGNUM_TARGET_GLES
+    if((data.flags & FlatGL2D::Flag::ObjectId) && !GL::Context::current().isExtensionSupported<GL::Extensions::EXT::gpu_shader4>())
+        CORRADE_SKIP(GL::Extensions::EXT::gpu_shader4::string() << "is not supported.");
+    if((data.flags & FlatGL2D::Flag::TextureArrays) && !GL::Context::current().isExtensionSupported<GL::Extensions::EXT::texture_array>())
+        CORRADE_SKIP(GL::Extensions::EXT::texture_array::string() << "is not supported.");
+    #endif
+
+    auto compileState = FlatGL<dimensions>::compile(data.flags);
+    CORRADE_COMPARE(compileState.flags(), data.flags);
+
```
mosra, line 878: I'd like to have the actual async compilation tested here -- to make sure there was no brainfart involved that makes it never finish, and to have this use case represented with some actually working code people can look at.

```cpp
while(!compileState.isLinkFinished())
    Utility::System::sleep(100);
    
FlatGL<dimensions> shader{std::move(compileState)};
```

That would also make the `CORRADE_VERIFY(shader.isLinkFinished());` unneeded.

Same for the uniform buffer variant below.

src/Magnum/GL/AbstractShaderProgram.h
```diff
@@ -1258,19 +1258,20 @@ class MAGNUM_GL_EXPORT AbstractShaderProgram: public AbstractObject {
         AbstractShaderProgram& dispatchCompute(const Vector3ui& workgroupCount);
         #endif
 
-    protected:
         /**
-         * @brief Link the shader
+         * @brief Non-blocking linking status check
+         *
+         */
+        bool isLinkFinished();
```
mosra, line 1265: Docs still missing here ;) Cross-link with the actual GL APIs being called:

```cpp
         * @see @fn_gl_keyword{GetProgram} with 
         *      @def_gl_extension{COMPLETION_STATUS,KHR,parallel_shader_compile}
```

And mention what the fallback does if the extension isn't supported, and it'll probably be also useful to note the issue with `COMPLETION_STATUS` being false in certain scases. 

Same for `isCompileFinished()`.

mosra (commented):
I'll be away until Monday, but I think that apart from these few comments this is ready to be replicated for other shaders.

Great work, thank you :+1:

mosra (commented):
Ah, nice, so the sleep was enough to fix that strange issue :)

It's just `DistanceFieldVector` left, right? Awesome. You also managed to match the library coding style quite well, if there will be any minor issues left I'll happily do that post-merge. And I can also write the overview docs / example snippet myself after -- don't want to put all the burden on your first contribution ;)

The `linux-nondeprecated` build is failing because it seems you're accidentally using a deprecated alias (`MeshVisualizer2D` instead of `MeshVisualizerGL2D`, the fomer is a relic of the past and not present on the non-deprecated build). Easiest way to discover all such cases is if you disable the `MAGNUM_BUILD_DEPRECATED` CMake option in your local build, then you get the same errors as on the CI.

dranikpg:
> Ah, nice, so the sleep was enough to fix that strange issue :)

Well, it does it for the tests 😅 The API still doesn't work "as expected", but i've left a warning in the doc comments

> if there will be any minor issues left I'll happily do that post-merge

I'll have a pass at the end and try to fix as much as I can. It's just that every time I forget how something is supposed to be aligned, I open a random file and look how its done there 😄  I know there is Corrades styleguide, but it doesn't cover cases like line-breaking a function with 5 arguments and 3 ifdefs

> The linux-nondeprecated build is failing

Just didn't get to it. Thanks for the build option, I have been fishing them out of the logs. I'm more worried about AppVeyour... It spits out like 2k lines of warning in other parts of magnum and then the linker fails. There seems to be something wrong with the templated shaders (because MeshVisualizerGL2D is not part of the error). Specifically, the `FooGL(Flags)` constructor seems to be unresolved (after being moved to the header).

I don't have a dev-ready windows pc right next to me to test on, so it'll be cool if you could tell me the cause if you now it right away. In case this requires a little investigation, you can leave it on me.

mosra:
> I'll have a pass at the end [...] line-breaking a function with 5 arguments and 3 ifdefs

:bow: thank you! For line breaking, it's comments on 79 cols (mainly just to make them easy to read, unless there's an excessively long identifier or expression that can't really be broken). For other stuff, having the whole signature as a single long line is fine. Preprocessor `#ifdef`s in signatures I indent with 4 extra spaces compared to surrounding code.

If I may, the only other thing I noticed is a space before `:`, I have it always without, be it a class base or a range-for. But again, don't worry too much, I can fix it up after. I can't stand the mess `clang-format` makes so there's no automated way and thus the formatting burden is on me.

> then the linker fails

It complains about a `vftable`, which is a vtable, which is due to `AbstractShaderProgram` having a virtual destructor. My theory is that, IIRC, the vtable usually gets attached to the first or last method declaration in the class, or something like that. So if the definition is inline and it's combined with the fact that it's a template, then it could happen that the vtable isn't stored in the DLL. Or something, heh. 

TL;DR: try to deinline the `explicit FlatGL(NoInitT) {}` (i.e., make a `= default`ed definition of it in the `.cpp` file), maybe that makes it go away? 

I don't have a Windows machine around either, so I usually just abuse AppVeyor, pushing random tries until it starts cooperating. Feel free to do the same, haha.

src/Magnum/Shaders/Test/DistanceFieldVectorGLTest.cpp
```diff
@@ -413,6 +455,59 @@ template<UnsignedInt dimensions> void DistanceFieldVectorGLTest::constructUnifor
...
+            CORRADE_SKIP(GL::Extensions::ANGLE::multi_draw::string() << "is not supported.");
+        #else
+        if(!GL::Context::current().isExtensionSupported<GL::Extensions::WEBGL::multi_draw>())
+            CORRADE_SKIP(GL::Extensions::WEBGL::multi_draw::string() << "is not supported.");
+        #endif
+    }
+    auto compileState = DistanceFieldVectorGL<dimensions>::compile(data.flags, data.materialCount, data.drawCount);
+    CORRADE_COMPARE(compileState.flags(), data.flags);
+    CORRADE_COMPARE(compileState.materialCount(), data.materialCount);
+    CORRADE_COMPARE(compileState.drawCount(), data.drawCount);
```
mosra, line 491: You could go even simpler here, the objective is just to verify the `CompileState` type gets all data fields populated correctly. I.e., pick something where you don't need to test for any extensions, and don't bother with `struct {...} data` either:

```suggestion
    auto compileState = DistanceFieldVectorGL<dimensions>::compile(DistanceFieldVectorGL2D::Flag::UniformBuffers|DistanceFieldVectorGL2D::Flag::TextureTransformation, 16, 48);
    CORRADE_COMPARE(compileState.flags(), DistanceFieldVectorGL2D::Flag::UniformBuffers|DistanceFieldVectorGL2D::Flag::TextureTransformation);
    CORRADE_COMPARE(compileState.materialCount(), 16);
    CORRADE_COMPARE(compileState.drawCount(), 48);
```

Similarly in other test cases.

dranikpg:
I guess we don't need full checks in constructor

```c++
FooGL(CompileState&&) {
    CORRADE_INTERNAL_ASSERT_OUTPUT(checkLink());
    CORRADE_INTERNAL_ASSERT_OUTPUT(cs._vert.checkCompile());
    CORRADE_INTERNAL_ASSERT_OUTPUT(cs._frag.checkCompile());
```

because in case the first check fail, the other two fill be skipped. In case the first is successful, the others are guaranteed to be the same

src/MagnumExternal/OpenGL/GL/flextGL.h
```diff
@@ -1523,7 +1523,7 @@ typedef void (APIENTRY *GLDEBUGPROC)(GLenum source,GLenum type,GLuint id,GLenum
 
 /* GL_VERSION_4_5 */
 
-#define GL_CONTEXT_LOST 0x0507
+#define GL_CONTEXT_LOST 0x0507
```
mosra, line 1526: Unneeded change here ;)

src/Magnum/GL/Implementation/ShaderState.h
```diff
@@ -53,6 +53,7 @@ struct ShaderState {
 
     void(Shader::*addSourceImplementation)(std::string);
     void(*cleanLogImplementation)(std::string&);
+    void(*completionStatusImplementation)(GLuint, GLenum, GLint* value);
```
mosra, line 56: To fix the 32bit Windows build, this should be enough I hope -- it's just copied from another `State.h` header:

```suggestion
    /* This is a direct pointer to a GL function, so needs a __stdcall on
       Windows to compile properly on 32 bits */
    void(APIENTRY *completionStatusImplementation)(GLuint, GLenum, GLint* value);
```

Same would be in `ShaderProgramState.h`, and the fallback function declarations in `GL/Shader.h` and `GL/AbstractShaderProgram.h` would then need `APIENTRY` as well -- hopefully this is the right place to put it:

```cpp
static MAGNUM_GL_LOCAL APIENTRY void completionStatusImplementationFallback(GLuint, GLenum, GLint*);
```

I can also do this myself during merge if you don't find time in the next days.

mosra (commented):
Thank you! Except for one 32-bit-specific issue, all tests are passing, so that's great.

I didn't have time to make a final review pass the past weekend, but on a quick look everything looks alright. Your comment about `CORRADE_INTERNAL_ASSERT_OUTPUT(cs._vert.checkCompile());` etc. makes sense, keep just the `checkLink()` assert there.

I'll hopefully find time to merge everything the following weekend. No further review rounds necessary I think, I just want to go over everything one last time locally. Thanks a lot for a top-notch contribution! :+1:

mosra:
Okay, finally -- merged this as 96f97d4e7a28700b595067820a60b9c3e93281e8 and eb17d775625439292eab029d71ce658b4d2ae143, and wrote some introductory docs in 4580c30d1e4fbe1cef744d980a137c8fe0923344 and 48326ac418590fa89a4d5ee21231bb61ad0c865e.

What I only realized when testing locally was that the linker error message alone doesn't contain anything useful if the compilation failed -- somehow I falsely remembered that it contains also the compilation errors. So to provide the full context in case of a failure, as of c9d7365937fa702c9a96f8a79e3d20b9f7e243cb the `checkLink()` now accepts also a list of the input shader instances, and it calls `checkCompile()` on those first. That felt like the easiest way to achieve this while keeping it as async as possible, compared to forcing the shader code to wait on / check for individual shader compilations. Finally, in eede67175546407ec1973773116fc52b68c0ddd0 I made a change to not need to include `GL/Shader.h` in all shader headers (which pulls in `<vector>` and other nasty STL stuff). I have to fix the root cause properly (#499), but until I find time to do that I didn't want to increase compile times for everybody.

But those are all just minor additions on top. Thanks again for this work, and sorry for the extreme delays! :+1:

Issue #577, opened 2022-07-16
[Feature Request] Split mainLoopIteration() methods 


AndreasLrx:
Currently mainLoopIteration methods (for the different platform applications) does multiple actions that could be split in 3 other methods:

- handleEvents: any events including the window close check
- update: tickEvent(), when available (like SDL)
- draw: all drawing stuff (and associated drawEvent)

This would allow a better control when making custom game loop.

mosra:
Sure, I'm not opposed to this idea. But first I'd like to know more about your use case, and what exactly is hard / impossible to achieve with the current API.

AndreasLrx:
I'm starting a small game engine for personnal use (mostly as a learning project) with Magnum for rendering/physics functionalities. 

Looking at the game loop of Magnum applications, you handle the update going faster than expected minimalPeriod (with a delay).  By doing something like this:
```
while (true)
{
  double start = getCurrentTime();
  processInput();
  update();
  render();

  sleep(start + MS_PER_UPDATE- getCurrentTime());
}
```

However you do not handle the opposite case: update taking more time than expected. This case could be handled like this:
```
double previous = getCurrentTime();
double lag = 0.0;
while (true)
{
  double current = getCurrentTime();
  double elapsed = current - previous;
  previous = current;
  lag += elapsed;

  processInput();

  if (lag > MS_PER_UPDATE * 10)
    lag = MS_PER_UPDATE * 10;
  while (lag >= MS_PER_UPDATE)
  {
    update();
    lag -= MS_PER_UPDATE;
  }

  render();
}
```

 This is a possible way to manage lagging but isn't the only one. Allowing separated functions call like I suggested above would allow users to handle it the way they want.

(For more informations see https://gameprogrammingpatterns.com/game-loop.html)

mosra:
I see, that makes a lot of sense, thanks!

How about this:

- I break the `mainLoopIteration()` into three pieces, `mainLoopEventIteration()`, `mainLoopTickEventIteration()` and `mainLoopDrawEventIteration()`
- Each of those would contain also the logic like "don't call into `drawEvent()` if nothing was drawn", "don't call into `tickEvent()` if it's not implemented", so you don't need to do that yourself.
- And the `mainLoopIteration()` would itself call into these three and only handle the delays on top. So if you don't use it, handling the delays would be completely your responsibility.

AndreasLrx:
Yes this sounds great ;)

mosra:
[...]

PR #580, opened 2022-08-06
Split Sdl2Application mainLoopIteration #577

AndreasLrx:
Split Sdl2Application mainLoopIteration in:

- mainLoopEventIteration: process the application's events
- mainLoopTickEventIteration: calls tickEvent() if implemented
- mainLoopDrawEventIteration: calls drawEvent() if the redraw flag is set, and unset it

I did not split up the mainLoopIteration of the GlfwApplication/AbstractXApplication yet because the tickEvent isn't implemented in either of them. 
Should I omit the mainLoopTickEvent() or implement the tickEvent() for the others applications with the same implementation as the one in Sdl2Application ?
Or either don't even split the mainLoopIteration ? (I didn't looked at their usages and therefore don't know if it might be usefull or not)

Linked: #577


src/Magnum/Platform/Sdl2Application.cpp
```diff
@@ -993,15 +979,50 @@ bool Sdl2Application::mainLoopIteration() {
...
 
-    /* Tick event */
-    if(!(_flags & Flag::NoTickEvent)) tickEvent();
+bool Sdl2Application::mainLoopTickEventIteration() {
+    if(!(_flags & Flag::NoTickEvent)) {
+        tickEvent();
+        return true;
+    }
+    return false;
+}
```
mosra, line 991: I thought about this for a bit, and personally I wouldn't even add any `mainLoopTickEventIteration()` -- keep the code directly inside `mainLoopIteration()`. 

The point of `tickEvent()` is to run something on every event loop iteration even if the screen doesn't need to be updated (such as in regular apps that refresh their window only on input events) because with the implicit event loop there's no other way to achieve that. But if you have your own event loop handling anyway, you can just call that code directly instead of implementing some virtual interface. 

That should also answer your question about what to do for Application implementations that don't have a `tickEvent()`. (Though, implementing `tickEvent()` for GLFW at least would be nice too, but that'd need adding also `setMinimalLoopPeriod()` and handling all possible interactions with draw and input events, and I don't want to force you to do all that work as well ;))

mosra (commented):
Terribly sorry for taking ages to reply -- I had just a week off but the pile of emergency things to look at grew a lot during that time, so I only got to this PR now.

Oh and thank you for willing to implement this change for the other applications as well, I appreciate that.

mosra:
This is awesome, thank you. I have no other comments, I'll get to merging this hopefully during the weekend.

AndreasLrx:
Sorry I forgot to remove the PR from draft. I might suggest other changes when working on my project ;).

mosra:
The `tickEvent()` part of this PR is merged as a9daddfb6384a2e0e909f7ecf2c567dc4d3630e9. The rest, based on the discussion in #577, was deemed unnecessary.

PR #582, opened 2022-08-10
Utility: use Corrade String and StringView in ConfigurationValue

hugoam:
Hello !
We're trying to optimize our compile times and I noticed that StlForwardString.h is actually very heavy on MSVC because it just directly includes <string>, so I thought let's go and try to clean this up.
Let me know if I'm going at this correctly 😅

Associated Corrade PR: https://github.com/mosra/corrade/pull/146

### DIFF
--- a/src/Magnum/DebugTools/FrameProfiler.cpp
+++ b/src/Magnum/DebugTools/FrameProfiler.cpp
@@ -682,14 +682,14 @@ namespace Corrade { namespace Utility {
 using namespace Magnum;
 
 #ifdef MAGNUM_TARGET_GL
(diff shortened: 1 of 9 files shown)


mosra:
As we discussed in mosra/corrade#146, I merged a part of this as 8b45a5170c9e21ad2a29a3c8224fc5b5995fe9b0. Thank you!

Leaving the PR open, will merge the rest when I get to cleaning up STL usage in `Utility::Arguments` and `Utility::Configuration` -- until then the remaining changes don't make a big difference.

hugoam:
Thanks so much ! Yeah, the rest doesn't matter too much for us, so, I don't mind, you can even close it hehe

mosra:
It matters for me ;)

PR #588, opened 2022-08-30
GL: Switch TransformFeedback attachBuffers to use ArrayView

hugoam:
Hello !

We are hitting an issue which is that the `TransformFeedback::attachBuffers` API is quite inpractical to use. Since it only takes a `std::initializer_list`, there is no way to attach or detach a number of buffers that is not known at compîle time.

In addition to that, there is no `detachBuffers`, so you have to use the mentioned `attachBuffers` to detach your buffers, but again, you can only pass a compile-time sized list to it, so in any situation when you don't know the size at compile time, you're basically unable to detach the buffers correctly, save from writing an ugly compile-time jump table redirecting to the correct sized initializer list.

So I'm suggesting to just take an `ArrayView` in all those functions, which I don't see any disadvantages to, but if you see one please let me know of what other approach you would have in mind to enhance this API for this use case :)

One advantage of taking `ArrayView`, on the other hand, would be that, if the user really wants to pass initializer lists to those functions, they still can, but the burden of including `initializer_list` now lies on the includer and not on the `TransformFeedback.h` header (although I do think I might have forgotten to remove it from there)

As an additional improvement I think we could use an actual Struct instead of `std::tuple` for the size-offset variant, but I didn't go the extra mile of doing that. Would also be happy to, though, if you think that's a good idea

I'm leaving it as a draft for now until I hear what you think :)

EDIT: Re-reading the PR I realize I had to modify `Buffer::bind` as well, which I had forgotten. I don't know how widely it is used in general, to deduce if this would be an okay change or not, so happy to hear your thoughts.

### DIFF
--- a/src/Magnum/GL/Buffer.cpp
+++ b/src/Magnum/GL/Buffer.cpp
@@ -151,13 +151,11 @@ void Buffer::unbind(const Target target, const UnsignedInt firstIndex, const std
     Context::current().state().buffer.bindBasesImplementation(target, firstIndex, {nullptr, count});
 }
 
-/** @todoc const std::initializer_list makes Doxygen grumpy */
-void Buffer::bind(const Target target, const UnsignedInt firstIndex, std::initializer_list<std::tuple<Buffer*, GLintptr, GLsizeiptr>> buffers) {
+void Buffer::bind(const Target target, const UnsignedInt firstIndex, Containers::ArrayView<const std::tuple<Buffer*, GLintptr, GLsizeiptr>> buffers) {
     Context::current().state().buffer.bindRangesImplementation(target, firstIndex, {buffers.begin(), buffers.size()});
 }
 
-/** @todoc const std::initializer_list makes Doxygen grumpy */
-void Buffer::bind(const Target target, const UnsignedInt firstIndex, std::initializer_list<Buffer*> buffers) {
+void Buffer::bind(const Target target, const UnsignedInt firstIndex, Containers::ArrayView<Buffer* const> buffers) {
     Context::current().state().buffer.bindBasesImplementation(target, firstIndex, {buffers.begin(), buffers.size()});
 }
 
@@ -406,7 +404,6 @@ void Buffer::bindImplementationMulti(const Target target, const GLuint firstInde
 }
 #endif
 
-/** @todoc const Containers::ArrayView makes Doxygen grumpy */
 void Buffer::bindImplementationFallback(const Target target, const GLuint firstIndex, Containers::ArrayView<const std::tuple<Buffer*, GLintptr, GLsizeiptr>> buffers) {
     for(std::size_t i = 0; i != buffers.size(); ++i) {
         if(buffers && std::get<0>(buffers[i]))
@@ -416,7 +413,6 @@ void Buffer::bindImplementationFallback(const Target target, const GLuint firstI
 }
 
 #ifndef MAGNUM_TARGET_GLES
-/** @todoc const Containers::ArrayView makes Doxygen grumpy */
 void Buffer::bindImplementationMulti(const Target target, const GLuint firstIndex, Containers::ArrayView<const std::tuple<Buffer*, GLintptr, GLsizeiptr>> buffers) {
     /** @todo use ArrayTuple */
     Containers::Array<GLuint> ids{buffers ? buffers.size() : 0};
--- a/src/Magnum/GL/Buffer.cpp
+++ b/src/Magnum/GL/Buffer.cpp
@@ -26,9 +26,6 @@
 
 #include "Buffer.h"
 
-#ifdef MAGNUM_BUILD_DEPRECATED
-#include <tuple>
-#endif
 #include <Corrade/Containers/Array.h>
 #include <Corrade/Containers/Triple.h>
 #ifndef MAGNUM_TARGET_WEBGL
@@ -161,17 +158,6 @@ void Buffer::bind(const Target target, const UnsignedInt firstIndex, Containers:
 void Buffer::bind(const Target target, const UnsignedInt firstIndex, std::initializer_list<Containers::Triple<Buffer*, GLintptr, GLsizeiptr>> buffers) {
     Context::current().state().buffer.bindRangesImplementation(target, firstIndex, {buffers.begin(), buffers.size()});
 }
-
-#ifdef MAGNUM_BUILD_DEPRECATED
-void Buffer::bind(const Target target, const UnsignedInt firstIndex, std::initializer_list<std::tuple<Buffer*, GLintptr, GLsizeiptr>> buffers) {
-    Containers::Array<Containers::Triple<Buffer*, GLintptr, GLsizeiptr>> copy{NoInit, buffers.size()};
-    for(std::size_t i = 0, max = buffers.size(); i != max; ++i) {
-        const auto& t = *(buffers.begin() + i);
-        copy[i] = {std::get<0>(t), std::get<1>(t), std::get<2>(t)};
-    }
-    bind(target, firstIndex, copy);
-}
-#endif
 
 void Buffer::bind(const Target target, const UnsignedInt firstIndex, Containers::ArrayView<Buffer* const> buffers) {
     Context::current().state().buffer.bindBasesImplementation(target, firstIndex, {buffers.begin(), buffers.size()});
--- a/src/Magnum/GL/Buffer.cpp
+++ b/src/Magnum/GL/Buffer.cpp
@@ -26,8 +26,11 @@
 
 #include "Buffer.h"
 
+#ifdef MAGNUM_BUILD_DEPRECATED
 #include <tuple>
+#endif
 #include <Corrade/Containers/Array.h>
+#include <Corrade/Containers/Triple.h>
 #ifndef MAGNUM_TARGET_WEBGL
 #include <Corrade/Containers/String.h>
 #endif
@@ -151,9 +154,24 @@ void Buffer::unbind(const Target target, const UnsignedInt firstIndex, const std
     Context::current().state().buffer.bindBasesImplementation(target, firstIndex, {nullptr, count});
 }
 
-void Buffer::bind(const Target target, const UnsignedInt firstIndex, Containers::ArrayView<const std::tuple<Buffer*, GLintptr, GLsizeiptr>> buffers) {
+void Buffer::bind(const Target target, const UnsignedInt firstIndex, Containers::ArrayView<const Containers::Triple<Buffer*, GLintptr, GLsizeiptr>> buffers) {
     Context::current().state().buffer.bindRangesImplementation(target, firstIndex, {buffers.begin(), buffers.size()});
 }
+
+void Buffer::bind(const Target target, const UnsignedInt firstIndex, std::initializer_list<Containers::Triple<Buffer*, GLintptr, GLsizeiptr>> buffers) {
+    Context::current().state().buffer.bindRangesImplementation(target, firstIndex, {buffers.begin(), buffers.size()});
+}
+
+#ifdef MAGNUM_BUILD_DEPRECATED
+void Buffer::bind(const Target target, const UnsignedInt firstIndex, std::initializer_list<std::tuple<Buffer*, GLintptr, GLsizeiptr>> buffers) {
+    Containers::Array<Containers::Triple<Buffer*, GLintptr, GLsizeiptr>> copy{NoInit, buffers.size()};
+    for(std::size_t i = 0, max = buffers.size(); i != max; ++i) {
+        const auto& t = *(buffers.begin() + i);
+        copy[i] = {std::get<0>(t), std::get<1>(t), std::get<2>(t)};
+    }
+    bind(target, firstIndex, copy);
+}
+#endif
 
 void Buffer::bind(const Target target, const UnsignedInt firstIndex, Containers::ArrayView<Buffer* const> buffers) {
     Context::current().state().buffer.bindBasesImplementation(target, firstIndex, {buffers.begin(), buffers.size()});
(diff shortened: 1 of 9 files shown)


src/Magnum/GL/Buffer.h
```diff
@@ -888,7 +888,7 @@ class MAGNUM_GL_EXPORT Buffer: public AbstractObject {
          * @overload
          * @m_since{2020,06}
          */
-        template<class T> explicit Buffer(TargetHint targetHint, std::initializer_list<T> data, BufferUsage usage = BufferUsage::StaticDraw): Buffer{targetHint, Containers::arrayView(data), usage} {}
+        template<class T> explicit Buffer(TargetHint targetHint, Containers::ArrayView<T> data, BufferUsage usage = BufferUsage::StaticDraw): Buffer{targetHint, Containers::arrayView(data), usage} {}
```
mosra, line 891: This is probably not a change you wanted to do -- it's already delegating to the ArrayView overload above. Same for the `setData()` overloads below.

hugoam:
Sorry, I'm switching this to not draft status to see if this allows CircleCI to run, but I don't think it does :/

mosra:
Sigh, CircleCI, not again... Maybe it'll work on the next push again, given that your previous PR on this repo worked.

src/Magnum/GL/Buffer.h
```diff
@@ -766,7 +766,7 @@ class MAGNUM_GL_EXPORT Buffer: public AbstractObject {
          *      WebGL 1.0, see particular @ref Target values for version
          *      requirements.
          */
-        static void bind(Target target, UnsignedInt firstIndex, std::initializer_list<std::tuple<Buffer*, GLintptr, GLsizeiptr>> buffers);
+        static void bind(Target target, UnsignedInt firstIndex, Containers::ArrayView<std::tuple<Buffer*, GLintptr, GLsizeiptr>> buffers);
```
mosra, line 769: Ah, yes, it would be nice to get rid of the 30k+LOC from `std::tuple` as well... Nowadays there's [Containers::Triple](https://doc.magnum.graphics/corrade/classCorrade_1_1Containers_1_1Triple.html) that could be used instead.

The original `std::initializer_list<std::tuple>` signature needs to be still present though, for backwards compatibility, and delegate to the new variant. Unfortunately `std::tuple` has the internal order of elements reversed so just an `arrayCast()` won't cut it, the backwards compat overload would need to allocate a new array.

The old signature would look like this, inside it you'll have to allocate a new Array, copy the elements over and call with it into the new bind() overload.

```cpp
        #ifdef MAGNUM_BUILD_DEPRECATED
        /** 
         * @m_deprecated_since_latest Use @ref bind(Target, UnsignedInt, std::initializer_list<Containers::Triple<Buffer*, GLintptr, GLsizeiptr>>) 
         *      instead.
         */
        #endif
        static CORRADE_DEPRECATED("use the Containers::Triple overload instead") void bind(Target target, UnsignedInt firstIndex, std::initializer_list<std::tuple<Buffer*, GLintptr, GLsizeiptr>> buffers);
```

And then the tuple forward declaration can be moved down to the conditional as well:

```cpp
#ifdef MAGNUM_BUILD_DEPRECATED
...
/* For deprecated bind(..., std::initializer_list<std::tuple>) */
#include <Corrade/Utility/StlForwardTuple.h>
#endif
```

And same for `#include <tuple>` if it appears inside `Buffer.cpp` or elsewhere.

Oh, and as I said for the other overload, a `std::initializer_list` variant needs to be still present as well. Similarly, just delegating:

```cpp
 void bind(Target target, UnsignedInt firstIndex, std::initializer_list<Containers::Triple<Buffer*, GLintptr, GLsizeiptr>> buffers) {
     return bind(target, firstIndex, Containers::arrayView(buffers);
 }
```

You may need to deinline it into the cpp file to not have to include the Triple header. Also there *may* be an issue with the `std::tuple` and `Containers::Triple` variants being ambiguous if one passes stuff like `{{&buffer, offset, size}}` to them, but hopefully not. At least not on C++11, where one had to use `std::make_tuple()` instead of `{}`. If you hit such corner case either in the tests here or in the Wonderland codebase, please let me know, I'll try to figure out a way to resolve the ambiguity.

src/Magnum/GL/TransformFeedback.h
```diff
@@ -340,7 +340,7 @@ class MAGNUM_GL_EXPORT TransformFeedback: public AbstractObject {
          *      eventually @fn_gl{BindTransformFeedback} and
          *      @fn_gl_keyword{BindBuffersRange} or @fn_gl_keyword{BindBufferRange}
          */
-        TransformFeedback& attachBuffers(UnsignedInt firstIndex, std::initializer_list<std::tuple<Buffer*, GLintptr, GLsizeiptr>> buffers);
+        TransformFeedback& attachBuffers(UnsignedInt firstIndex, Containers::ArrayView<std::tuple<Buffer*, GLintptr, GLsizeiptr>> buffers);
```
mosra, line 343: Same here and in the overload below:  

- the `std::initializer_list` overload needs to be kept in order to allow conveniently passing a `{...}` list
- the `std::tuple` should get replaced by a Triple, with a deprecated backwards compatibility overload that allocates a temporary `Array<Triple>` and calls the new function; the `#include <tuple>` / `#include <StlForwardTuple.h>` then get included only conditionally

Hope it's clear enough, please complain if not ;)

mosra (commented):
And here I thought I already got rid of all init-list-only APIs, heh. Apparently not :sweat_smile:

mosra:
Re CircleCI, compared to e.g. #580 the builds are somehow running on `hugoam/magnum` instead of `mosra/magnum`, so that is probably the root of the problem. 

Not sure what exactly could fix that, would it be possible for you to somehow stop building the fork ("unfollow" it on CircleCI)? Some more info is here: https://support.circleci.com/hc/en-us/articles/360008097173-Troubleshooting-why-pull-requests-are-not-triggering-jobs-on-my-organization-

src/Magnum/GL/Buffer.h
```diff
@@ -1171,10 +1186,15 @@ class MAGNUM_GL_EXPORT Buffer: public AbstractObject {
          * @overload
          * @m_since{2019,10}
          */
-        template<class T> Buffer& setSubData(GLintptr offset, std::initializer_list<T> data) {
+        template<class T> Buffer& setSubData(GLintptr offset, Containers::ArrayView<T> data) {
```
mosra, line 1189: Oh, forgot to mention (and that's why all builds fail now) -- everything that used to take `std::initializer_list<T>` needs to be `Containers::ArrayView<const T>` now, not `Containers::ArrayView<T>` (the `const` is implicit in the init list type).

src/Magnum/GL/Buffer.h
```diff
@@ -796,7 +796,7 @@ class MAGNUM_GL_EXPORT Buffer: public AbstractObject {
          *      WebGL 1.0, see particular @ref Target values for version
          *      requirements.
          */
-        static void bind(Target target, UnsignedInt firstIndex, std::initializer_list<Buffer*> buffers);
+        static void bind(Target target, UnsignedInt firstIndex, Containers::ArrayView<Buffer*> buffers);
```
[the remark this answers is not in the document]
hugoam: I see, I'll do this change ASAP

Although just for the sake of theoretical completeness, I've read arguments from people more clever than me, that allowing constructing array views from initializer lists wouldn't make array views more dangerous than they already are by nature, (for example something as trivial as `ArrayView<T> view = returnsAnArrayByValue();`), and that the whole discussion about forbidding this specific conversion from initializer lists is kind of overblown

src/Magnum/GL/Buffer.h
hugoam: I'm doing only the `std::initializer_list` change in a first commit, and will do the `std::tuple` change afterwards

src/Magnum/GL/Buffer.h
hugoam: Oh, right, careless use of search and replace hehe, reverted

src/Magnum/GL/TransformFeedback.h
hugoam: Ok! I'm doing the `std::initializer_list` change in a first commit, and will do the `std::tuple` one afterwards

src/Magnum/GL/Buffer.h
hugoam: Oh, you're right, I missed that. Any preference for east const or west? I'm seeing an east const associated with a pointer in `AbstractTexture.h` but west consts in other places, don't know if that's the standard coding style or not 😅

src/Magnum/GL/Buffer.h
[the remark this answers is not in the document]
mosra: True. Ideally, there would be a way to differentiate between conversion-constructing a variable from a r-value (which becomes a dangling view and so should be forbidden) and passing a r-value to an an argument (which is fine), but so far I wasn't able to invent anything like that.

I personally got bitten by the init-list-to-view conversion quite a lot in the past, which is why I special-case for initializer lists everywhere. Compared to that, `ArrayView<T> view = returnsAnArrayByValue();` isn't something I'm hitting that often... although that started to change with strings and string views :D I might reconsider this position eventually, but for now I'd like to have `std::initializer_list` handled consistently everywhere.

hugoam:
I think it's starting to look like what you want, there is just a minor thing where I got confused by a comment `/** @todoc const std::initializer_list makes Doxygen grumpy */`, started removing it because I thought it wasn't relevant anymore, but after seeing the other occurences I'm not sure anymore of what it's supposed to mean and if it still needs to be there

hugoam:
Not sure what to do about the coverage thing. It doesn't even make much sense to me how project can be down 15%

src/Magnum/GL/Buffer.h
```diff
@@ -1171,10 +1194,15 @@ class MAGNUM_GL_EXPORT Buffer: public AbstractObject {
          * @overload
          * @m_since{2019,10}
          */
-        template<class T> Buffer& setSubData(GLintptr offset, std::initializer_list<T> data) {
+        template<class T> Buffer& setSubData(GLintptr offset, Containers::ArrayView<T> data) {
             return setSubData(offset, {data.begin(), data.size()});
         }
 
+        /** @overload */
+        template<class T> Buffer& setSubData(GLintptr offset, std::initializer_list<T> data) {
```
mosra, line 1202: All these changes to `setData()` and `setSubData()` are unnecessary -- there are `Containers::ArrayView<const void>` overloads right above that takes care of everything. Revert please ;)

src/Magnum/GL/Buffer.cpp
```diff
@@ -28,6 +28,7 @@
 
 #include <tuple>
```
mosra, line 29: This include could get wrapped in an `#ifdef MAGNUM_BUILD_DEPRECATED`.

src/Magnum/GL/Test/BufferGLTest.cpp
```diff
@@ -27,6 +27,7 @@
 #include <tuple>
```
mosra, line 27: This include is probably not needed anymore?

src/Magnum/GL/Test/TransformFeedbackGLTest.cpp
```diff
@@ -25,6 +25,7 @@
 
 #include <tuple>
```
mosra, line 26: Also not needed anymore.

src/Magnum/GL/TransformFeedback.cpp
```diff
@@ -30,6 +30,7 @@
 #ifndef MAGNUM_TARGET_WEBGL
```
mosra, line 30: The `<tuple>` include could be wrapped in an `#ifdef MAGNUM_BUILD_DEPRECATED `.

mosra (commented):
Last review round and then I'm happy to merge. Sorry for the extreme delays, I'm struggling to keep up with all the PRs and issues :sweat_smile: Worst case, if you don't get to updating the PR, I'll do the changes myself during merge. 

> I got confused by a comment `/** @todoc const std::initializer_list makes Doxygen grumpy */`, started removing it because I thought it wasn't relevant anymore, but after seeing the other occurences I'm not sure anymore of what it's supposed to mean and if it still needs to be there

Those are my stupid copypaste errors. You did the right thing with removing them ;)

About the coverage, it seems that the CircleCI jobs are somehow still running on your `hugoam/magnum` fork instead of my repo, and then it attempts to upload coverage to my codecov project, and fails because the repos don't match. And CircleCI is apparently responsible for 15% of code coverage that isn't covered by AppVeyor Windows builds. Did you try to "unfollow" the forked repo on CircleCI, as I mentioned above? It isn't really blocking the merge as you are mainly just changing existing code that was covered before, but it would be good to have that sorted out for the future.

src/Magnum/GL/Test/TransformFeedbackGLTest.cpp
```diff
@@ -488,9 +489,10 @@ void TransformFeedbackGLTest::attachRanges() {
         .setCount(2);
 
     TransformFeedback feedback;
+    using BufferOffset = Containers::Triple<Buffer*, GLintptr, GLsizeiptr>;
     feedback.attachBuffers(0, {
-        std::make_tuple(&output1, 256, 2*sizeof(Vector2)),
-        std::make_tuple(&output2, 512, 2*sizeof(Float))
+        BufferOffset{&output1, 256, 2*sizeof(Vector2)},
+        BufferOffset{&output2, 512, 2*sizeof(Float)}
```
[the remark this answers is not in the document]
hugoam: That makes the MAGNUM_BUILD_DEPRECATED build of this test fail, since the call is ambiguous (could be either the std::initializer_list variant or the ArrayView variant)

src/Magnum/GL/TransformFeedback.cpp
hugoam: Done

src/Magnum/GL/Test/TransformFeedbackGLTest.cpp
hugoam: Done

src/Magnum/GL/Test/BufferGLTest.cpp
hugoam: Done

src/Magnum/GL/Buffer.cpp
hugoam: Done

src/Magnum/GL/Buffer.h
hugoam: Done

src/Magnum/GL/Test/TransformFeedbackGLTest.cpp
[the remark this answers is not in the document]
hugoam: Hmm whoops, no actually you were right, but only once I removed the `<tuple>` include, the expression is not ambiguous anymore! So, done

src/Magnum/GL/Test/BufferGLTest.cpp
```diff
@@ -254,10 +255,10 @@ void BufferGLTest::bindRange() {
 
     MAGNUM_VERIFY_NO_GL_ERROR();
 
-    /** @todo C++14: get rid of std::make_tuple */
+    using BufferOffset = Containers::Triple<Buffer*, GLintptr, GLsizeiptr>;
     Buffer::bind(Buffer::Target::Uniform, 7, {
-        std::make_tuple(&buffer, 256, 13), {},
-        std::make_tuple(&buffer, 768, 64)});
+        BufferOffset(&buffer, 256, 13), {},
+        BufferOffset(&buffer, 768, 64)});
```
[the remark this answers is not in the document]
hugoam: Hmm whoops, no actually you were right, but only once I removed the <tuple> include, the expression is not ambiguous anymore! So, done

hugoam:
No problem for the delay, it's an open source project hehe.
I did a pass on your comments, should look pretty good now, and unfollowed the github repos on CircleCI, waiting for the CI builds now

src/Magnum/GL/Test/BufferGLTest.cpp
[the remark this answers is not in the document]
hugoam: Damn. Now the CI is failing on this file, because probably depending on the compiler `<tuple>` still gets included through an STL include chain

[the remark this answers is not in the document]
mosra: Okay, that's not good. I'll try to find a better solution, having to specify the whole type here would be a usability regression.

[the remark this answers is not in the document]
hugoam: Yeah, the thing is that if people used the "implicit" idiom with just breace initializers, we wouldn't even need to provide the deprecated alternative as it would still work with the new API. The problem is if people used `make_tuple` explicitly, then it breaks without the deprecated signature. But if we keep the deprecated signature, the implicit idiom is broken as soon as `<tuple>` is included anywhere.

I don't really see any option here, it's kind of "pick your poison" moment. Either break *some* code, or have an ambiguous signatures that force spelling out the whole type.

Another non solution is, we just drop the triple change from this PR. The change that we do need, on our side, is actually just the `ArrayView` one.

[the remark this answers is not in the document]
mosra: Right, good point.

If have to pick between porting halfway (and suffering the `<tuple>` include further) or breaking some (assuming rather rare) code, then I pick the breakage. So, can you drop the deprecated tuple overload altogether? And removing all deprecated (forward) tuple includes now as well.

Sorry for the extra work this caused 😅

mosra:
Merged as 62363094ae521d07d76cb0d007cf7ae14ad1f494 and bc1b9e39d7749416fa3238b22975296ac606772a, thank you!

PR #589, opened 2022-09-10
Add GDB pretty-printers for some types.

williamjcm:
In the wake of mosra/corrade#147, here the Magnum version of the linked PR.

Similarly, not all types are covered. Math types were prioritised, with some other types in the root Magnum namespace as well as `Animation::Track` also getting pretty-printers.


mosra:
Merged as 13d8291cbc459dbac197c627d26de5fda592c141, thanks!

Issue #591, opened 2022-09-12
Make GL::AbstractShaderProgram::use() protected


hugoam:
Hello,

We have to circumvent a bug in Oculus Quest which requires us to call a GL function on the program of an `GL::AbstractShaderProgram` directly, which requires the program to be bound.
Unfortunately, `GL::AbstractShaderProgram::use()` is private, so we can't call it from the derived class, and we can't either call `glUseProgram()` ourselves without triggering an inconsistency with Magnum internal state (which will not know we changed the currently bound program and risk causing more silent bugs down the line).

Would making `GL::AbstractShaderProgram::use()` protected make sense to allow more fine grained control from the derived class? Or do you have another suggestion to solve this use case?

Thanks :)

mosra:
This would ideally be turned into a Quest-specific driver workaround, i.e. something that's handled by Magnum itself and not an application-specific "patch". I had to do quite a few of these for Mac GL drivers, for Intel drivers on Windows, some for ARM Mali drivers, etc., so it would be nothing uncommon. Plus that way Magnum itself can benefit from that, instead of the bug having to be fixed again by every app using Magnum on the Quest.

Can you elaborate on what exactly is the bug and what needs to be done to circumvent it?

Squareys:
@mosra Unfortunately, this is all on WebGL and while detecting Quest might be possible somehow, it is super inconvenient. Magnum currently doesn't have any framework for detecting browser version etc, and even then, with the upcoming Quest Pro, that would no longer be sufficient to detecting the Quest 1/2 here 🤔

mosra:
Yes, but still -- I have to deal with Quest as well, and any Quest bugs that aren't worked around by Magnum directly are viewed by certain end users as "Magnum is broken", which is not the case. 

Thus I'd really like to have this handled directly here, even if it would ultimately mean one has to explicitly tell the library that "this is a Quest, do your best".:D

hugoam:
Here is a link to the Oculus bug: https://forums.oculusvr.com/t5/Quest-Development/WebGL-glUniform1iv-driver-bug-when-setting-sampler-texture-units/m-p/985095#M4929

Circumventing the bug would basically involve using `glUniform1i` instead of `glUniform1iv` in the `AbstractShaderProgram::setUniform(Int, Int)` signature, the latter being broken.

[...]

PR #592, opened 2022-09-13
ShaderProgram: Add glUniform for single value calls using non-v variants

hugoam:
After discussion in this issue: https://github.com/mosra/magnum/issues/591
Here is a draft of adding the `setUniform` calls mapping to non-v variants of `glUniform` functions :)


mosra:
So it was 16 variants after all, not 12 :) Thanks! I'll merge it ASAP, there isn't much else to do here.

mosra:
Wow, the codecov reports are utter garbage. What is going on in there.

mosra:
Merged as 40d132664fd2f2401a94120c915d01907790cbeb (I did a rebase against latest master), thank you!

PR #594, opened 2022-09-27
Update FindSDL2.cmake to find static lib

hsdk123:
Be able to find `SDL2-static` if sdl was built with static configurations externally.


mosra:
Thanks! Don't worry about the failed CIs, I'll sort that out.

mosra:
Merged as f58e6766ffea28a462130709830aa10d09ef4534, thank you!

PR #596, opened 2022-10-09
GL: don't reallocate std::string in DebugOutput

sthalik:


mosra:
If that's okay with you, I think this will be better handled without STL altogether, which is what #499 is meant to do. I'll finish & merge that PR once I have some free time again, unfortunately the top-priority tasks I have for this month don't include it.

mosra:
Just FYI, this is fixed with d89b882ccf4d4af6c4a1f9e9c7ef63ddffaee85b. The `std::string` callback is still present for backwards compatibility, but to avoid a temporary allocation use a `StringView` instead.

PR #597, opened 2022-10-09
Add constexpr to Vector{2,3,4} and Vector<N, T>

sthalik:
Let's sprinkle constexpr with reckless abandon.

I'm still writing tests for all vector operators and subclasses. Would like your comments as to whether this is a reasonable approach.

### DIFF
--- a/src/Magnum/Math/Test/CMakeLists.txt
+++ b/src/Magnum/Math/Test/CMakeLists.txt
@@ -41,6 +41,13 @@ corrade_add_test(MathVectorTest VectorTest.cpp LIBRARIES MagnumMathTestLib)
 corrade_add_test(MathVector2Test Vector2Test.cpp LIBRARIES MagnumMathTestLib)
 corrade_add_test(MathVector3Test Vector3Test.cpp LIBRARIES MagnumMathTestLib)
 corrade_add_test(MathVector4Test Vector4Test.cpp LIBRARIES MagnumMathTestLib)
+
+foreach(_test VectorTest Vector2Test Vector3Test Vector4Test)
+    corrade_add_test(Cpp14Math${_test} ${_test}.cpp LIBRARIES MagnumMathTestLib)
+    set_target_properties(Cpp14Math${_test} PROPERTIES CORRADE_CXX_STANDARD 14)
+    target_compile_definitions(Cpp14Math${_test} PRIVATE TESTING_CONSTEXPR CORRADE_GRACEFUL_ASSERT)
+endforeach()
+
 corrade_add_test(MathColorTest ColorTest.cpp LIBRARIES MagnumMathTestLib)
 
 corrade_add_test(MathRectangularMatrixTest RectangularMatrixTest.cpp LIBRARIES MagnumMathTestLib)
--- a/src/Magnum/Math/BitVector.h
+++ b/src/Magnum/Math/BitVector.h
@@ -35,6 +35,7 @@
 #ifndef CORRADE_NO_DEBUG
 #include <Corrade/Utility/Debug.h>
 #endif
+#include <Corrade/Utility/Macros.h>
 
 #include "Magnum/Types.h"
 #include "Magnum/Math/Math.h"
@@ -174,14 +175,14 @@ template<std::size_t size> class BitVector {
          * Equivalent to @ref all().
          * @see @ref any(), @ref none()
          */
-        explicit operator bool() const { return all(); }
+        explicit MAGNUM_CONSTEXPR14 operator bool() const { return all(); }
 
         /**
          * @brief Whether all bits are set
          *
          * @see @ref none(), @ref any(), @ref operator bool()
          */
-        bool all() const;
+        MAGNUM_CONSTEXPR14 bool all() const;
 
         /**
          * @brief Whether no bits are set
@@ -361,7 +362,7 @@ template<std::size_t size> inline bool BitVector<size>::operator==(const BitVect
     return true;
 }
 
-template<std::size_t size> inline bool BitVector<size>::all() const {
+template<std::size_t size> MAGNUM_CONSTEXPR14 inline bool BitVector<size>::all() const {
     /* Check all full segments */
     for(std::size_t i = 0; i != size/8; ++i)
         if(_data[i] != FullSegmentMask) return false;
--- a/src/Magnum/Math/Math.h
+++ b/src/Magnum/Math/Math.h
@@ -116,4 +116,17 @@ namespace Implementation {
 
 }}
 
+// TODO remove this once corrade#152 gets merged
+#if defined __cpp_constexpr && __cpp_constexpr >= 201304 || \
+    defined CORRADE_TARGET_MSVC && _MSC_VER >= 1910 && CORRADE_CXX_STANDARD >= 201402L
+#define MAGNUM_CONSTEXPR14 constexpr
+#else
+#define MAGNUM_CONSTEXPR14
+#endif
+#if defined(__clang__) && __clang_major__ >= 9 || defined(__GNUG__) && __GNUG__ >= 9 || defined(_MSC_VER) && _MSC_VER >= 1931
+#define MAGNUM_CONSTEVAL (__builtin_is_constant_evaluated())
+#elif CORRADE_CXX_STANDARD >= 202002L
+#define MAGNUM_CONSTEVAL (std::is_constant_evaluated())
+#endif
+
 #endif
--- a/src/Magnum/Math/Test/Cpp14VectorTest.h
+++ b/src/Magnum/Math/Test/Cpp14VectorTest.h
@@ -1,0 +1,14 @@
+#ifndef Magnum_Math_Test_Cpp14Vector
+#define Magnum_Math_Test_Cpp14Vector
+
+#define CE
+#ifdef TESTING_CONSTEXPR
+#if __cpp_constexpr >= 201304
+#undef CE
+#define CE constexpr
+#else
+#define SKIP_TESTING
+#endif
+#endif
+
+#endif
--- a/src/Magnum/Math/Test/Vector2Test.cpp
+++ b/src/Magnum/Math/Test/Vector2Test.cpp
@@ -32,6 +32,8 @@
 #include "Magnum/Math/StrictWeakOrdering.h"
 #include "Magnum/Math/Swizzle.h"
 
+#include "Cpp14VectorTest.h"
+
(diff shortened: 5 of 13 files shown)


mosra:
Thank you for starting the work on this. I think having post-C++11 constexpr is desirable in general, but -- as we discussed on Gitter -- I still don't know how to resolve the major blocker here:

Let's say this would get merged, enabling constexpr for everything that can have it under C++14, including for example a 4x4 matrix multiplication, or a vector cross product, and people gradually start relying on these being constexpr. But then I'd want to add SIMD-optimized variants of those two (which I already have in a private branch), and I wouldn't be able to because in order to use SIMD intrinsics I'd have to drop the constexpr again, breaking everyone's code. Which I really don't want to. So how should we solve this? I see these options:

1. Enable post-C++11 constexpr only for various mutable getters, and not for anything that is potentially SIMD-optimizable. Which probably goes against your use case, and it's also unclear where we should draw the line.
2. Enable post-C++11 constexpr only on compilers that implement function "overloading" based on whether the function is constant-evaluated (i.e., that `__builtin_constant_evaluated()` you mentioned). Then, when I'd want to equip those with SIMD, I'd just add an "overload". What I'm not sure about is how those "overloads" are meant to be done, everything I found was about adding some a branch inside the implementation and deciding there. Is that the way? Or can I have some hypotetical `int foo()` and `constexpr int foo()` overloads?
3. Or decide that SIMD will never be added directly to vector classes, and only used in other algorithms, and enable post-C++11 constexpr everywhere, as you do now. Sure, using SIMD in singular matrix/vector APIs is considered a "naive SIMD implementation" due to its relative inefficiency, nevertheless I'd like to keep that possibility.
4. Other options / ideas?

So far I think option 2 would be the best. Let's say with a `CORRADE_CONSTEXPR14_IF_CONSTEVAL` macro that's non-empty only if the compiler supports C++14 constexpr *and* `__builtin_constant_evaluated()` or equivalent is available in the compiler in the currently chosen standard.

src/Magnum/Math/Test/VectorConstexprOps14.cpp
```diff
@@ -0,0 +1,100 @@
+/*
```
mosra, line 1: For the tests, I think a more feasible option would be to reuse the existing test files, compile them a second time with C++14 enabled, and add a constexpr variant of each test case where C++14 constexpr got enabled.

Let's say, for example, the [VectorTest::addSubtract()](https://github.com/mosra/magnum/blob/a55fe9db4de187f8192d4e329f376e9b4668a20b/src/Magnum/Math/Test/VectorTest.cpp#L402-L409) would be extended like this:

```cpp
void VectorTest::addSubtract() {
    Vector4 a(1.0f, -3.0f, 5.0f, -10.0f);
    Vector4 b(7.5f, 33.0f, -15.0f, 0.0f);
    Vector4 c(8.5f, 30.0f, -10.0f, -10.0f);

    CORRADE_COMPARE(a + b, c);
    CORRADE_COMPARE(c - b, a);
    
    CORRADE_CONSTEXPR14 Vector4 ca(1.0f, -3.0f, 5.0f, -10.0f);
    CORRADE_CONSTEXPR14 Vector4 cb(7.5f, 33.0f, -15.0f, 0.0f);
    CORRADE_CONSTEXPR14 Vector4 cc(8.5f, 30.0f, -10.0f, -10.0f);

    CORRADE_CONSTEXPR14 Vector4 caPlusCb = ca + cb;
    CORRADE_CONSTEXPR14 Vector4 ccMinusCb = cc - cb;
    CORRADE_COMPARE(caPlusCb, cc);
    CORRADE_COMPARE(ccMinusCb, ca);
}
```

That way both the runtime and the constexpr tests will be localized close to each other, making it relatively easy to ensure that all constexpr variants are indeed tested. And when the runtime and compile-time implementation diverges due to SIMD, having the two cases together makes it easy to verify both still do the same in various potential corner cases.

src/Magnum/Math/Test/CMakeLists.txt
```diff
@@ -41,6 +41,11 @@ corrade_add_test(MathVectorTest VectorTest.cpp LIBRARIES MagnumMathTestLib)
 corrade_add_test(MathVector2Test Vector2Test.cpp LIBRARIES MagnumMathTestLib)
 corrade_add_test(MathVector3Test Vector3Test.cpp LIBRARIES MagnumMathTestLib)
 corrade_add_test(MathVector4Test Vector4Test.cpp LIBRARIES MagnumMathTestLib)
+corrade_add_test(MathVectorConstexprOps14 VectorConstexprOps14.cpp LIBRARIES MagnumMathTestLib)
+set_target_properties(
+    MathVectorConstexprOps14
+    PROPERTIES CORRADE_CXX_STANDARD 14)
```
mosra, line 47: For various reasons I'm keeping compatibility with GCC 4.8, which doesn't support C++11 yet, so the C++14-only tests need to be compiled only if the compiler supports that. [Here is how it's done inside Corrade](https://github.com/mosra/corrade/blob/9e35d4eb31b60c8d0aaa2d0028eae469b59ed7b3/src/Corrade/Test/CMakeLists.txt#L61-L72), tho the condition should probably be put into some global place (a variable in `UseCorrade.cmake`?) and reused, instead of copypasting it thousand times.

Continuing with my suggestion of compiling the same test again, this would look like

```cmake
if(...) 
    corrade_add_test(MathVector2Cpp14Test Vector2Test.cpp LIBRARIES MagnumMathTestLib)
    ...
    set_target_properties(
        MathVector2Cpp14Test
        ...
        PROPERTIES CORRADE_CXX_STANDARD 14)
endif()
```

mosra (commented):
Some more initial ideas for this.

Oh, and apologies in advance -- until end of October I have very little free time left, so it's very likely I won't be able to do another review round. Sorry.

sthalik:
> Enable post-C++11 constexpr only on compilers that implement function "overloading" based on whether the function is constant-evaluated (i.e., that __builtin_constant_evaluated() you mentioned). Then, when I'd want to equip those with SIMD, I'd just add an "overload". What I'm not sure about is how those "overloads" are meant to be done, everything I found was about adding some a branch inside the implementation and deciding there. Is that the way?

A hypothetical `CONSTEVAL` macro could be defined as:

```c++
#if CORRADE_CXX_STANDARD >= 202002L
#    define CONSTEVAL (std::is_constant_evaluated())
#elif defined(__GNUG__) || defined(_MSC_VER) && _MSVC_LANG >= 202002L
#    define CONSTEVAL (__builtin_is_constant_evaluated())
#elif CORRADE_CXX_STANDARD <= 201402L
#    define CONSTEVAL constexpr(false)
#else
#    define CONSTEVAL constexpr(Implementation::simd_compiled_out())
#endif
```

> Other options / ideas?

There are lots of out-of-class inline functions and conversion helpers. Which is why I attempted to put constexpr on everything in the first place.

Rewriting the vector code to be a single type and having less out-of-class inline functions lying around could be an option. Regarding not wanting it to become another Eigen, that shouldn't be the case unless it thinks it's a vector (including 3D cross product) and a matrix and a column vector (including 3D cross product) at the same time. I've had my own experience with `cv::MatExpr` as well so I'd rather not have it become like that as well.

I'll have to keep pushing to this branch in order to run it on your CI before it even compiles in the base non-constexpr case. Before that it's hard for me to answer as to what should be done about tests or which specific functions can be enabled at which compiler and standard conformance switch.

> Or can I have some hypotetical int foo() and constexpr int foo() overloads?

Unfortunately not, and C++23 didn't fix it either.

src/Magnum/Math/Test/VectorConstexprOps14.cpp
sthalik: If you do that then the entire test including its runtime logic will fail to compile in case of bugs specifically pertaining to constexpr rather than fail particular subtests in CORRADE_COMPARE.

What I recommend instead is #including the .cpp file from another .cpp test file and adding a 'Constexpr' suffix via a #define. Before that, do `#define CONSTEXPR constexpr` and have the original test do:

```c++
#ifndef CONSTEXPR
#define CONSTEXPR
#endif
```
and then plaster it in front of each and every vector declaration.

If you somehow dislike #including .cpp files, it could be added a second time in CMake with `-DTEST_CONSTEXPR` added as as a target compile definition. Then a header file in `Math/Test` can contain the boilerplate pertaining to it.

mosra:
Sorry for abruptly cancelling the CIs -- can you merge (or rebase on) current `master`, in particular b9726a982305a59273b2eae8109ce748311d29fc? I'm trying to avoid unnecessary CircleCI builds to not run out of credits at the end of the month again :] 

By all means, feel free to continue to abuse the CI for testing if stuff compiles, that's what it is for anyway, just merge that commit first. Thank you. It'll now run the essential Linux jobs first and only if they pass it'll continue further with other jobs and other platforms (such as [shown here](https://app.circleci.com/pipelines/github/mosra/magnum/861/workflows/0a64ad18-16e2-48b5-b5a0-fb32190f059d) for current `master` build).

sthalik:
It's pretty much ready in the current state and rebased onto master. But with some exceptions:

- needs to be squashed into a single commit with a new message written.
- `MAGNUM_CONSTSEXPR14` should be replaced with `CORRADE_CONSTEXPR14` once that definition doesn't trip up GCC 4.8.

mosra:
I just commited the C++14 constexpr additions to all mutable getters, which is the majority of this PR, along with corresponding tests in f297d593bf7d730870220591a6b19739443bdaea. The rest I currently don't intend to mark as `constexpr` to leave room for runtime optimization, as we discussed above. Some of the APIs _might_ eventually become `constexpr` on C++20 only, but so far I don't have any convincing use cases.

LB--:
Maybe partially discussed already - for the ones you want to optimize, instead of constexpr and non-constexpr overloads, you can just default to the non-constexpr versions with better runtime performance and require clients to upgrade to C++20 or C++23 if they want the ability to use them at compile time as well (or specify some build option in earlier C++ standards), since you can add guarded use of [`if(std::is_constant_evaluated()) {` (C++20)](https://en.cppreference.com/w/cpp/types/is_constant_evaluated.html) and [`if consteval {`/`if !consteval {` (C++23)](https://en.cppreference.com/w/cpp/language/if.html#Consteval_if) to provide the constexpr versions within the same function. C++ allows functions to be marked constexpr as long as there is some control flow that can be evaluated at compile time, so the unused runtime-only branches are generally fine to leave there.

PR #601, opened 2022-10-14
Anti-aliased wide line rendering shader and related MeshTools

mosra:
TODOs left:

- [x] Create a MeshTools utility that converts an arbitrary Lines / LineLoop MeshData to a list of quads to be fed to the shader, flexible enough to support (packed) vertex colors, object ID, line distance attributes as well
  - [x] Test -- have an internal variant that produces a MeshData and verify sanity on that mainly, rendering tests just to be sure all attributes are correctly bound
  - [x] Document
  - [x] ~~Add an option to figure out joins from the index buffer for `Lines` inputs (by default those are treated as separate segments), it should somehow check that a point is always a part of two segments at most, not more~~ postponed
    - [x] Conversely, if the `Lines` input is just segments alone, ~~the prev/next attributes could alias the position as they're not used/needed at all~~ not really but yes, there's a similar optimization -- postponed
      - [x] ~~Maybe even add an option to the shader for this, to treat everything as standalone segments to speed up?~~ postponed
  - [x] Fill the index buffer in a way that makes the configured mesh compatible with regular GL lines (of course with overlapping lines as well as zero-length lines, but that's fine), for a potential opt-in glTF `MAGNUM_wide_lines` extension
    - [x] Add a test for that specificially
  - [x] ~~And a utility that can do this in a streaming way instead of always from scratch (adding line strips, single points, segments...)~~ not critical for MVP, postponed
- [x] Configurable line cap (butt, square, round, triangle)
- [x] Configurable line join (miter, bevel)
  - [x] ~~Round join and [miter clip join](https://www.w3.org/TR/svg-strokes/#LineJoin) as well, once I figure out the math (hint: vector projection)~~ postponed
- [x] Configurable line join angle at which it flips from a miter to a bevel
  - [x] Rework the calculation to avoid overlapping geometry if possible by adding an extra triangle
- [x] Put point categorization bits into a separate attribute that can be an 8-bit integer instead, saving ~10% of the data
- [x] Investigate the overlapping vertex layout to save 66% (?) of the data, at the cost of disjoint lines needing extra dummy vertices
  - [x] Investigate shared bevel points on inner edge / miter points on outer edge to save another ~25% / ~33% of the rest 
    - [x] ~~Needs extra effort to calculate proper edge factors in the shared vertices~~ postponed
  - [x] ~~Investigate quad instancing to save 84% (?) of the data~~ not worth, as I'd need five attributes instead of four (first line point, second line point, two neighbors and categorization bits for all four), for which there isn't much binding slot space left, and the vertex shader would have to be significantly rewritten 
- [x] Add regression test cases for the remaining (literal) corner cases / artifacts
  - [x] Fix when line gets thinner at certain acute joint angles -- seems to be solved now (doesn't manifest in the demo anymore), there was a case where some of the calculations weren't done in screen space which was the cause of this
  - [x] ~~Fix smoothing factor calculation for long bevel joints -- needs to pass an extra attribute from the VS containing the bevel edge length, which gets combined with the edge distance and endpoint distances~~ postponed
  - [x] ~~Fix if a join next to the line cap extends beyond the center point of the line, the round cap looks wrong -- the point needs to remember not just whether the point itself is a cap, but also if the other segment point is a cap and act accordingly~~ postponed
  - [x] ~~Fix various nasty corner cases when a line join has endpoints too close to the other line~~ postponed
  - [x] ~~Fix the case where two joined line segments overlapping each other *exactly* disappear in the output -- some NaN or infinity causing issues probably, might be the same fix as for the above~~
- [x] Test & document behavior with zero-length line segments
  - [x] ~~Probably would cause issues when a part of a line strip, investigate~~ postponed
- [x] Android WebGL browser shows a strange extra artifact in the demo -- not anymore
- [x] Expand to 3D
  - [x] use `noperspective` to avoid artifacts, add a `/gl_Position.w` fallback on platforms that don't have it
    - [x] ~~backport the same thing to `MeshVisualizer` so it looks always the same independently of `GL_NV_shader_noperspective_interpolation` being present on ES~~ postponed
  - [x] Test depth handling (render together with Phong)
    - [x] ~~Fix to be more precise with wider lines (it needs to extend the caps/joins including depth, not just screen-space)~~ postponed
- [x] Vertex colors
- [x] Object ID output
- [x] Instancing
- [x] UBO support
  - [x] Multidraw support
- [x] Manually CSE the shader code
  - [x] Clean up the messy code in `*.frag`
- [x] ~~ES2/WebGL1 support? Would need a manual `gl_VertexID` input, should I still bother?~~ Not bothering with this anymore, I didn't bother with ES2 for skinning either and neither I will for PBR shaders.
- [x] Documentation
  - [x] ~~Credit https://twitter.com/hypersomniac_/status/1575776294096490496~~ will do once I actually use the overlapping layout
- [x] Explicitly verify the shader on a bunch of WebGL and GLES implementations, in addition to the CI jobs

A gradually-updated demo is at https://tmp.magnum.graphics/line-playground/

line-playground


mosra:
This branch got to a practically usable state now, so I'm merging it to unblock people waiting on it being mainlined:

- [Extensive documentation of `Shader::LineGL`](https://doc.magnum.graphics/magnum/classMagnum_1_1Shaders_1_1LineGL.html)
- [Documentation of the `MeshTools::compileLines()` utility](https://doc.magnum.graphics/magnum/namespaceMagnum_1_1MeshTools.html#afdd01db68b89b3149f0a72817564adb5)

TODOs from the above list that are marked as "postponed" are tracked in the projects view, and will be worked on them directly on the `master` branch. While there are known artifacts, they're only significant in really specific corner cases and shouldn't affect practical use too much, nevertheless I'll be fixing them. As well as implementing the more efficient mesh data layout (as documented) and support for stippling and textures.

PR #603, opened 2022-10-23
Platform: use const argv in Sdl2Application

sthalik:


mosra:
Same as before -- to save us both time, next time you need a certain change, *please* ask first and describe your use case. You're on Gitter already anyway, so just post a question there. Magnum is not the type of a project where I just merge any description-less PR that gets opened, and if things are done in a certain way, in most cases there's a reason why, and every change requires backwards compatibility considerations to avoid breaking dozens of existing projects.

In this case, the `*Application` classes are deliberately using a mutable `int&` *and* a mutable `char**` to allow user code to modify the argc/argv parameters. For example to filter away certain arguments to avoid "unknown argument" errors. [Qt does the same, by the way.](https://doc.qt.io/qt-5/qapplication.html#QApplication)

PR #604, opened 2022-10-26
add simple timer functionality to Timeline

sthalik:
The Timeline class already has all the data it needs to act as a stopwatch timer.

One example use-case is jitter-correcting framerate throttling when minimized or lacking window focus.

### DIFF
--- a/src/Magnum/Timeline.cpp
+++ b/src/Magnum/Timeline.cpp
@@ -61,4 +61,19 @@ Float Timeline::previousFrameTime() const {
     return duration_cast<microseconds>(_previousFrameTime-_startTime).count()/1e6f;
 }
 
+Float Timeline::currentFrameDuration() const {
+    if (!running) return 0;
+
+    auto now = high_resolution_clock::now();
+    auto duration = UnsignedInt(duration_cast<microseconds>(now-_previousFrameTime).count());
+    return duration/1e6f;
+}
+
+Float Timeline::currentFrameTime() const {
+    if (!running) return 0;
+
+    auto now = high_resolution_clock::now();
+    return duration_cast<microseconds>(now-_startTime).count()/1e6f;
+}
+
 }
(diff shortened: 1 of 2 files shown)


src/Magnum/Timeline.h
```diff
@@ -134,6 +134,8 @@ class MAGNUM_EXPORT Timeline {
          */
         Float previousFrameDuration() const { return _previousFrameDuration; }
 
+        float currentFrameDuration() const;
```
mosra, line 137:
```suggestion
        Float currentFrameDuration() const;
```

And a documentation block wouldn't hurt, especially clarifying the behavior for what it does when the timeline is stopped -- what should it return in that case? 0? I'm thinking it probably shouldn't return some nasty huge number.

src/Magnum/Timeline.cpp
```diff
@@ -61,4 +61,10 @@ Float Timeline::previousFrameTime() const {
     return duration_cast<microseconds>(_previousFrameTime-_startTime).count()/1e6f;
 }
 
+float Timeline::currentFrameDuration() const
+{
```
mosra, line 65:
```suggestion
Float Timeline::currentFrameDuration() const {
```

src/Magnum/Timeline.cpp
```diff
@@ -61,4 +61,10 @@ Float Timeline::previousFrameTime() const {
     return duration_cast<microseconds>(_previousFrameTime-_startTime).count()/1e6f;
 }
```
mosra, line 62: For feature parity I'm thinking `currentFrameTime()` could make sense as well; this one also appropriately handling the case of the timeline being stopped.

mosra:
Merged as 28999d113ada927aeed9de4bc26c48217dbeb585, thanks! 

This class was quite neglected so I attempted adding a test as well... and then the CIs were way more flaky than expected, so it took quite some time to get passing.

PR #605, opened 2022-11-10
Platform: silence Sdl2App warning on <SDL.h>

sthalik:
Magnum is otherwise clean on `-Wold-style-cast`. This warning is enabled by default on clang, but not presently enabled by default on GCC.

### DIFF
--- a/src/Magnum/Platform/Sdl2Application.cpp
+++ b/src/Magnum/Platform/Sdl2Application.cpp
@@ -27,6 +27,10 @@
 
 #include "Sdl2Application.h"
 
+#ifdef CORRADE_TARGET_GCC
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wold-style-cast"
+#endif
 #ifdef CORRADE_TARGET_CLANG_CL
 /* SDL does #pragma pack(push,8) and #pragma pack(pop,8) in different headers
    (begin_code.h and end_code.h) and clang-cl doesn't like that, even though it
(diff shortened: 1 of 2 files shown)


mosra:
> This warning is enabled by default on clang, but not presently enabled by default on GCC.

I'm enabling it on both, tho: https://github.com/mosra/corrade/blob/6f115b267502859c2d713c74f9843002c10b153b/modules/UseCorrade.cmake#L105

Does it mean you get the warning on Clang only, and not on GCC? Can you post the warning here, just so I'm aware which part of the header causes it? Thanks.

Also, this is due to SDL being a subproject, which means the include paths are not supplied with `-isystem` that'd suppress the warnings, right? In other words, if you'd use `add_subdirectory(SYSTEM)` [from CMake 3.25](https://cmake.org/cmake/help/latest/release/3.25.html#commands), the warning would be gone, right?

sthalik:
Here's the warning on clang,

```console
In file included from F:/dev/game/external/magnum/src/Magnum/Platform/Sdl2Application.cpp:41:
In file included from F:/dev/game/build/external/sdl2/include/SDL.h:41:
F:/dev/game/build/external/sdl2/include/SDL_events.h:676:65: warning: use of old-style cast [-Wold-style-cast]
SDL_COMPILE_TIME_ASSERT(SDL_Event, sizeof(SDL_Event) == sizeof(((SDL_Event *)NULL)->padding));
                                                                ^            ~~~~
F:/dev/game/build/external/sdl2/include/SDL_stdinc.h:379:57: note: expanded from macro 'SDL_COMPILE_TIME_ASSERT'
#define SDL_COMPILE_TIME_ASSERT(name, x)  static_assert(x, #x)
                                                        ^
1 warning generated.
```

```c++
SDL_COMPILE_TIME_ASSERT(SDL_Event, sizeof(SDL_Event) == sizeof(((SDL_Event *)NULL)->padding));
```

GCC somehow doesn't emit a warning in that case.

mosra:
As the warning doesn't happen with SDL included from a system location (where CMake uses `-isystem`), the solution here is to use `add_subdirectory(SYSTEM)` in CMake 3.25+ as mentioned above, or a corresponding equivalent with CMake's `FetchContent` and other alternatives.

PR #610, opened 2022-12-12
Fix a compile error in Line.vert under gles.

dynon-msb:


mosra:
Thanks! I'll try to get back to #601 as soon as I'm able to.

It's interesting that this was a problem, given that the shader compiled just fine under WebGL which is a GLES subset (or at least in my testing it did).

mosra:
This is now integrated into #610, thank you!

PR #612, opened 2023-01-03
Fix Emscripten TextureImage.cpp no deprecated build

hugoam:
The Emscripten build in non-deprecated mode is broken due to this file, this seems to be the only issue with that build configuration

### DIFF
--- a/src/Magnum/DebugTools/TextureImage.cpp
+++ b/src/Magnum/DebugTools/TextureImage.cpp
@@ -37,6 +37,7 @@
 #if defined(MAGNUM_TARGET_GLES) && !defined(MAGNUM_TARGET_GLES2)
 #include <Corrade/Containers/Iterable.h>
 #include <Corrade/Containers/Reference.h>
+#include <Corrade/Containers/StringView.h>
 #include <Corrade/Utility/Resource.h>
 
 #include "Magnum/GL/AbstractShaderProgram.h"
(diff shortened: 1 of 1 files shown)


src/Magnum/DebugTools/TextureImage.cpp
```diff
@@ -56,7 +57,7 @@ static void importDebugToolsResources() {
 namespace Magnum { namespace DebugTools {
 
 #if defined(MAGNUM_TARGET_GLES) && !defined(MAGNUM_TARGET_GLES2)
-using namespace Containers::Literals;
+using namespace Corrade::Containers::Literals;
```
mosra, line 60: This part shouldn't be needed I think, right? Only the include.

mosra (commented):
Thank you! And sorry for the breakage :)

src/Magnum/DebugTools/TextureImage.cpp
hugoam: Hmm yep it seems that you are right

mosra:
Merged without the literal change as ecd0d5572643c546035ba58ebdc2b2a9d5f99b44, and the other issue you reported on Gitter as d3c690b865afdb8a829414e3490ef4a5eae6c44e and mosra/magnum@d3c690b865afdb8a829414e3490ef4a5eae6c44e (with your name). Thanks a lot!

PR #613, opened 2023-01-05
Find static debug sdl2

hsdk123:
As per title.


mosra:
Makes total sense, thank you! :+1:

PR #617, opened 2023-04-27
Don't force static for Android

bansan85:
Related to mosra/corrade#170

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -330,7 +330,15 @@ option(MAGNUM_BUILD_DEPRECATED "Include deprecated API in the build" ON)
 set(MAGNUM_DEPLOY_PREFIX "."
     CACHE STRING "Prefix where to put final application executables")
 
-option(MAGNUM_BUILD_STATIC "Build static libraries (default are shared)" OFF)
+if(CORRADE_TARGET_EMSCRIPTEN OR CORRADE_TARGET_ANDROID)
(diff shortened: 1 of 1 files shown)


bansan85:
`libMagnumEglContext` is still forced to static. `flextGLInit` is not an exported symbol.

mosra:
Amazing, thanks for updating and testing this side as well :+1: 

> `libMagnumEglContext` is still forced to static

Yes, those are deliberately static everywhere always since they're extremely tiny and are meant to be only used from the top-level application executable/library to initialize GL, and from nowhere else after that. Given that usage pattern, so far there weren't any problems with that, but of course let me know if your use case would be different.

bansan85:
In my case, it works well. But EGL is used by only one of my shared libraries.

mosra:
Merged as a9898a05ac5eb6de9f1b068094ec4edfbf86c61d, sorry for the delay. The `MAGNUM_BUILD_PLUGINS_STATIC` is however still forced on Android, because the PluginManager itself doesn't have any support for loading dynamic libraries there. It'd need to get implemented in it first. I think it's not as much of a problem for the plugins, since they're relatively small and don't contain any shared mutable global state that could cause problems.

I also mirrored this change in other repos as mosra/magnum-plugins@127a9ecbbba3c3beccd22f41d928e7017817368d, mosra/magnum-extras@f567dba09c552132bb6bf1698c08cdd31a16b025 and mosra/magnum-integration@35c05fda3e243b696e2b3a3dcd230c819ca108e0.

Thank you!

PR #623, opened 2023-07-13
debugger/natvis additions for Vector, Quaternion, Deg & Rad types

fauder:
Many thanks to @pezcode for the initial work he did. I also stole that file and used it in my own opengl framework repo 🙏

I've made some modifications to the magnum.natvis file in order to augment debug visualization for:

- `Vector` & `Quaternion` types: to display whether they are normalized (i.e, unit vectors/quaternions) or not,
- Angle (`Deg` & `Rad`) types: to display the current value nicely (with degrees symbol for `Deg` & `radians` suffix for `Rad`) as well as display the other equivalent angle value (`Rad` for `Deg` & `Deg` for `Rad` respectively) in parentheses.

Magnum's epsilon (from `Magnum::Math::TypeTraits`) values were utilized for unit vector queries, by creating two intrinsic .natvis functions `Dot()` & `IsNormalized()`, the latter utilizing said epsilons (added to and subtracted from 1.0, as we can not use `abs()` in the .natvis file).

Similarly, `Magnum::Math::Constants<T>::pi()` was used in `Deg`/`Rad` conversions.

Separate `Float` & `Double` variants were added for all types mentioned since they use different epsilons/conversion multipliers.

 `'h'` (hexadecimal integer) format specifier was used for `Radd` & `Degd`, to display more digits in the debugger.
 
 Let me know if you want me to revise anything.

### DIFF
--- a/src/debuggers/natvis/magnum.natvis
+++ b/src/debuggers/natvis/magnum.natvis
@@ -132,10 +132,21 @@
     <DisplayString IncludeView="simple">{{ {hue._value,g}, {saturation,g}, {value,g} }}</DisplayString>
     <DisplayString ExcludeView="simple">{{ h={hue._value,g} s={saturation,g} v={value,g} }}</DisplayString>
   </Type>
-  <!-- Math::Quaternion -->
-  <Type Name="Magnum::Math::Quaternion&lt;*&gt;">
+  <!-- Math::Quaternion< Float > -->
+  <Type Name="Magnum::Math::Quaternion&lt;float&gt;">
+    <Intrinsic Name="Dot" Expression="_vector._data[0]*_vector._data[0]+_vector._data[1]*_vector._data[1]+_vector._data[2]*_vector._data[2]+_scalar*_scalar"></Intrinsic>
+    <Intrinsic Name="IsNormalized" Expression="Dot()&gt;0.99980000000500002&amp;Dot()&lt;1.0001999999950000"></Intrinsic>
     <DisplayString IncludeView="simple">{{ {_vector._data[0],g}, {_vector._data[1],g}, {_vector._data[2],g}, {_scalar,g} }}</DisplayString>
-    <DisplayString ExcludeView="simple">{{ x={_vector._data[0],g} y={_vector._data[1],g} z={_vector._data[2],g} w={_scalar,g} }}</DisplayString>
+    <DisplayString ExcludeView="simple" Condition="!IsNormalized()">{{ x={_vector._data[0],g} y={_vector._data[1],g} z={_vector._data[2],g} w={_scalar,g} }}</DisplayString>
+    <DisplayString ExcludeView="simple" Condition="IsNormalized()">{{ UNIT QUATERNION x={_vector._data[0],g} y={_vector._data[1],g} z={_vector._data[2],g} w={_scalar,g} }}</DisplayString>
+  </Type>
+  <!-- Math::Quaternion< Double > -->
+  <Type Name="Magnum::Math::Quaternion&lt;double&gt;">
+    <Intrinsic Name="Dot" Expression="_vector._data[0]*_vector._data[0]+_vector._data[1]*_vector._data[1]+_vector._data[2]*_vector._data[2]+_scalar*_scalar"></Intrinsic>
+    <Intrinsic Name="IsNormalized" Expression="Dot()&gt;0.99999999999998002&amp;Dot()&lt;1.0000000000000200"></Intrinsic>
+    <DisplayString IncludeView="simple">{{ {_vector._data[0],g}, {_vector._data[1],g}, {_vector._data[2],g}, {_scalar,g} }}</DisplayString>
+    <DisplayString ExcludeView="simple" Condition="!IsNormalized()">{{ x={_vector._data[0],g} y={_vector._data[1],g} z={_vector._data[2],g} w={_scalar,g} }}</DisplayString>
+    <DisplayString ExcludeView="simple" Condition="IsNormalized()">{{ UNIT QUATERNION x={_vector._data[0],g} y={_vector._data[1],g} z={_vector._data[2],g} w={_scalar,g} }}</DisplayString>
   </Type>
   <!-- Math::RectangularMatrix -->
   <Type Name="Magnum::Math::RectangularMatrix&lt;*&gt;">
--- a/src/debuggers/natvis/magnum.natvis
+++ b/src/debuggers/natvis/magnum.natvis
@@ -161,24 +161,14 @@
       </ArrayItems>
     </Expand>
   </Type>
-  <!-- Magnum::Math::Deg< Float > -->
-  <Type Name="Magnum::Math::Deg&lt;float&gt;">
-    <DisplayString IncludeView="simple">{_value,g}° ({_value * 3.14159274f / 180.0f,g} radians)</DisplayString>
-    <DisplayString ExcludeView="simple">{_value,g}° ({_value * 3.14159274f / 180.0f,g} radians)</DisplayString>
+  <!-- Magnum::Math::Deg -->
+  <Type Name="Magnum::Math::Deg&lt;*&gt;">
+    <DisplayString IncludeView="simple">{_value,g}° ({_value * 3.1415926535897931 / 180.0,g} radians)</DisplayString>
+    <DisplayString ExcludeView="simple">{_value,g}° ({_value * 3.1415926535897931 / 180.0,g} radians)</DisplayString>
   </Type>
-  <!-- Magnum::Math::Deg< Double > -->
-  <Type Name="Magnum::Math::Deg&lt;double&gt;">
-    <DisplayString IncludeView="simple">{_value,h}° ({_value * 3.1415926535897931 / 180.0,h} radians)</DisplayString>
-    <DisplayString ExcludeView="simple">{_value,h}° ({_value * 3.1415926535897931 / 180.0,h} radians)</DisplayString>
-  </Type>
-  <!-- Magnum::Math::Rad< Float > -->
-  <Type Name="Magnum::Math::Rad&lt;float&gt;">
-    <DisplayString IncludeView="simple">{_value,g} radians ({180.0f * _value / 3.14159274f,g}°)</DisplayString>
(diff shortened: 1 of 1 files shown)


src/debuggers/natvis/magnum.natvis
```diff
@@ -38,16 +39,64 @@
     <DisplayString IncludeView="simple">{{ {_data[0],g}, {_data[1],g} }}</DisplayString>
     <DisplayString ExcludeView="simple">{{ x={_data[0],g} y={_data[1],g} }}</DisplayString>
   </Type>
+  <!-- Math::Vector2< Float > -->
+  <Type Name="Magnum::Math::Vector&lt;2,float&gt;">
+    <Intrinsic Name="Dot" Expression="_data[0]*_data[0]+_data[1]*_data[1]"></Intrinsic>
+    <Intrinsic Name="IsNormalized" Expression="Dot()&gt;0.99980000000500002&amp;Dot()&lt;1.0001999999950000"></Intrinsic>
+    <DisplayString IncludeView="simple">{{ {_data[0],g}, {_data[1],g} }}</DisplayString>
+    <DisplayString ExcludeView="simple" Condition="!IsNormalized()">{{ x={_data[0],g} y={_data[1],g} }}</DisplayString>
+    <DisplayString ExcludeView="simple" Condition="IsNormalized()">{{ UNIT VECTOR x={_data[0],g} y={_data[1],g} }}</DisplayString>
```
mosra, line 48: For consistency with how the APIs are named I'd maybe say `NORMALIZED VECTOR` instead, or put a lowercase `(normalized)` in parentheses after. 

Being affter the value could make the actual numeric values aligned better, and that way people don't have to pay attention to the normalization property if they don't need to. I think.

Same in other cases.

src/debuggers/natvis/magnum.natvis
```diff
@@ -38,16 +39,64 @@
     <DisplayString IncludeView="simple">{{ {_data[0],g}, {_data[1],g} }}</DisplayString>
     <DisplayString ExcludeView="simple">{{ x={_data[0],g} y={_data[1],g} }}</DisplayString>
   </Type>
+  <!-- Math::Vector2< Float > -->
+  <Type Name="Magnum::Math::Vector&lt;2,float&gt;">
+    <Intrinsic Name="Dot" Expression="_data[0]*_data[0]+_data[1]*_data[1]"></Intrinsic>
+    <Intrinsic Name="IsNormalized" Expression="Dot()&gt;0.99980000000500002&amp;Dot()&lt;1.0001999999950000"></Intrinsic>
```
mosra, line 45: It might be worth to add a comment with how the values were calculated in case the epsilons in TypeTraits get updated in the future. Something like `<!-- 2*epsilon + epsilon*epsilon, epsilon = 1.0e-whatever  -->` could be sufficient I think, similarly for the double version.

I'm thinking if it would be better to use the full expression in there, but since the same value is repeated several times, it'd be a lot of noise. Not sure.

src/debuggers/natvis/magnum.natvis
```diff
@@ -101,4 +161,24 @@
       </ArrayItems>
     </Expand>
   </Type>
+  <!-- Magnum::Math::Deg< Float > -->
+  <Type Name="Magnum::Math::Deg&lt;float&gt;">
+    <DisplayString IncludeView="simple">{_value,g}° ({_value * 0.0174532924,g} radians)</DisplayString>
```
mosra, line 166: Same here and below, just a `<!-- pi/180 -->` etc. Hmm, or maybe it could be typed out in full, i.e. `_value* (3.14whatever/180.0)`? Is that supported in this context or you can do that just in the `Intrinsic`?

mosra (commented):
Looks good to me apart from these few things. Thanks!

@pezcode, since you're a direct user of these and I'm not, can you have a look as well? Thank you!

src/debuggers/natvis/magnum.natvis
fauder: Makes sense. On it 👍

src/debuggers/natvis/magnum.natvis
fauder: Yes, it is supported. It would be clearer to do so 👍

src/debuggers/natvis/magnum.natvis
fauder: I can add the comments right away.

Regarding writing the full expressions, I'd say it would be clearer from a maintenance perspective. Let me know if you'd like me to revise it.

mosra: > I'd say it would be clearer from a maintenance perspective.

Yeah, true, then do that :)

fauder: Oh wait I see now, yeah that line would get really long and harder to read if we put the epsilon's literal value and add/subtract with 1.0. Plus a lot of repetition of this same line.

How about we place a single comment block up top and explain how the values were calculated?

src/debuggers/natvis/magnum.natvis
fauder: I went with typing it all out route.

src/debuggers/natvis/magnum.natvis
fauder: I forgot to push the quaternions so there are separate commits for vectors and quaternions. Squash/fixup if desired.

pezcode: Another option for displaying this information is through an extra item in the expanded state. Requires you to open the item in the variables window, so not quite as compact, but it seems a bit more "idiomatic" (for whatever that's worth, even MS is rather inconsistent with their std natvis definitions).

Example here: https://github.com/mosra/corrade/blob/b192240cfeeb1d8a8a5d7fc13431480909cc9125/src/debuggers/natvis/corrade.natvis#L190

I think both options are fine, just throwing it out there.

src/debuggers/natvis/magnum.natvis
```diff
@@ -101,4 +161,24 @@
...
+  <!-- Magnum::Math::Rad< Float > -->
+  <Type Name="Magnum::Math::Rad&lt;float&gt;">
+    <DisplayString IncludeView="simple">{_value,g} radians ({180.0f * _value / 3.14159274f,g}°)</DisplayString>
+    <DisplayString ExcludeView="simple">{_value,g} radians ({180.0f * _value / 3.14159274f,g}°)</DisplayString>
+  </Type>
+  <!-- Magnum::Math::Rad< Double > -->
+  <Type Name="Magnum::Math::Rad&lt;double&gt;">
+    <DisplayString IncludeView="simple">{_value,h} radians ({180.0 * _value / 3.1415926535897931,h}°)</DisplayString>
+    <DisplayString ExcludeView="simple">{_value,h} radians ({180.0 * _value / 3.1415926535897931,h}°)</DisplayString>
+  </Type>
```
pezcode, line 183: Is the float/double specialization here really needed? Using the float value of pi even for doubles should be enough for quick and reasonably accurate debug output ☺️

src/debuggers/natvis/magnum.natvis
```diff
@@ -38,16 +39,64 @@
...
+    <Intrinsic Name="IsNormalized" Expression="Dot()&gt;0.99999999999998002&amp;Dot()&lt;1.0000000000000200"></Intrinsic>
+    <DisplayString IncludeView="simple">{{ {_data[0],g}, {_data[1],g} }}</DisplayString>
+    <DisplayString ExcludeView="simple" Condition="!IsNormalized()">{{ x={_data[0],g} y={_data[1],g} }}</DisplayString>
+    <DisplayString ExcludeView="simple" Condition="IsNormalized()">{{ x={_data[0],g} y={_data[1],g} (normalized) }}</DisplayString>
+  </Type>
   <!-- Math::Vector3 -->
   <Type Name="Magnum::Math::Vector&lt;3,*&gt;">
     <DisplayString IncludeView="simple">{{ {_data[0],g}, {_data[1],g}, {_data[2],g} }}</DisplayString>
     <DisplayString ExcludeView="simple">{{ x={_data[0],g} y={_data[1],g} z={_data[2],g} }}</DisplayString>
   </Type>
```
pezcode, line 62: Is there any use in keeping the generic versions? Personally I think no, but it should at least be consistent. Quaternion doesn't have a generic one anymore, neither do Deg/Rad

src/debuggers/natvis/magnum.natvis
mosra: Yes so basically what I suggested at first :)

src/debuggers/natvis/magnum.natvis
mosra: Good point. I think using the double value of pi even for floats (i.e., the inverse of your suggestion) wouldn't hurt.

src/debuggers/natvis/magnum.natvis
mosra: These are still needed for integer variants, I'd say. Quaternion or Deg/Rad doesn't have integer variants, so those have just float/double specializations.

src/debuggers/natvis/magnum.natvis
fauder: I was fine with it but wasn't sure so went the safe route.
Then I'm removing the specializations and using the double pi value as @mosra said 👍

fauder:
I'll resolve the remaining two points of interest tonight.

src/debuggers/natvis/magnum.natvis
fauder: How is this for the comment?

```
  ...

<!--For Vector & Quaternion types, IsNormalized() functionality was implemented by comparing the Dot() against 1.0f/1.0 with 2*epsilon
    where epsilon = 1.0e-5f for float & epsilon = 1.0e-14 for double variants.-->
  
  <!-- Math::Vector2 -->
  <Type Name="Magnum::Math::Vector&lt;2,*&gt;">
    <DisplayString IncludeView="simple">{{ {_data[0],g}, {_data[1],g} }}</DisplayString>
    <DisplayString ExcludeView="simple">{{ x={_data[0],g} y={_data[1],g} }}</DisplayString>
  </Type>
  
  ...
```

mosra:
Added the explanation from your comment above, squashed the commits together a bit and merged as 3b1fb704ca1ef2d5584f0da388a379632c199fee...2eee19276cd84ca6e8ea84207f82581d749b4d55. Thank you!

PR #641, opened 2024-07-08
Return EM_BOOL from emscripten callback lambdas to fix build on 3.1.62

Twinklebear:
In Emscripten 3.1.62 the type def for `EM_BOOL` was changed from `int` to `bool`, see [emscripten 3.1.61 `EM_BOOL`](https://github.com/emscripten-core/emscripten/blob/3.1.61/system/include/emscripten/em_types.h#L32) vs. [emscripten 3.1.62 `EM_BOOL`](https://github.com/emscripten-core/emscripten/blob/3.1.62/system/include/emscripten/em_types.h#L34). This change broke the `EmscriptenApplication`'s callback registration calls since the lambdas are marked as returning `Int` and no longer have the right type. This PR fixes this issue by changing the return type to `EM_BOOL` so it should build on both Emscripten 3.1.61 and 3.1.62


mosra:
This makes sense, thank you, and thanks for the background info as well!

PR #650, opened 2024-10-06
Magnum RPM package and build script

alkavan:
This package build was checked on clean Fedora 40, and all dependencies were installed correctly.

### DIFF
--- a/package/rpm/magnum.spec
+++ b/package/rpm/magnum.spec
@@ -1,0 +1,109 @@
+Name: magnum
+URL: http://magnum.graphics
+Version: 2020.06.2972.g8b832382a
+Release: 1
+Summary: C++11/C++14 graphics middleware for games and data visualization
+License: MIT
+Source: %{name}-%{version}.tar.gz
+BuildRoot: %{_tmppath}/%{name}-%{version}-%{release}-buildroot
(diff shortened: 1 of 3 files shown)


package/rpm/magnum.spec
```diff
@@ -0,0 +1,109 @@
+Name: magnum
+URL: http://magnum.graphics
+Version: 2020.06.2972.g8b832382a
+Release: 1
+Summary: C++11/C++14 graphics middleware for games and data visualization
+License: MIT
+Source: %{name}-%{version}.tar.gz
+BuildRoot: %{_tmppath}/%{name}-%{version}-%{release}-buildroot
+Requires: corrade, openal-soft, mesa-libGL, mesa-libEGL, SDL2, glfw, vulkan-loader, glslang
+BuildRequires: cmake, git, gcc-c++, openal-soft-devel, mesa-libGL-devel, mesa-libEGL-devel, SDL2-devel, glfw-devel, vulkan-loader-devel, glslang-devel
```
mosra, line 10: Just for the record, I think `glslang-devel` isn't needed here and I'm going to remove it. 

It will be needed for `GlslangShaderConverter` in the plugins tho.

mosra:
Merged with similar modification as in mosra/corrade#186 as d3b83f090c15cc0e31ac182baa2b52996420a95a. Thanks a lot!

PR #653, opened 2024-11-10
gdb: Fix missing comma in printers.py

asmaloney:


mosra:
Thank you! Sorry that this slipped through.

PR #660, opened 2025-01-23
Android) add mechanics to resume application from background

hsdk123:
Currently, the application will crash if sent to the background and brought back to the foreground.
The following provides a mechanism to continue the application.

### DIFF
--- a/src/Magnum/Platform/AndroidApplication.cpp
+++ b/src/Magnum/Platform/AndroidApplication.cpp
@@ -120,15 +120,17 @@ bool AndroidApplication::tryCreate(const Configuration& configuration, const GLC
         return false;
     }
 
+    _glConfiguration = glConfiguration;
+
     /* Choose config */
     const EGLint configAttributes[] = {
         EGL_SURFACE_TYPE, EGL_WINDOW_BIT,
-        EGL_RED_SIZE, glConfiguration.colorBufferSize().r(),
-        EGL_GREEN_SIZE, glConfiguration.colorBufferSize().g(),
-        EGL_BLUE_SIZE, glConfiguration.colorBufferSize().b(),
-        EGL_ALPHA_SIZE, glConfiguration.colorBufferSize().a(),
-        EGL_DEPTH_SIZE, glConfiguration.depthBufferSize(),
-        EGL_STENCIL_SIZE, glConfiguration.stencilBufferSize(),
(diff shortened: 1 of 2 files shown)


src/Magnum/Platform/AndroidApplication.cpp
```diff
@@ -176,6 +178,71 @@ bool AndroidApplication::tryCreate(const Configuration& configuration, const GLC
...
+    }
+
+    /* Make the context current */
+    CORRADE_INTERNAL_ASSERT_OUTPUT(eglMakeCurrent(_display, _surface, _surface, _glContext));
+    return true;
+}
+
+void AndroidApplication::pauseContext() {
+    /* Pause the context current */
+    CORRADE_INTERNAL_ASSERT_OUTPUT(eglMakeCurrent(_display, nullptr, nullptr, nullptr));
```
mosra, line 223: Would this work if you kept the context current and just made it bound to no surface at all?

```suggestion
    /* Keep the GL context current but without any surface */
    CORRADE_INTERNAL_ASSERT_OUTPUT(eglMakeCurrent(_display, EGL_NO_SURFACE, EGL_NO_SURFACE, _glContext));
```

Because I'm wondering if it's guaranteed that no app code gets executed between `APP_CMD_LOST_FOCUS` and `APP_CMD_INIT_WINDOW`. And if it does, it'd be probably good to have the GL context current.

mosra (commented):
If I understand correctly, this is about handling the case where GL context isn't lost and just the window surface is, right? So it's not like existing GL objects become invalid afterwards.

I'll have to think how to expose this in a way that makes sense with other apps. I can imagine that for example on desktop with SDL/GLFW one might want to attach to some `focusEvent()` / `blurEvent()` as well, when the window gets active or inactive.

src/Magnum/Platform/AndroidApplication.cpp
hsdk123: The suggested change, at least from my side, seems to work fine.

hsdk123: > I'm wondering if it's guaranteed that no app code gets executed between APP_CMD_LOST_FOCUS and APP_CMD_INIT_WINDOW

I'm not sure about there being a guarantee, but it's one of the reasons I added `_application_has_valid_surface` - such that the user has some info to branch on.

PR #674, opened 2025-06-02
Fix use of deprecated UDL syntax

falbrechtskirchinger:
C++23 has deprecated spaces between quotes and suffix in user-defined string literals. See [C++23 draft N4950 [depr.lit] (Annex D.9)](https://timsong-cpp.github.io/cppwp/n4950/depr.lit).

This addresses the following GCC 15 compiler warnings:
```
warning: space between quotes and suffix is deprecated in C++23 [-Wdeprecated-literal-operator]
```

See also:
- https://github.com/mosra/corrade/pull/201

### DIFF
--- a/src/Magnum/Math/Angle.h
+++ b/src/Magnum/Math/Angle.h
@@ -179,7 +179,7 @@ Example usage:
 @see @link operator""_degf() @endlink, @link operator""_rad() @endlink
 @m_keywords{_deg deg}
 */
-constexpr Deg<Double> operator"" _deg(long double value) { return Deg<Double>(Double(value)); }
+constexpr Deg<Double> operator""_deg(long double value) { return Deg<Double>(Double(value)); }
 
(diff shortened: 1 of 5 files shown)


mosra:
I have a warning suppression in place for this for Clang, so I'll expand it to GCC 15 as well. I don't compile with C++23 enabled on my end, so didn't notice GCC 15 started to be annoying about this as well, thanks for letting me know!

mosra:
I expanded the warning suppression in da40122233aac7c26d88136e775bf87f1a50e02c, and another one that fires only for `-std=c++17` and up in be6926d781c4c3f8718c097ef2826597f93e3b64.

It's likely that the old syntax will become unsupported by the compilers at some point, but I hope that by then I won't need to support GCC 4.8 anymore and will thus be able to clean up all this mess and use just the space-less syntax with no warning suppression.

PR #695, opened 2026-05-31
Expose font index in AbstractFont

arsnyder16:
Adds first-class AbstractFont support for opening indexed fonts from collections, including openFile() / openData() overloads with fontIndex, plus fontIndex() and fontCount() accessors.


mosra:
Merged a variant of this as 8ae70a038abb0667c953553f4a57b46b693af4c6. I ended up doing the API changes a bit differently which took a while, along with other changes to allow for zero-copy loading in 2401a2243ddcae6ef88e9e7c2f47689c96546b07, prep work for "always open" fonts in 53a500bf47b53d94cb66eedad797fba9faff8ea2 and corresponding backwards compatibility interfaces & tests so existing plugins still work with the new code.
 
Thanks for working on this!

